PR #7, opened 2024-02-12
TSC-105-Bypass-Default-Age-Request-in-Java-File

asivath:
Finished adding popup that asks users if they want to use the default ages in the Datapack.


app/src/state/actions/GeneralActions.ts
```diff
@@ -226,6 +226,17 @@ export const resetState = action("resetState", () => {
   state.settingsJSON = {};
 });
 
+// Shows the user a popup before chart generation if there are age spans on the datapack
+export const showPopupOnSuggAge = action("showPopupOnSuggAge", () => {
```
paologum, line 230: the workflow in my opinion would be better like this 
1. use this function to check if we should show popup, if not we call generate chart in this function. 
2. if we need to show a popup, set popup to true has usual. then handle generating a chart in the handler of that dialog. 
this would remove the need for a userResponded var

app/src/Home.tsx
```diff
@@ -47,7 +47,24 @@ export const Home = observer(function Home() {
...
+      actions.setUseSuggestedAge(response);
+      actions.setUseCache(false);
+    }
+    actions.setUserResponded(true);
+    actions.setShowSuggestedAgePopup(false);
+  };
+  const handleCloseDialog = () => {
+    actions.setShowSuggestedAgePopup(false);
+  };
+  useEffect(() => {
```
paologum, line 61: as said in `GeneralActions.ts` comment, don't use `useEffect` to listen for dependency changes and instead in the `handlePopupResponse`, set the required items and generate a chart accordingly. (call `generateChart` in `handlePopupResponse`.

app/src/Home.tsx
```diff
@@ -47,7 +47,24 @@ export const Home = observer(function Home() {
   const { state, actions } = useContext(context); 
   const theme = useTheme();
   const navigate = useNavigate();
-
+  const handlePopupResponse = (response) => {
```
paologum, line 50: Move this instead to an actions file. It probably would be good just for convenience for you to make a new file called `GenerateChartActions.ts` where you have your method `ShowPopupOnSuggAge` and this `HandlePopupResponse` and `generateChart` the large function.

paologum: you will need to export it out of `index.ts` when you make the file

app/src/Home.tsx
```diff
@@ -90,8 +115,7 @@ export const Home = observer(function Home() {
                 <TSCButton 
                   className="info-button"
                   onClick={() => {
-                    actions.generateChart();
-                    navigate('/chart');
```
paologum, line 94: just call whatever you decide to call `showPopupOnSuggAge` here.
it might be better to call it something general since this will be the `generate chart` function. i'm thinking maybe something like `initiateChartGeneration` which everyone would understand to use

might also want to change the current `generateChart` to something else so that we discern the starting point of every attempt to generate a chart. something like `fetchChartFromServer` since that seperates it from `initiateChartGeneration`

app/src/Home.tsx
```diff
@@ -58,6 +75,14 @@ export const Home = observer(function Home() {
           <TSCPresetHighlights key={type} navigate={navigate} configArray={configArray} type={type}/>
         )
       })}
+      <TSCPopupDialog
```
paologum, line 78: If we execute this properly, we might have multiple uses of generate chart throughout the website, so the popup should be able to be visible everywhere. Therefore you might want to put this in `App.tsx` similar to the snack bar i put in.

app/src/Home.tsx
paologum: additionally, you will have to change the `Column.tsx` `generateChart` button to whatever `initiateChartGeneration` becomes

app/src/state/state.ts
```diff
@@ -18,6 +18,9 @@ export type State = {
   tab: number;
   madeChart: boolean;
   showPresetInfo: boolean;
+  showSuggestedAgePopup: boolean;
+  userResponded: boolean;
```
paologum, line 22: deprecate `userResponded`

paologum (changes requested):
changing workflow to better encapsulate the intended use of the popup.

PS for multi datapack use i haven't aggregated the correct fields so if you're testing that it puts multiple files together in the `setDatapackConfig` it doesn't work atm.

app/src/Home.tsx
paologum: `index.ts` of `actions.ts`

paologum (changes requested):

PR #9, opened 2024-02-13
Tsc 165 parsing age range

JacquiFreeElectron:


server/src/parse-datapacks.ts
```diff
@@ -253,6 +330,38 @@ async function getFacies(filename: string, facies: Facies) {
...
+
+function processBlock(line: string): SubBlockInfo | null {
+  let currentSubBlockInfo = {
+    label: "",
+    age: 0,
+    info: "",
+    lineType: ""
+  }
+  const tabSeperated = line.split('\t')
+  // if (tabSeperated.length < 4) return null (don't know if it's the same for blocks too)
```
paologum, line 342: you probably should find the minimum `tabSeperated.length` possible. otherwise if you don't want to return a null object, then you need to change the return type to just `SubBlockInfo` and change line 316 to not check if `SubBlockInfo` exists. however i strongly support the first suggestions i made.

server/src/parse-datapacks.ts
```diff
@@ -242,7 +297,29 @@ async function getFacies(filename: string, facies: Facies) {
...
+      //get a single sub block
+      let SubBlockInfo = processBlock(line)
+      if (SubBlockInfo) {
+
+        SubBlockInfos.push(SubBlockInfo);
+      }
+    }
   }
+
   if (inFaciesBlock) {
```
paologum, line 323: this is actually a very rare edge case but as you can see after the for loop ends, i check if we're in a faciesblock, this is because the file can end, meaning that the readstream for loop stops without us inserting a block/facies entry since that requires us to be in the loop. therefore when we exit, we have to check if we're in a block and add it accordingly to the blocks map.

paologum (changes requested):
couple comments that need to be resolved but overall great 😄

server/src/parse-datapacks.ts
```diff
@@ -329,171 +338,190 @@ async function getFaciesOrBlock(filename: string, facies: Facies, blocksMap: Map
...
+      currentSubBlockInfo.label = label
+    }
+    currentSubBlockInfo.age = age
+    if (info) {
+      currentSubBlockInfo.info = info
+    }
+    if (lineType) {
+      currentSubBlockInfo.lineType = lineType
+    }
+
```
paologum, line 386: i missed this in the first code review but you should `assertSubBlockInfo` here in a `try-catch` and return null if there is one. see line 420 with my implementation

shared/dist/index.d.ts
```diff
@@ -197,7 +197,7 @@ export type SubBlockInfo = {
 };
 export type ChartRequest = {
     settings: string;
-    // columnSettings: object;
+    columnSettings: string;
```
paologum, line 200: this actually doesn't relate to your task and might throw a merge conflict so i'm just gonna say that this line needs to be deleted

paologum (changes requested):
good stuff, just some more comments

server/src/parse-datapacks.ts
```diff
@@ -301,14 +275,21 @@ async function getFaciesOrBlock(filename: string, facies: Facies, blocksMap: Map
...
+        block.on = false
+      }
+
       let info = tabSeperated[tabSeperated.length - 1]
       const pattern = /\"*\"/
 
-      if (pattern.test(info)) {
-        block.info = info
+      if (info) {
+        if (pattern.test(info)) {
```
paologum, line 287: can you make this just `if(info && pattern.test(info)` instead

shared/src/index.ts
```diff
@@ -405,12 +405,32 @@ export function assertSubBlockInfo(o: any): asserts o is SubBlockInfo {
...
     throw new Error("SubBlockInfo must have a lineType value with type string")
 }
 
+export function assertBlock(o: any): asserts o is Block {
+  if (!o || typeof o !== "object")
+    throw new Error("Block must be a non-null object");
+
+  if (typeof o.name !== "string")
+    throw new Error("Block must have a name with string type")
+  for (const subBlockInfo in o.subBlockInfo) {
```
paologum, line 420: there's actually a difference between `in` and `of` when using enchanced for loops

`in` gives you the keys
`of` gives you the values

in this case we want the values of the array(`SubBlockInfo`) and not the keys(`..1,2,3,...`)

paologum (changes requested):

paologum (approved):

PR #10, opened 2024-02-13
Front end dialog box

paologum:
This is for the front end to display an error to the client. If multiple errors are thrown, multiple errors can be seen. Also refactored `fetchPresets` and `fetchDatapackInfo` in `initialize.ts` to `GeneralActions.ts`

image


shared/dist/index.js
jay-jhk-lee, line 1: This was done when I moved the parsing to be done on the app side, the app only sends the xml settings to the server.

shared/dist/index.d.ts
jay-jhk-lee, line 1: This was done when I moved the parsing to be done on the app side, the app only sends the xml settings to the server.

jay-jhk-lee (commented):
I removed the column settings from chart request when moving the parsing to the app side

jay-jhk-lee (approved):

PR #14, opened 2024-02-16
added expand/collapse all functionality

jay-jhk-lee:
https://github.com/earthhistoryviz/tsconline/assets/89719669/01dd6be5-dfe7-4717-931b-f1b4c338542e


app/src/SettingsTabs/Column.tsx
```diff
@@ -108,20 +131,51 @@ const ColumnAccordion: React.FC<ColumnAccordionProps> = observer(
...
+  ]);
+  //if column not in expanded list, add it
+  //if column in expanded list, remove it
+  const accordionClicked = (name: string) => {
+    if (expandedAccordions.includes(stringToHash(name))) {
+      setExpandedAccordions(
+        expandedAccordions.filter((number) => number !== stringToHash(name))
+      );
+    } else setExpandedAccordions([...expandedAccordions, stringToHash(name)]);
+    console.log(expandedAccordions);
```
paologum, line 148: remove log

app/src/SettingsTabs/Column.tsx
```diff
@@ -18,23 +18,37 @@ import {
...
+}
+
 const ColumnAccordion: React.FC<ColumnAccordionProps> = observer(
-  ({ name, details }) => {
+  ({ name, details, expandedAccordions, accordionClicked }) => {
     const theme = useTheme();
     const { state, actions } = useContext(context);
-    const [open, setOpen] = useState(false);
     //for keeping the original name for array access
     let ogName = useRef(name);
```
paologum, line 51: not related to the commits but wondering why you use ref
useRef is reserved for htmlElements so you probably should make this a useState
but also at the same time do you need to rerender everytime it changes or can you just leave it a normal variable?

app/src/SettingsTabs/Column.tsx
```diff
@@ -18,23 +18,37 @@ import {
...
 type ColumnAccordionProps = {
   name: string;
   details: ColumnInfo;
+  expandedAccordions: number[];
+  accordionClicked: (name: string) => void;
 };
 
+//for using integers instead of strings inside expandAccordion state array
+//to increase speed
+function stringToHash(string: string): number {
```
paologum, line 32: is this because it runs `.includes` faster on an array with numbers?

app/src/SettingsTabs/Column.tsx
```diff
@@ -77,11 +91,18 @@ const ColumnAccordion: React.FC<ColumnAccordionProps> = observer(
...
+          accordionClicked(ogName.current);
+        }}
+      >
         <AccordionSummary aria-controls="panel-content" id="panel-header">
           <div
-            onClick={() => {
-              toggleAccordion(open);
+            onClick={(event) => {
+              //stops accordion from expanding/collapsing when clicking on the name or checkbox
+              event.stopPropagation();
```
paologum, line 105: wait so how do you expand an accordion now if clicking it doesn't change the expandedAccordion

paologum: or i guess push its name unless i'm missing something

paologum (changes requested):

app/src/SettingsTabs/Column.tsx
jay-jhk-lee: the accordion doesn't expand when it clicks the name or checkbox, which is intentional. It still opens if you click the icon or anywhere outside the name

app/src/SettingsTabs/Column.tsx
jay-jhk-lee: yea

app/src/SettingsTabs/Column.tsx
jay-jhk-lee: Good catch, that was a holdover from when there was no distinction between name and editName, so whenever editName changed the name changed also. Because of that, we needed the first editName for array access. Since those two are separate now, it can be removed.

app/src/SettingsTabs/Column.tsx
jay-jhk-lee: ok

app/src/state/parseSettings.ts
```diff
@@ -203,9 +203,13 @@ function generateSettingsXml(settings: any, indent: string): string {
     if (Object.prototype.hasOwnProperty.call(settings, key)) {
       const value = settings[key];
       if (typeof value === "object") {
+        //overhaul top age and base age
+        //top age and baseage settings can have two children: text and stage
+        //the source attribute determines which one is used
         if (key === "topAge" || key === "baseAge") {
+          console.log(value);
```
paologum, line 210: one more log

app/src/state/parseSettings.ts
```diff
@@ -203,9 +203,13 @@ function generateSettingsXml(settings: any, indent: string): string {
...
           xml += `${indent}<setting name="${key}" source="${value.source}" unit="${value.unit}">\n`;
-          xml += `${indent}    <setting name="text">${value.text}</setting>\n`;
+          xml += `${indent}    <setting name="${value.source}">${value.text}</setting>\n`;
```
paologum, line 212: this is more of a comment but we could make the xml only use one or the other since we create the json ourselves if this ends up being too complicated. but if it works though we definitely should do it this way

paologum (approved):

app/src/state/parseSettings.ts
jay-jhk-lee: Yeah that could be a better method

PR #16, opened 2024-02-16
TSC-199-font-options

lbostre:
fontOptions
- finished frontend for the font options popup
- allows user to change options for the font
5c000580b4c6e672a43207749b9ddd21
- here is mobile view


app/src/SettingsTabs/FontMenu.tsx
```diff
@@ -99,7 +99,7 @@ const FontMenuRow: React.FC<{
                         <MenuItem value={"Verdana"}>Verdana</MenuItem>
                     </Select>
                 </FormControl>
-                <TextField sx={{width: "60px", marginRight: "10px"}} label="Size" variant="outlined"
+                <TextField sx={{width: "60px", marginRight: "10px", minWidth:"60px"}} id="FontSizeInput"label="Size" variant="outlined"
```
paologum, line 102: you define both an id for the css and sx? probably pick one or the other or add a comment saying that you have to do both of these things.

however you set `min-width` in css and here so at least choose one or the other there. preferably all within `FontSizeInput` if possible

server/src/parse-datapacks.ts
```diff
@@ -318,7 +304,7 @@ function recursive(
     name: currentColumn,
     editName: currentColumn,
     on: true,
-    fontsInfo: fontsInfo,
+    fontsInfo: JSON.parse(JSON.stringify(defaultFontsInfo)),
```
jay-jhk-lee, line 307: Can you double check if you need to stringify and parse? if defaultFontsInfo is already a json object, I think it shouldn't be necessary

paologum: however if you know for a fact it won't change the shared object then you can remove it

app/src/SettingsTabs/FontMenu.tsx
lbostre: CSS styling doesn't work on this specific component for some reason, so I removed the CSS class and kept the inline styling. I also added a comment signifying that only inline styling works on this component as well.

paologum (changes requested):
I found something while playing around with the viewbox and you can see that the right examples of the sample text are coming out of the dialog box. You can either fix this with  switching over earlier to the flex-wrap.

Screenshot 2024-02-18 at 7 25 22 PM

app/src/SettingsTabs/FontMenu.tsx
```diff
@@ -1,200 +1,221 @@
...
+            displayEmpty
+          >
+            <MenuItem value={"Arial"}>Arial</MenuItem>
+            <MenuItem value={"Courier"}>Courier</MenuItem>
+            <MenuItem value={"Verdana"}>Verdana</MenuItem>
+          </Select>
+        </FormControl>
+        {/* CSS styling doesn't work on TextField */}
+        <TextField
+          sx={{ width: "60px", marginRight: "10px", minWidth: "60px" }}
```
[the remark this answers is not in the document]
lbostre: Applied and pushed!

app/src/SettingsTabs/FontMenu.css
```diff
@@ -32,6 +32,8 @@
   width: 180px;
 }
 
+.MuiTextField-root.font-size-container { width: 60px; margin-right: 10px; min-width: 60px; }
```
paologum, line 35: little picky here but can you format this like the rest of the id classes in the css file 😃

paologum (changes requested):

PR #20, opened 2024-02-19
datapack frontend page

rupp3:
TSC-131-datapack-picture


app/src/Datapack.tsx
```diff
@@ -1,23 +1,114 @@
...
+  const [selectedDatapacks, setSelectedDatapacks] = useState<number[]>([]);
+
+  const handleCheckboxChange = (id: number) => {
+    const isSelected = selectedDatapacks.includes(id);
+    if (isSelected) {
+      setSelectedDatapacks(selectedDatapacks.filter((selectedId) => selectedId !== id));
+    } else {
+      setSelectedDatapacks([...selectedDatapacks, id]);
+    }
+    setCheckedNotApplied(true);
```
paologum, line 65: i think the logic here should just take a `name: string` and you can just do the same logic but with `state.config.datapacks`. That variable is what we use to keep track of the datapacks used.

paologum: otherwise deprecate `setSelectedDatapacks` and use `state.config.datapacks` instead

this will mean you need to make an action that does what you want since you can't change the state object outside of an action. See `GeneralActions.ts` for more info on that.

app/src/Datapack.tsx
```diff
@@ -1,23 +1,114 @@
...
-        width: "100%",
-        minHeight: "100vh",
-        background: theme.palette.settings.light,
-      }}
-    >
+    <div className="container">
+      <Typography variant="h2" className="heading">
+        TimeScale Creator Datapacks
+      </Typography>
+      <Typography style={{ fontSize: 18, marginTop: "1em", marginBottom: "3vh"}}>
```
paologum, line 73: move this style to css

for typography you will need to make a css class like so

`.MuiTypography-root.[class-name] {
....
}`

app/src/Datapack.tsx
```diff
@@ -1,23 +1,114 @@
...
+            {datapacks.map((datapack) => (
+              <tr key={datapack.id}>
+                <td className="checkbox-cell">
+                  <TSCCheckbox
+                    checked={selectedDatapacks.includes(datapack.id)}
+                    onChange={() => handleCheckboxChange(datapack.id)}
+                  />
+                </td>
+                <td className="name-cell">
+                  <div style = {{display: "flex", justifyContent: "space-between", alignItems: "center"}}>
```
paologum, line 88: add this to css

app/src/Datapack.tsx
```diff
@@ -1,23 +1,114 @@
...
+      event.preventDefault();
+    }
+  };
+
+  // dummy datapacks data
+  const datapacks = [
+    { id: 1, name: "British Isles", description: "British Isles regional formations with map interface. Approximately two dozen sub-regions, with each formation linked to the BGS lexicon; compiled in coordination with British Geological Survey, BGS." },
+    { id: 2, name: "Belgium", description: "Belgium regional formations with map interface. Detailed Table of contents and references at Belgium_LithostratContents.pdf. Compiled by Jane Block in coordination with Belgium Stratigraphic Commission (Noel Vandenberghe, Hance Luc, et al.)." },
+    { id: 3, name: "Australia", description: "Data for Australia" },
+  ];
```
paologum, line 54: just a comment but to see the actual forms the `datapack` see `DatapackParsingPack` in `shared`

might have to edit this to include descriptions but that's the object you'll be using for this page

paologum (changes requested):

app/src/Datapack.tsx
paologum: that name could be wrong but it should be relatively called that

app/src/Datapack.tsx
```diff
@@ -1,23 +1,79 @@
...
+
 export const Datapack = observer(function Datapack() {
-  const { actions } = useContext(context)
   const theme = useTheme();
+  const { actions } = useContext(context)
+  const [open, setOpen] = useState(false);
+
+  const handleClickOpen = () => {
+    setOpen(true);
+  };
```
paologum, line 22: is this used? if not remove

paologum (changes requested):

paologum (approved):
looks good! but lmk how this looks on mobile/post in the description

rupp3:
image

this is most recent change shown on mobile. i had to fix the table layout. nothing happens for the info icon so I will probably add a pop out for the info button if it is pressed on mobile (?)

rupp3:
image
image

most recent mobile view, added the background theme. when i click on a different tab and go back to the datapack tab the format changes to be centered but when i first go to the datapack tab it starts centered to the left. When i scroll over to the info tab the theme also does not carry to the rest of the box

app/src/Datapack.css
```diff
@@ -0,0 +1,51 @@
...
+  align-items: center;
+  justify-content: center;
+  padding: 16px;
+  border: 1px solid gray;
+  border-radius: 4px;
+  max-width: 600px;
+  margin: 0 auto;
+  margin-top: 20px;
+  margin-bottom: 20px;
+  width: 85%; /*this fixed the window scaling issue, changed from 100 */
```
paologum, line 27: hmmm ok i'm assuming this fixes the mobile issue where there's whitespace

paologum (approved):
looks good!

app/src/Datapack.css
rupp3: yeah, is that not allowed?

rupp3:
image

tobyonyekwere1 (commented):
Code is properly written and easy to understand as well. Great work!

JacquiFreeElectron (approved):
looks great

PR #28, opened 2024-02-22
Facies attached to ColumnInfo, bug fixes to block parsing, deprecating aliases, throwError in `shared`

paologum:
this removes the "third party" `Facies` variable and `Aliases` to now have everything included within the `ColumnInfo` variable

reference for what aliases are: 
we take the first child's facies info
this is due to map points taking their first child's facies information
Therefore meaning the child has the facies information but the map point is not 
called by the child name but is called the parent name
`parent -> child -> facies`
  displayed as `parent -> facies`
facies event exists for this map point
if `parent -> [childOne, childTwo]` we do `parent->childOneFacies`

`METACOLUMN_OFF` and `METACOLUMN_ON` both exists so implemented functionality for both of those things

also added `throwError` in `shared` to be used to consistently throw errors

additionally adds `@fastify/compress` to allow for auto compressed data communicated over the REST API

review as much as you can of `parse-datapacks.ts` since most of the work is around there


app/src/state/actions/GeneralActions.ts
```diff
@@ -146,9 +142,6 @@ export const setDatapackConfig = action(
           datapackParsingPack.columnInfoArray
         );
         // concat all facies
```
jay-jhk-lee, line 144: looks like unneeded comment now

jay-jhk-lee (commented):
Looks good

server/src/parse-datapacks.ts
```diff
@@ -88,28 +84,22 @@ export async function parseDatapacks(
...
-    maxAge: -99999,
-    aliases: {},
-  };
   const isChild: Set<string> = new Set();
-  const isFacies: Set<string> = new Set();
   const allEntries: Map<string, ParsedColumnEntry> = new Map();
   let datapackAgeInfo: DatapackAgeInfo = { datapackContainsSuggAge: false };
+  const faciesMap: Map<string, Facies> = new Map();
   const blocksMap: Map<string, Block> = new Map();
   //const faciesMap: Map<String, string[]> = new Map();
```
JacquiFreeElectron, line 92: it's actually me putting this comment in here in my previous task but I forgot to get rid of it😭, since you've implemented faciesMap I think we should just delete this comment

JacquiFreeElectron (commented):
👍🏻

PR #29, opened 2024-02-23
changed active jar

jay-jhk-lee:
changed active jar to generate default settings file


paologum (approved):
was the settings file bad?

SejalKumar03 (approved):
Looks good!

jay-jhk-lee:
The TSC2020 presets was using a settings file for the africa bights datapack, so I replaced it with the correct one

PR #30, opened 2024-02-23
if map point is not in range selected by user disable

paologum:
new `util.ts` that holds a method to check if data is in user range
color feedback appreciated
new colors to represent icons better
implements range according to the user selection
if out of range disables properly
Screenshot 2024-02-23 at 1 47 19 AM
Screenshot 2024-02-23 at 1 47 30 AM
Screenshot 2024-02-23 at 1 48 19 AM


SejalKumar03 (approved):
This looks good!

lbostre (approved):
Looks good!

PR #33, opened 2024-02-25
Tsc 242 fix time range and column on off

jay-jhk-lee:
moved parsing from xml to json to app side, added more fields for background color, modified jar file to fix chart generation, changed settingsJson route to settingsXml


app/src/state/parseSettings.ts
```diff
@@ -357,11 +379,10 @@ function generateColumnXml(
         let currName = extractName(jsonColumn._id);
         let childName = extractName(jsonColumn[key]._id);
         //TODO: pass the state column of the column itself, not the children array of its parent
-        let params: { one: any; two: any; three: string; four: string } = {
+        let params: { one: any; two: any; three: string } = {
           one: jsonColumn[key],
           two: null,
-          three: currName,
-          four: `${indent}    `,
+          three: `${indent}    `,
```
paologum, line 385: rename these to be more representative of what they represent

app/src/state/parseSettings.ts
```diff
@@ -207,9 +228,7 @@ function generateSettingsXml(
     if (Object.prototype.hasOwnProperty.call(settings, key)) {
       const value = settings[key];
       if (typeof value === "object") {
-        //overhaul top age and base age
-        //top age and baseage settings can have two children: text and stage
-        //the source attribute determines which one is used
+        //TODO: right now it can only do age, implement stage later
```
paologum, line 231: on the app side we keep track of the age specifically so you shouldn't need to implement stage at all

paologum (changes requested):
looks good, put on backlog to change the params of the xmlToJson to not be `any`

app/src/state/parseSettings.ts
```diff
@@ -307,7 +306,6 @@ function generateColumnXml(
         let useEditName = false;
         if (
           colName !== "Chart Root" &&
```
paologum, line 308: i forgot to mention that the root name for the charts is now `Root` not `Chart Root`

app/src/state/parseSettings.ts
```diff
@@ -378,24 +376,23 @@ function generateColumnXml(
...
             if (stateColumn.children[i].name == childName) {
-              params.two = stateColumn.children[i];
+              params.stateColumn = stateColumn.children[i];
               break;
             }
           }
         }
 
-        xml += generateColumnXml(params.one, params.two, params.three);
+        xml += generateColumnXml(params.jsonColumn, params.stateColumn, params.indent);
```
paologum, line 395: hmmm do you need a `params` object or can you just apply them directly here

paologum (changes requested):

app/src/state/parseSettings.ts
jay-jhk-lee: Good idea

app/src/state/parseSettings.ts
jay-jhk-lee: This should still be Chart Root, it's checking the names for the default settings object not the columnInfo object

server/src/routes.ts
```diff
@@ -28,8 +27,8 @@ export const fetchSettingsJson = async function fetchSettingsJson(
     const contents = (
       await readFile(`${decodeURIComponent(settingFile)}`)
     ).toString();
-    const settingJson = await xmlToJson(contents);
-    reply.send(settingJson);
+    const settingsXml = contents;
```
paologum, line 30: ok two comments
1. probably just remove `contents` since it's not used for anything and just go straight to `const settingsXml =  await readFile(${decodeURIComponent(settingFile))).toString();`
2. more of a comment but this could go bad with large settings file so at some point try a very large one to test (since we're reading it all in at once with readFile)

paologum (changes requested):
one thing

app/src/state/parseSettings.ts
```diff
@@ -291,17 +291,16 @@ function generateColumnXml(
   indent: string
 ): string {
   let xml = "";
+  console.log(stateColumn);
```
paologum: remove

app/src/state/parseSettings.ts
```diff
@@ -338,8 +337,15 @@ function generateColumnXml(
       } else if (key === "orientation") {
         xml += `${indent}<setting name="${xmlKey}" orientation="${jsonColumn[key]}"/>\n`;
       } else if (key === "isSelected") {
+        //use preset value for event columns, remove later when event columns are covered
```
paologum, line 339: make this a `TODO` comment

app/src/state/parseSettings.ts
```diff
@@ -97,21 +120,19 @@ function processColumn(node: any): any {
...
-                useNamed: useNamedValue,
-                text: child.textContent.trim(),
-              };
-            } else {
-              result[settingName] = child.textContent.trim();
-            }
-          } else if (justificationValue.length !== 0) {
+            result[settingName] = {
+              standardized: standardizedValue ? standardizedValue : undefined,
+              useNamed: useNamedValue ? useNamedValue : undefined,
```
paologum, line 130: is there a reason you can't just set this to `useNamedVlaue` and `standardizedValue`? Both `undefined` and `null` are falsy so they'll pass any if statements.

paologum (changes requested):

paologum (approved):
at some point the one console log should either be taken out or formatted to a nicer string but overall good

PR #34, opened 2024-02-25
move facies button to a bar above map

paologum:
added new header bar to replace bottom drawer
added new method to compare vh to px
styled for mobile and smaller/longer screens
added new package for number inputs
deprecated TSCNumberInput

image
image
image
image


app/src/util/util.ts
```diff
@@ -8,4 +8,32 @@
...
+ */
+export function compareVhAndPx(vh: number | string, px: number | string): number {
+    if (typeof vh === "string") {
+        if (vh.length <= 2) throw Error(`vh param in wrong format ${vh}`)
+        vh = Number(vh.slice(0, -2))
+    }
+    if (typeof px === "string") {
+        if (px.length <= 2) throw Error(`px param in wrong format ${px}`)
+        px = Number(px.slice(0, -2))
+    }
```
asivath, line 28: could check that it ends width "vh" or "px" instead of just length. Also should check that the conversion is successful.

jay-jhk-lee (approved):
Looks good

PR #35, opened 2024-02-25
TSC-146-Read-Excel-Sheet-and-Send-to-App

SejalKumar03:
Changed useEffect to useContext and fixed dropdown connection with text


server/assets/datapacks/AfricaBight.map
paologum, line 1: remove this file

paologum: and the directory `assets/datapacks`

server/assets/datapacks/Hominids_wImages_GTS2020_Encrypted.dpk
paologum, line 1: same

server/assets/datapacks/TimeTree_of_Life_GTS2020_ENCRYPTED_23Jan2020.dpk
paologum, line 1: same

server/src/index.ts
```diff
@@ -175,6 +179,95 @@ server.post<{ Params: { usecache: string, useDatapackSuggestedAge: string } }>(
   routes.fetchChart
 );
 
+// Serve timescale data endpoint
+server.get('/timescale', routes.fetchTimescale);
+
+server.post('/charts', async (request, reply) => {
```
paologum, line 185: this should be the `routes.fetchChart` method we talked about

server/src/routes.ts
```diff
@@ -228,3 +231,50 @@ export const fetchChart = async function fetchChart(
   });
   reply.send({ chartpath: chart_urlpath, hash: hash });
 };
+
+function readExcelFile(filePath: string) {
```
paologum, line 235: can you put this method in a file called `parse-excel-file` and rename to `parseExcelFile`

app/src/state/state.ts
```diff
@@ -54,6 +58,9 @@ export type State = {
   settingsXML: string;
   settingsJSON: any;
   settings: {
+    geologicalTopStageAges: TimescaleItem[];
+    geologicalBaseStageAges: TimescaleItem[];
```
paologum, line 62: we should only have two arrays `geologicalTopStageAges` and `geologicalBaseStageAges` and they should be located where `GeologicalTopStageAges` and `GeologicalBaseStageAges` are currently so just those two and make them camel case

app/src/SettingsTabs/Time.tsx
```diff
@@ -4,140 +4,128 @@ import { actions, state } from "../state";
...
-            }}
-            style={{ marginBottom: "20px", width: "100%" }}
-          />
-        </FormControl>
+          name="vertical-scale-text-field"
+          value={state.settings.topStageAge.toString()}
+          onChange={(event) => {
+            const age = parseFloat(event.target.value);
+            if (!isNaN(age) && age >= 0 && age <= state.settings.baseStageAge) {
+              actions.setSelectedTopStage(age.toString());
```
paologum, line 67: set it to N/A instead of `age.toString()` since the `age` probably won't correspond to a key

app/src/SettingsTabs/Time.tsx
```diff
@@ -4,140 +4,128 @@ import { actions, state } from "../state";
...
         <TextField
+          className="BaseAgeTextField"
+          label="Base Age"
+          type="number"
+          name="vertical-scale-text-field"
+          value={state.settings.baseStageAge.toString()}
+          onChange={(event) => {
+            const age = parseFloat(event.target.value);
+            if (!isNaN(age) && age >= 0 && state.settings.topStageAge <= age) {
+              actions.setSelectedBaseStage(age.toString());
```
paologum, line 107: same

paologum (changes requested):
almost done!!

server/assets/datapacks/AfricaBight.map
SejalKumar03: You want me to delete the entire server/assets/datapacks? But there's other things in there. Should I move things outside the directory?

paologum: you should just delete them with `git rm --cached` because the datapacks shouldb't be on the remote repo. they should just be available on your local machine. normally they should be .gitignored automatically. you can do `git rm -r --cached server/assets/datapacks`

server/src/parse-excel-file.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function parseExcelFile(filePath: string) {
+    const workbook = XLSX.readFile(filePath);
+    const sheetName: string = workbook.SheetNames[0] || "";
+    const sheet = workbook.Sheets[sheetName];
+    if (!sheet) return [];
+  
+    const jsonData: any[] = XLSX.utils.sheet_to_json(sheet, { header: 1 });
+  
+    const dataWithoutFirstRow = jsonData.slice(1);
```
paologum, line 11: can you rename this `dataWithoutHeader` so anyone reading this understands why you sliced

server/src/parse-excel-file.ts
```diff
@@ -0,0 +1,18 @@
...
+  
+    if (dataWithoutFirstRow.length > 0) {
+      dataWithoutFirstRow[0] = dataWithoutFirstRow[0].slice(3);
+    }
```
paologum, line 15: if this is empty you should throw an error that says no timescale data to parse

paologum: also can i ask why you `slice` here?

server/src/routes.ts
```diff
@@ -228,3 +230,33 @@ export const fetchChart = async function fetchChart(
...
   reply.send({ chartpath: chart_urlpath, hash: hash });
 };
+
+// Serve timescale data endpoint
+export const fetchTimescale = async function (
+  _request: FastifyRequest,
+  reply: FastifyReply
+) {
+  try {
+    const filePath = '../default_timescale.xlsx';
```
paologum, line 240: since this is an asset can you move `default_timescale.xlsx` into the `server/assets` folder and make a new field in the `server/assets/config.json` that specifices this filepath.

you'll have to change the `AssetConfig` type and `asserts` in the `server/src/types.ts` file to account for your new change.

Once you do you can just use `assetconfigs.timescaleFilepath` or whatever you name the field in the `config.json` file.

paologum: let me know if you have any questions about this

paologum (changes requested):
just some code styling/refactoring but the functionality looks great!

server/src/parse-excel-file.ts
SejalKumar03: Because I am getting rid of the first 3 columns since they are just black in the excel sheet

app/src/state/actions/general-actions.ts
```diff
@@ -52,13 +54,19 @@ export const fetchFaciesPatterns = action("fetchFaciesPatterns", async () => {
  */
 export const resetSettings = action("resetSettings", () => {
   state.settings = {
+    // geologicalTopStageAges: [],
+    // geologicalBaseStageAges: [],
```
paologum, line 58: remove

app/src/state/state.ts
```diff
@@ -58,7 +62,21 @@ export type State = {
   chartHash: string;
   settingsXML: string;
   settingsJSON: any;
-  settings: Settings;
+  settings: {
+    // geologicalTopStageAges: TimescaleItem[];
+    // geologicalBaseStageAges: TimescaleItem[];
```
paologum, line 67: remove

app/src/state/state.ts
```diff
@@ -118,13 +139,19 @@ export const state = observable<State>({
   settingsXML: "",
   settingsJSON: {},
   settings: {
+    // geologicalTopStageAges: [],
+    // geologicalBaseStageAges: [],
```
paologum, line 143: remove

app/src/SettingsTabs/Time.tsx
```diff
@@ -1,128 +1,129 @@
...
-              Object.keys(state.settingsTabs.geologicalTopStages).map((key) => (
-                <MenuItem value={key} key={key}>
-                  {`${key} (${state.settingsTabs.geologicalTopStages[key]} Ma top)`}
-                </MenuItem>
-              ))}
+              const selectedValue = event.target.value;
+              const selectedAgeItem = state.geologicalTopStageAges.find((item) => item.key === selectedValue);
+              const selectedAge = selectedAgeItem?.value?.toString() || "0";
+              const selectedAgeNumber = parseFloat(selectedAge);
+              if (selectedAgeNumber >= 0 && selectedAgeNumber <= state.settings.baseStageAge) {
```
paologum, line 47: hmm there isn't a reason to turn the value to a string and then parse the float since it is already a number so just change it to use `const selectedAgeItem = .... || 0` and check on that

app/src/SettingsTabs/Time.tsx
```diff
@@ -1,128 +1,129 @@
...
-              Object.keys(state.settingsTabs.geologicalBaseStages).map((key) => (
-                <MenuItem value={key} key={key}>
-                  {`${key} (${state.settingsTabs.geologicalBaseStages[key]} Ma base)`}
-                </MenuItem>
-              ))}
+              const selectedValue = event.target.value;
+              const selectedAgeItem = state.geologicalBaseStageAges.find((item) => item.key === selectedValue);
+              const selectedAge = selectedAgeItem?.value?.toString() || "0";
+              const selectedAgeNumber = parseFloat(selectedAge);
+              if (selectedAgeNumber >= 0 && selectedAgeNumber >= state.settings.topStageAge) {
```
paologum, line 86: same as above

server/src/parse-excel-file.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  const jsonData: string[][] = XLSX.utils.sheet_to_json(sheet, { header: 1 });
+
+  const dataWithoutHeader = jsonData.slice(1);
+
+  // if (dataWithoutHeader.length > 0) {
+  //   dataWithoutHeader[0] = dataWithoutHeader[0].slice(3);
+  // } else {
+  //   assertTimescaleDataWithoutHeader(dataWithoutHeader);
+  // }
```
paologum, line 18: remove

server/src/parse-excel-file.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!dataWithoutHeader || dataWithoutHeader.length === 0) {
+    assertTimescaleDataWithoutHeader(dataWithoutHeader);
+    return [];
+  }
+
+  if (dataWithoutHeader[0]) {
+    dataWithoutHeader[0] = dataWithoutHeader[0].slice(3);
+  }
```
paologum, line 27: since you make a check if `length == 0` you don't need to check that one index exists, so you can just slice the 3 without the `if`

shared/src/index.ts
```diff
@@ -386,6 +386,11 @@ export function assertPatterns(o: any): asserts o is Patterns {
     assertColor(pattern.color);
   }
 }
+export type TimescaleItem = {
+  key: string;
+  value: number;
+};
+
```
paologum, line 393: can you move this to the top, so all of the types are grouped together

shared/src/index.ts
```diff
@@ -756,3 +761,18 @@ export function assertSVGStatus(o: any): asserts o is SVGStatus {
 function throwError(obj: string, variable: string, type: string, value: any) {
   throw new Error(`Object '${obj}' must have a '${variable}' ${type} property.\nFound value: ${value}`);
 }
+
+export function assertTimescale(val: any): asserts val is TimescaleItem {
+  if (!val || typeof val !== "object") {
+    throw new Error("Timescale object must be a non-null object");
+  }
+  if (typeof val.key !== "string" || typeof val.value !== "number") {
+    throw new Error(`Timescale object must have 'key' of type string and 'value' of type number`);
```
paologum, line 770: can you use the new `throwError` method

shared/src/index.ts
```diff
@@ -756,3 +761,18 @@ export function assertSVGStatus(o: any): asserts o is SVGStatus {
...
+  }
+}
+
+export function assertTimescaleDataWithoutHeader(data: any[]): asserts data is any[] {
+  if (data.length === 0) {
+    throw new Error("No timescale data to parse");
+  }
+}
```
paologum, line 778: i don't think this method is very useful in regards to it being refactored and in shared. can you just make this `if` statement where it's being used? i'll add another comment there too

server/src/parse-excel-file.ts
```diff
@@ -0,0 +1,30 @@
...
+  // if (dataWithoutHeader.length > 0) {
+  //   dataWithoutHeader[0] = dataWithoutHeader[0].slice(3);
+  // } else {
+  //   assertTimescaleDataWithoutHeader(dataWithoutHeader);
+  // }
+
+  if (!dataWithoutHeader || dataWithoutHeader.length === 0) {
+    assertTimescaleDataWithoutHeader(dataWithoutHeader);
+    return [];
+  }
```
paologum, line 23: hmmm this is where you would change it but that return will never get hit because that `assert` method throws an error if it's empty and if it reaches this point, it will always be empty. you probably should just throw an error then and not return and deprecate `assertTimescaleDataWithoutHeader`

package.json
```diff
@@ -27,7 +27,8 @@
     "fs": "^0.0.1-security",
     "nvm": "^0.0.4",
     "ts-node": "^10.9.2",
-    "typescript": "^5.0.4"
+    "typescript": "^5.0.4",
+    "xlsx": "^0.18.5"
```
paologum, line 31: can you remove xlsx from the `root` packages, it should only be needed in the `server` dir

paologum (changes requested):

package.json
paologum: you can just do `yarn remove xlsx` in `tsconline`

app/src/SettingsTabs/Time.tsx
```diff
@@ -1,128 +1,127 @@
...
   const handleButtonClick = () => {
     actions.setTab(1);
-    actions.fetchChartFromServer(navigate);
+    // actions.setAllTabs(true);
+
+    actions.updateSettings();
+
+    actions.generateChart();
+
+    navigate("/chart");
```
paologum, line 28: oops i forgot to mention this is supposed to be `actions.fetchChartFromServer(navigate)` and `actions.setTab(1)`
and get rid of everything else 

after you do this just verify it works one more time just to make sure

paologum (changes requested):
ok last thing 😭

app/src/SettingsTabs/Time.tsx
```diff
@@ -1,128 +1,127 @@
...
 
+import { useContext } from "react";
+import { context } from "../state/index";
+import "./Time.css";
+import theme from "../theme";
+
 export const Time = observer(function Time() {
-  const theme = useTheme();
   const navigate = useNavigate();
+  const { state } = useContext(context);
```
paologum, line 19: okok and one more thing don't import `actions` from `state`, the only thing you should import from `state` is `context`

instead to access actions do `const { state, actions } = useContext(context)`  on line 18 and get rid of the import on line 3

paologum (approved):
great!

app/src/SettingsTabs/Time.tsx
```diff
@@ -1,128 +1,120 @@
...
-                  {`${key} (${state.settingsTabs.geologicalTopStages[key]} Ma top)`}
-                </MenuItem>
-              ))}
+              const selectedValue = event.target.value;
+              const selectedAgeItem = state.geologicalTopStageAges.find((item) => item.key === selectedValue);
+              const selectedAge = selectedAgeItem?.value || 0;
+              if (selectedAge >= 0 && selectedAge <= state.settings.baseStageAge) {
+                actions.setSelectedTopStage(selectedValue);
+                actions.setTopStageAge(selectedAge);
+              }
```
asivath, line 42: having an error message that pops up when stage is invalid would be user friendly. easiest way would be to use the TSCError component and the pushError action. same for any other user input

SejalKumar03:
https://github.com/earthhistoryviz/tsconline/assets/119654886/3e6227e0-99c1-4b08-9a14-9732180095f4

app/src/state/actions/general-actions.ts
```diff
@@ -99,6 +105,31 @@ export const loadIndexResponse = action("loadIndexResponse", (response: IndexRes
...
+        const item = stages[i];
+        const value = i > 0 ? stages[i - 1].value : 0;
+        state.geologicalTopStageAges.push({ ...item, value });
+      }
+      setGeologicalBaseStageAges(state.geologicalBaseStageAges);
+      setGeologicalTopStageAges(state.geologicalTopStageAges);
+
+      console.log("Time Scale Data Loaded");
+    } else {
+      console.error("Server responds with:", response.status);
```
paologum, line 127: i forgot to mention this but can you call `displayError` with the appropriate parameters here
i believe `displayError(null, response, "{message you want}")` would work

app/src/state/actions/general-actions.ts
```diff
@@ -99,6 +105,31 @@ export const loadIndexResponse = action("loadIndexResponse", (response: IndexRes
...
+    }
+  } catch (error) {
+    console.error("Error fetching timescale data:", error);
```
paologum, line 130: same here you can either call it with `displayError(null, null, message)` or `pushError(message)`

paologum (changes requested):

app/src/state/actions/general-actions.ts
paologum: except you want the response json since we need to access the error message so you need `await response.json()`

app/src/state/actions/general-actions.ts
```diff
@@ -99,6 +105,31 @@ export const loadIndexResponse = action("loadIndexResponse", (response: IndexRes
...
+
+      for (let i = 0; i < stages.length; i++) {
+        const item = stages[i];
+        const value = i > 0 ? stages[i - 1].value : 0;
+        state.geologicalTopStageAges.push({ ...item, value });
+      }
+      setGeologicalBaseStageAges(state.geologicalBaseStageAges);
+      setGeologicalTopStageAges(state.geologicalTopStageAges);
+
+      console.log("Time Scale Data Loaded");
```
paologum, line 125: multiple warning also occur when you edit `state.geologicalBaseStageAges` and `state.geologicalTopStageAges` in place so instead just use a temporary variable called `geologicalBaseStageAges` and `state.geologicalTopStageAges` and then set them as you do on 122 instead of doing it on `state`

asivath (approved):

paologum (approved):

PR #37, opened 2024-02-25
TSC-225-CI-CD-Pipeline

asivath:
The merge-pipeline workflow runs two jobs on opening a PR, reopening a PR, and committing changes to a PR:

code-quality:
After setting up the environment it runs ESLint and Prettify. If either reports issues, it will cause the job to fail and will add a comment to the PR describing how to fix the issue or what changes need to be made.

build-and-test:
After setting up the environment in tries to build the app, and if the command returns any errors the job will fail and it will add a comment to the PR describing what failed. I left a "Run Tests" step for running any testing.

I also added four different yarn scripts:
yarn prettier-format: Makes prettier to fix issues it finds
yarn prettier-check: Makes prettier to list all issues without fixing
yarn lint-check: Makes eslint list all errors
yarn lint-format: Makes eslint fix any errors it can


paologum (approved):
eyes are glazed but looks good

PR #41, opened 2024-02-29
add patterns to legend with added filter

paologum:
new server packages:
color-thief-node: finding dominant color in an image
nearest-color: nearest neighbors (KNN problem) to get nearest color given the dominant color

new app packages: 
react-menu: new react menu that supports hover and more capabilities (used for filter)
simplebar-core: types to allow for wrapping of simplebar package

features:
can view patterns on the front end that exist on the server
can filter by present in map (will change as you move slider)
can filter by color, alphabetically, or by search value
each image only has one color associated with it
Because of the algorithm, it seems some "other" colors get grouped into "Golden Brown" because of weird dominant averages
might fix at a later time
util to compare strings

Questions:
could be convinced for images to have a color palette to bypass the dominant average color
any new filters you might think would be useful
any rendering/styling comments

https://github.com/earthhistoryviz/tsconline/assets/89664317/57467d75-6615-4695-a7c0-57ee7635082b


app/src/SettingsTabs/Legend.tsx
```diff
@@ -0,0 +1,345 @@
...
+        className={`scroll-top ${isScrolled ? "show" : ""}`}
+        style={{ top: `calc(${LEGEND_HEADER_HEIGHT} + 1vh)` }}
+        onClick={() => {
+          scrollRef.current?.contentWrapperEl?.scrollTo({ top: 0, left: 0, behavior: "smooth" });
+        }}>
+        <Lottie key="legend-arrow-up" width="inherit" height="inherit" animationData={ArrowUpIcon} playOnHover />
+      </IconButton>
+      <StyledScrollbar
+        ref={scrollRef}
+        style={{
```
jay-jhk-lee, line 123: is the inline styling necessary because of the calc you're doing?

app/src/components/TSCComponents.tsx
```diff
@@ -11,7 +11,32 @@ export const ColoredIconButton = styled(IconButton)(({ theme }) => ({
...
+    borderRadius: "2px"
+  },
+  position: "relative",
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "left",
+  fontSize: "1.1rem",
+  paddingLeft: "21px",
+  paddingBottom: "30px"
+}));
```
paologum, line 34: refactored this to prevent className overlap (this happens on every shared component and this is my best solution) if it happens to you use `key` to stop the overlap but i like this solution better

app/src/SettingsTabs/Legend.tsx
```diff
@@ -0,0 +1,345 @@
...
+          className="sub-menu-item"
+          key={color}
+          checked={colorFilter.has(color)}
+          onClick={() => toggleColor(color)}
+          type="checkbox">
+          <>
+            <TypographyText key={color} className="color-text">
+              {color}
+            </TypographyText>
+            <div className="colored-box" style={{ backgroundColor: colorHash.get(color)?.hex }} />
```
jay-jhk-lee, line 286: is the inline styling necessary because you're accessing the hashmap?

server/assets/colors.json
jay-jhk-lee, line 1: where did you get the colors from? Are they all the colors available for TSC facies patterns?

app/src/SettingsTabs/Legend.tsx
paologum: yup

app/src/SettingsTabs/Legend.tsx
paologum: hmmm i guess since in this instance the header doesn't conditionally load i can port it to a css variable but let me think about it for a second since this'll mean making a `className` on the styledscrollbar which would mean all `StyledScrollbar`'s would need a unique `key` or all of them would have that height. now that i'm typing it out it probably shouldn't be done like that and this is necessary.

jay-jhk-lee:
For the colors there's a section in the tsc manual that lists all the colors that tsc uses; that might be a good reference. The interface looks great!

paologum:
mobile view of border circles

image

app/src/SettingsTabs/Legend.tsx
```diff
@@ -0,0 +1,345 @@
...
+    if (isPresent && matchesSearch) {
+      colors.add(value.color.name);
+      colorHash.set(value.color.name, value.color);
+    }
+    return isPresent && matchesSearch && hasColor;
+  });
+  if (alphabeticalSort) {
+    filteredPatterns.sort((a, b) => {
+      return compareStrings(a.formattedName, b.formattedName);
+    });
```
asivath, line 92: idk if this actually has an impact but this sort happens on every re-render? could rapid state changes cause this to become laggy on less powerful devices?

paologum: we technically rerender everything so if anything the filter will actually be more stressful than this. Alternatively I could just sort it when it gets loaded onto the app but if we do that simply for runtime i don't think it would have much effect. We deal with a very small < 500 patterns which is a large overestimate. Right now we have ~50. Additionally, any rerenders will limit the amount of patterns a good deal making this sort ~10-20 patterns since we rerender on filter changes. We can't really avoid rerunning over the array without going through every possible combination of filters and storing in the state somewhere but that doesn't seem very intuitive or efficient

server/src/routes.ts
```diff
@@ -1,19 +1,66 @@
...
+          hex: color.value,
+          rgb: color.rgb
+        }
+      };
+    }
+    assertPatterns(patterns);
+    reply.status(200).send({ patterns });
+  } catch (e) {
+    console.error(e);
+    reply.status(404).send({ error: e });
```
asivath, line 61: this is mean to be a catch all right, so wouldn't 500 be better since 404 specifically means endpoint not found?

server/src/types.ts
```diff
@@ -6,8 +6,22 @@ export type AssetConfig = {
...
 };
 
+export type Colors = {
+  [color: string]: string;
+};
+
+export function assertColors(o: any): asserts o is Colors {
+  if (typeof o !== "object" || !o) throw "AssetConfig must be an object";
+  for (const color in o) {
+    if (typeof color !== "string") throw 'Colors must have a "color" key that is a string';
```
asivath, line 20: this is looping over the keys right so isn't it always a string?

server/src/util.ts
```diff
@@ -74,14 +83,42 @@ export async function grabFilepaths(files: string[], topDirectory: string, botDi
...
+/**
+ * Converts rgb of three numbers to hex code
+ * @param r
+ * @param g
+ * @param b
+ * @returns
+ */
+export function rgbToHex(r: number, g: number, b: number) {
+  return "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
+}
```
asivath, line 124: validate rgb is within 0-255

app/src/SettingsTabs/Legend.tsx
paologum: the main problem is having to filter on fast rerendering by reiterating over the array so do you think we should keep it real time or have an apply filter to keep it low

this also reruns everytime the user changes the text field value

shared/src/index.ts
```diff
@@ -345,6 +362,27 @@ export type VertBounds = {
...
 };
 
+export function assertColor(o: any): asserts o is Color {
+  if (!o || typeof o !== "object") throw new Error("Color must be a non-null object");
+  if (typeof o.name !== "string") throwError("Color", "name", "string", o.color);
+  if (typeof o.hex !== "string") throwError("Color", "hex", "string", o.hex);
+  if (typeof o.rgb !== "object") throwError("Color", "rgb", "object", o.rgb);
+  if (typeof o.rgb.r !== "number") throwError("Color", "r", "number", o.rgb.r);
+  if (typeof o.rgb.g !== "number") throwError("Color", "g", "number", o.rgb.g);
+  if (typeof o.rgb.b !== "number") throwError("Color", "b", "number", o.rgb.b);
```
asivath, line 372: check that the rgb values are within 0-255

shared/src/index.ts
```diff
@@ -345,6 +362,27 @@ export type VertBounds = {
...
+}
+export function assertPatterns(o: any): asserts o is Patterns {
+  if (!o || typeof o !== "object") throw new Error("Patterns must be a non-null object");
+  for (const key in o) {
+    if (typeof key !== "string") throwError("Patterns", "key", "string", key);
```
asivath, line 377: again won't keys always be strings?

server/src/routes.ts
paologum: good catch will fix

server/src/util.ts
paologum: we technically don't use this but i'll keep it on the code base just in case and validate the rgb values

shared/src/index.ts
paologum: you can index by anything so we have to validate that it is a string ie `type ex = {[key: number]: value}`

shared/src/index.ts
paologum: good catch

app/src/SettingsTabs/Legend.tsx
asivath: well if it's not that many patterns than it's probably not worth it premature optimization is not good anyway

shared/src/index.ts
asivath: doesn't js cast anything to a string anyway though?

paologum: ok yes that's true will fix.

will probably change a lot more on the shared folder since many of them assume it doesn't cast

asivath (approved):
looks good!

PR #43, opened 2024-03-02
Tsc 244 reformat add block columns to reflect documentation

JacquiFreeElectron:


paologum (changes requested):
some changes

server/src/parse-datapacks.ts
```diff
@@ -226,7 +229,21 @@ async function getFaciesOrBlock(filename: string, faciesMap: Map<string, Facies>
...
+      block.title = trimQuotes(tabSeperated[0]!);
+      if (tabSeperated[2]) {
+        block.width = Number(tabSeperated[2]);
+      }
+      if (tabSeperated[3]) {
+        if (!(tabSeperated[3] === "nocolor" || tabSeperated[3] === "UNESCO-Named")) {
+          const rgb = tabSeperated[3].split("/");
+          block.color.r = Number(rgb[0]!);
+          block.color.g = Number(rgb[1]!);
+          block.color.b = Number(rgb[2]!);
```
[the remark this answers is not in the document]
paologum: you probably could just check it with `assertRGB` and not with an `isNaN` here

server/src/parse-datapacks.ts
```diff
@@ -36,16 +38,15 @@ function spliceArrayAtFirstSpecialMatch(array: string[]): ParsedColumnEntry {
...
-          parsedColumnEntry.on = true;
-        } else {
-          parsedColumnEntry.on = false;
-        }
+        parsedColumnEntry.on = array[i] === "_METACOLUMN_ON";
+      } else {
+        parsedColumnEntry.enableTitle = array[i] === "_TITLE_ON";
       }
       array.splice(i, 1);
       i = i - 1;
```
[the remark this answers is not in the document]
paologum: instead of nested if statements, they're each their own one `if` statement

shared/dist/tempCodeRunnerFile.ts
paologum, line 1: what is this file? you should probably delete this if you don't know

server/src/parse-datapacks.ts
```diff
@@ -238,21 +253,53 @@ export async function getFaciesOrBlock(
...
-      block.name = trimQuotes(tabSeperated[0]!);
+      block.title = trimQuotes(tabSeperated[0]!);
+
+      if (tabSeperated[2]) {
+        block.width = Number(tabSeperated[2]);
+      }
+      if (isNaN(block.width)) {
+        console.log(`Error found while processing block width, setting width to 100`);
+        block.width = 100;
+      }
```
paologum, line 264: you should put the `isNaN` inside the first if statement after `block.width` gets set since you'd be making unnecessary checks on widths that don't exist.

server/src/parse-datapacks.ts
```diff
@@ -238,21 +253,53 @@ export async function getFaciesOrBlock(
...
       //get a single sub block
-      const subBlockInfo = processBlock(line);
+
+      //using parameterRGB make sure we don't pass by reference
+      const parameterRGB: RGB = {
+        r: block.rgb.r,
+        g: block.rgb.g,
+        b: block.rgb.b
+      };
+      const subBlockInfo = processBlock(line, parameterRGB);
```
paologum, line 302: you can just do `const subBlockInfo = processBlock(line, {...block.rgb});`
which is an inline shallow copy of what you're doing

server/src/parse-datapacks.ts
```diff
@@ -238,21 +253,53 @@ export async function getFaciesOrBlock(
...
       if (tabSeperated[5] && tabSeperated[5] === "off") {
         block.on = false;
       }
-      const popup = tabSeperated[tabSeperated.length - 1];
-      const pattern = /"*"/;
 
-      if (popup && pattern.test(popup)) {
-        block.popup = popup;
+      if (tabSeperated[6] && patternForPopup.test(tabSeperated[6])) {
+        block.popup = tabSeperated[6];
```
paologum, line 289: hmm there might be some popups might not have "" surrounding them so you can just have "if it exists at index 6" it is guranteed to be a popup.

unless you think otherwise

server/src/parse-datapacks.ts
```diff
@@ -238,21 +253,53 @@ export async function getFaciesOrBlock(
...
+        block.rgb.b = Number(rgbSeperated[2]!);
+      }
+      try {
+        assertRGB(block.rgb);
+      } catch (e) {
+        console.log(`Error ${e} found while processing block rgb, setting rgb to white`);
+        block.rgb.r = 255;
+        block.rgb.g = 255;
+        block.rgb.b = 255;
+      }
```
paologum, line 279: move the `try` also into the above `if` statement on line 266 because we should only check if we actually change the value

paologum (changes requested):
some changes but overall great

server/__tests__/parse-datapacks.test.ts
```diff
@@ -78,6 +110,90 @@ describe("process facies line tests", () => {
   });
 });
 
+describe("process blocks line tests", () => {
+  it("should process block line for top label of age 100 with default color and default lineStyle", () => {
+    const defaultColor = { r: 255, g: 255, b: 255 };
```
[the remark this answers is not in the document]
paologum: this will also mean removing the declaration at all the other tests for `color` but i won't spam you and make all those comments 💀

server/src/parse-datapacks.ts
```diff
@@ -238,21 +253,53 @@ export async function getFaciesOrBlock(
 
     // we found a block
     if (!inBlockBlock && tabSeperated[1] === "block") {
-      block.name = trimQuotes(tabSeperated[0]!);
+      block.title = trimQuotes(tabSeperated[0]!);
+
+      if (tabSeperated[2]) {
+        block.width = Number(tabSeperated[2]);
```
paologum, line 259: this is something i'm doing for event headers but since this is getting kind of big, can you refactor this to a function called `setBlockHeaders(block, tabSeperated)`

paologum (approved):
looks good!

asivath (approved):

PR #47, opened 2024-03-05
map pack testing

paologum:
added map pack tests


server/src/parse-map-packs.ts
```diff
@@ -350,7 +259,158 @@ function grabRectBounds(headerLabels: { [key: number]: { label: string } }, info
...
+        vertBounds.centerLon = Number(info[i]);
+        break;
+      case "CENTER LAT":
+        vertBounds.centerLat = Number(info[i]);
+        break;
+      case "HEIGHT":
+        vertBounds.height = Number(info[i]);
+        break;
+      case "SCALE":
+        vertBounds.scale = Number(info[i]);
```
asivath, line 287: everywhere you cast you should check for nan

paologum: the fields of the variables are all checked to be the correct type in their corresponding `asserts`
these tests are just mainly for making sure the fields are being set
i mocked the asserts to all return true for isolation

server/src/parse-map-packs.ts
```diff
@@ -350,7 +259,158 @@ function grabRectBounds(headerLabels: { [key: number]: { label: string } }, info
...
+    }
+  }
+  return vertBounds;
+}
+
+function grabTransects(headerLabels: string[], info: string[]) {
+  const transect: any = {};
+  let name = "";
+  for (let i = 1; i < info.length; i++) {
+    if (!info[i] || !headerLabels || !headerLabels[i] || !headerLabels[i]!) continue;
```
asivath, line 300: !headerLabels[i]! shouldn't this cause an error?

server/src/parse-map-packs.ts
asivath: okay cool

server/src/parse-map-packs.ts
paologum: will take out

asivath (approved):
Looks good

PR #49, opened 2024-03-07
Tsc 227 add ssh to update dev and add files to workflow

asivath:
Added check-files job to merge-pipeline workflow that checks for disallowed files
Added second workflow that will update the dev site. Added the URL to a .env file instead of hardcoded inside utils.ts to get around merges on dev site.
Also added more linter settings to check for variable cases and filename conventions


.github/workflows/merge-pipeline.yml
```diff
@@ -1,12 +1,52 @@
-name: Default Merge Pipeline
+name: Merge Pipeline
 
 on:
   pull_request:
     types: [opened, synchronize, reopened] # Runs on opening a pr, pushing to a pr, and reopening a pr
     branches:
-      - main
+      - main  # This needs to be changed to main
```
paologum, line 7: is this comment still needed or

app/.env
```diff
@@ -0,0 +1 @@
+VITE_SERVER_URL=http://localhost:3000
```
paologum, line 1: wait how does this work because if the `.env` is tracked on the repo won't it create conflicts when pulling on the dev side since they're different

app/src/constant.ts
paologum, line 1: wait i don't know if this file is deleted or if the contents are deleted. just make sure this whole file is removed.

paologum (changes requested):

.github/workflows/merge-pipeline.yml
asivath: will remove

app/src/constant.ts
asivath: this is still in main branch but I can remove it

paologum (approved):

app/src/util.ts
```diff
@@ -1,6 +1,12 @@
 // If you are running in dev mode, prefix URL's with the dev server URL:
 
-const devurl = import.meta.env.VITE_SERVER_URL;
+let devurl: string;
+
+if (import.meta.env.VITE_SERVER_URL) {
+  devurl = import.meta.env.VITE_SERVER_URL;
+} else {
+  devurl = "http://localhost:3000";
+}
```
paologum, line 9: hmm you definitely don't have to change this but i feel like for more compacted code we could do
`let devurl: string = "http://localhost:3000"` instead of an else statement. it's like three less lines but it would look neater.
if you're fine with how it is already it's also fine.

paologum (approved):

PR #50, opened 2024-03-08
Tsc 215 connect font option changes to display on chart

lbostre:
- font options menu connects to display on chart (works for the column titles)

DEMO
https://github.com/earthhistoryviz/tsconline/assets/29764221/5477188b-7550-434c-8cfc-506591cf22ca


shared/src/index.ts
```diff
@@ -187,51 +207,159 @@ export type FontsInfo = {
...
     inheritable: boolean;
+    fontFace: "Arial" | "Courier" | "Verdana";
+    size: number;
+    bold: boolean;
+    italic: boolean;
+    color: string;
   };
 };
 
 export const defaultFontsInfo: FontsInfo = {
```
paologum, line 250: can you make a seperate file called `constants.ts` and export this from there to here and export again? this is just because now with prettier this gets pretty large and it's not exactly a type so it should have its own space

shared/src/index.ts
```diff
@@ -566,13 +712,12 @@ export function assertColumnInfo(o: any): asserts o is ColumnInfo {
 export function assertFontsInfo(o: any): asserts o is FontsInfo {
   if (typeof o !== "object") throw new Error("FontsInfo must be an object");
   for (const key in o) {
```
paologum, line 714: since we have the `ValidFontOptions` you should also check the string is of that type

paologum (changes requested):
some changes but great work👍

app/src/SettingsTabs/FontMenu.tsx
```diff
@@ -24,21 +24,15 @@ import FormatItalicIcon from "@mui/icons-material/FormatItalic";
...
 }> = observer(({ target }) => {
   const { state, actions } = useContext(context);
   const [fontTarget, setFontTarget] = useState(false);
   const [font, setFont] = useState("Arial");
-  const [formats, setFormats] = useState(["bold", "italic"]);
-
+  const [formats, setFormats] = useState([""]);
   const fontOpts = state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected as string)!.fontsInfo[target];
   const handleChange = (event: SelectChangeEvent) => {
     if (/Arial|Courier|Verdana/.test(event.target.value)) return;
```
paologum, line 38: i realize this as part of regex but you want to test the whole string so it should be `^/Arial|Courier|Verdana/$`

shared/src/index.ts
```diff
@@ -135,114 +137,46 @@ export type ColumnPrototypeInfo = {
...
+
+export const defaultFontsInfo: FontsInfo = defaultFontsInfoConstant;
+
+export type FaciesLocations = {
+  [location: string]: {
+    faciesTimeBlockArray: FaciesTimeBlock[];
+    minAge: number; // the min age of this specific location
+    maxAge: number; // the max age of this specific location
   };
 };
```
paologum, line 175: this is deprecated

shared/src/index.ts
```diff
@@ -135,114 +137,46 @@ export type ColumnPrototypeInfo = {
...
-  "Ruler Label": { inheritable: false },
-  "Ruler Tick Mark Label": { inheritable: false },
-  "Sequence Column Label": { inheritable: false },
-  "Uncertainty Label": { bold: false, color: "#000000", fontFace: "Arial", inheritable: false, italic: false, size: 5 },
-  "Zone Column Label": { bold: false, color: "#000000", fontFace: "Arial", inheritable: false, italic: false, size: 12 }
+export type FaciesTimeBlock = {
+  rockType: string; // rock type that is the name of the png in /public/patterns/
+  label?: string; // the label
+  age: number; // the base gge of the facies time block
 };
```
paologum, line 180: also deprecated

app/src/state/actions/column-actions.ts
```diff
@@ -64,160 +65,97 @@ export const setcolumnSelected = action((name: string) => {
...
+  if (!state.settingsTabs.columnSelected) {
+    throw new Error("state.settingsTabs.columnSelected is null");
+  }
+  const columnHashMapEntry = state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected);
+  if (!columnHashMapEntry) {
+    throw new Error(`Entry for ${state.settingsTabs.columnSelected} not found in columnHashMap`);
+  }
+  assertFontsInfo(columnHashMapEntry.fontsInfo[target]);
+  columnHashMapEntry.fontsInfo[target].inheritable = isInheritable;
+  assertFontsInfo(columnHashMapEntry?.fontsInfo[target]);
```
paologum, line 78: i'm slightly confused, you are sending the `assertFontsInfo` a `FontLabelOptions` object here right? In this case wouldn't you either send the `columnHashMapEntry?.fontsInfo` or make an `assertFontLabelOptions` and pass this variable there?

paologum (changes requested):
some questions/changes

shared/src/index.ts
```diff
@@ -1,5 +1,7 @@
 // Shared types between app and server (i.e. messages they send back and forth)
 
+import { defaultFontsInfoConstant } from "./constants";
```
paologum, line 3: does this work? did you not need to change this to `./constants.js`?

paologum:
ok one last thing, can you upload a demo here of you changing the font options and doing all the different options for fonts then generating a chart? just so i can verify in the future or whatever. after i see that i'll approve and you can merge

paologum (approved):

PR #51, opened 2024-03-08
switched datapack tab to settings page

rupp3:
image
image

Should I remove the title on the datapack tab now that it's not its own tab? Also mobile looks a little weird for the box margins so I can fix that a little later


app/src/Settings.tsx
```diff
@@ -28,6 +29,8 @@ export const Settings = observer(function Settings() {
         return <Font />;
       case "mappoints":
         return <MapPoints />;
+      case "datapack":
+        return <Datapack />;
```
paologum, line 33: can you instead name it datapacks just because it'll be about multiple datapacks more often than not

paologum (changes requested):
one small request 😄

paologum (approved):

PR #53, opened 2024-03-11
throw more uniform frontend error

paologum:
if stacking margin is too big lmk

Screenshot 2024-03-12 at 5 54 13 PM

image

image

image


app/src/state/actions/general-actions.ts
```diff
@@ -509,14 +519,27 @@ export const handleCloseSnackbar = action(
   }
 );
 
-export const removeError = action("removeError", (id: number) => {
-  state.errorAlerts = state.errorAlerts.filter((error) => error.id !== id);
+export const removeError = action("removeError", (id: number, text: string) => {
```
asivath, line 522: might be useful to have this return the generated id so that the component could have it without needing to access the errorAlerts map. I'm thinking this could be useful in Sejal's stage thing where a user inputs a valid stage combination and the error should just disappear instead of waiting there for the user to close it.

app/src/state/actions/general-actions.ts
```diff
@@ -509,14 +519,27 @@ export const handleCloseSnackbar = action(
...
+  state.errors.errorAlertIDs.delete(id);
+  state.errors.errorAlerts.delete(text);
 });
 export const pushError = action("pushError", (text: string) => {
```
asivath, line 526: might be useful to have this return the generated id so that the component could have it without needing to access the errorAlerts map. I'm thinking this could be useful in Sejal's stage thing where a user inputs a valid stage combination and the error should just disappear instead of waiting there for the user to close it.

asivath, line 526: might be useful to have this return the generated id so that the component could have it without needing to access the errorAlerts map. I'm thinking this could be useful in Sejal's stage thing where a user inputs a valid stage combination and the error should just disappear instead of waiting there for the user to close it.

asivath: this should be on pushError instead of removeError i can't fix it for some reason

app/src/state/actions/general-actions.ts
```diff
@@ -509,14 +519,27 @@ export const handleCloseSnackbar = action(
...
-  });
+  if (state.errors.errorAlerts.has(text)) {
+    state.errors.errorAlerts.get(text)!.errorCount += 1;
+    return;
+  }
+  let id = new Date().getTime();
+  // to prevent error alert id collisions on same millisecond errors
+  while (state.errors.errorAlertIDs.has(id)) {
+    id += Math.random();
+  }
```
asivath, line 535: this could be handled by a uuid library which would make collisions impossible

app/src/state/actions/general-actions.ts
paologum: so you're saying that when the user puts in an invalid, pushError, when they fix it, remove the error automatically?

asivath: Yea

jay-jhk-lee (approved):
looks good!

app/src/state/actions/general-actions.ts
paologum: i changed this so now it relies on a context where i've created a context "library" of errors to ensure no collisions in the long run. it also allows us to remove known errors easily given "third party" error resolutions

asivath (approved):

PR #54, opened 2024-03-15
fixed subdir on map points and adds column types for event and range (with all tests)

paologum:
Now we parse event and range types
I've abstracted the common column headers to `ColumnHeaderProps` so we wouldn't have duplicate code
will start work on other column types once Professor gets back on examples I can test on.


JacquiFreeElectron (approved):
looks good

SejalKumar03 (commented):
pretty good!

SejalKumar03 (approved):

server/__tests__/parse-datapacks.test.ts
```diff
@@ -198,77 +202,205 @@ describe("process blocks line tests", () => {
...
+  });
+  it("should return null on empty line", () => {
+    const line = "";
+    expect(processEvent(line)).toBeNull();
+  });
+  it("should throw error on NaN age", () => {
+    const line = "\tlabel\tbadNumber";
+    expect(() => processEvent(line)).toThrow();
+  });
+});
```
asivath, line 242: testing like this where you're testing the same function with different inputs can be simplified with .each

server/__tests__/parse-datapacks.test.ts
```diff
@@ -198,77 +202,205 @@ describe("process blocks line tests", () => {
...
+    const line3 = "\tlabel\t100\trare\tpopup";
+    expect(processRange(line3)).toEqual({ label: "label", age: 100, abundance: "rare", popup: "popup" });
+    const line4 = "\tlabel\t100\tcommon\tpopup";
+    expect(processRange(line4)).toEqual({ label: "label", age: 100, abundance: "common", popup: "popup" });
+  });
+  it("should return null on small line", () => {
+    const line = "\tlabel";
+    expect(processRange(line)).toBeNull();
+  });
+  it("should return null on large line", () => {
```
asivath, line 267: specify what is meant by a large line

server/__tests__/parse-datapacks.test.ts
```diff
@@ -198,77 +202,205 @@ describe("process blocks line tests", () => {
...
+      expectedEventMap.set(val, key["column-types-test-4-key"][val]);
+    }
+    expect(blockMap.size).toBe(0);
+    expect(faciesMap.size).toBe(0);
+    expect(eventMap.size).toBe(2);
+    expect(rangeMap.size).toBe(0);
+    expect(blockMap).toEqual(expectedBlockMap);
+    expect(faciesMap).toEqual(expectedFaciesMap);
+    expect(eventMap).toEqual(expectedEventMap);
+    expect(rangeMap).toEqual(expectedRangeMap);
```
asivath, line 375: this pattern is repeated a couple times, could be refactored into a function

shared/src/index.ts
```diff
@@ -373,6 +398,39 @@ export type TimescaleItem = {
...
+}
+export function assertColumnHeaderProps(o: any): asserts o is ColumnHeaderProps {
+  if (!o || typeof o !== "object") throw new Error("ColumnHeaderProps must be an object");
+  if (typeof o.name !== "string") throwError("ColumnHeaderProps", "name", "string", o.name);
+  if (typeof o.minAge !== "number") throwError("ColumnHeaderProps", "minAge", "number", o.minAge);
+  if (typeof o.maxAge !== "number") throwError("ColumnHeaderProps", "maxAge", "number", o.maxAge);
+  if (typeof o.enableTitle !== "boolean") throwError("ColumnHeaderProps", "enableTitle", "boolean", o.enableTitle);
+  if (typeof o.on !== "boolean") throwError("ColumnHeaderProps", "on", "boolean", o.on);
+  if (typeof o.width !== "number") throwError("ColumnHeaderProps", "width", "number", o.width);
+  if (typeof o.popup !== "string") throwError("ColumnHeaderProps", "popup", "string", o.popup);
```
asivath, line 431: idk if you want to do this but you could make a generic assert function to clean this up, something like this:
`function assertType(value: any, expectedType: string, obj: string, variable: string): void {
  if (typeof value !== expectedType) {
    throwError(obj, variable, expectedType, value);
  }
}`

shared/src/index.ts
```diff
@@ -383,6 +441,23 @@ export function assertRGB(o: any): asserts o is RGB {
...
+  if (!Array.isArray(o.subEventInfo)) throwError("Event", "subEventInfo", "array", o.subEventInfo);
+  for (const subEvent of o.subEventInfo) {
+    assertSubEventInfo(subEvent);
+  }
+  assertColumnHeaderProps(o);
+}
+export function assertSubEventInfo(o: any): asserts o is SubEventInfo {
+  if (!o || typeof o !== "object") throw new Error("SubEventInfo must be a non-null object");
+  if (typeof o.label !== "string") throwError("SubEventInfo", "label", "string", o.label);
+  if (typeof o.age !== "number") throwError("SubEventInfo", "age", "number", o.number);
```
asivath, line 455: should be o.age

shared/src/index.ts
```diff
@@ -383,6 +441,23 @@ export function assertRGB(o: any): asserts o is RGB {
...
+  if (typeof o.popup !== "string") throwError("SubEventInfo", "popup", "string", o.popup);
+  if (typeof o.lineStyle !== "string" && /dotted|dashed|solid/.test(o.lineStyle))
```
asivath, line 457: this check should be || right? also wrap the regex in ^ and & to ensure whole string gets matched

asivath, line 457: this check should be || right? also wrap the regex in ^ and & to ensure whole string gets matched

shared/src/index.ts
```diff
@@ -576,8 +643,22 @@ export function assertColumnInfo(o: any): asserts o is ColumnInfo {
...
+    for (const facies of o.subFaciesInfo) {
+      assertSubFaciesInfo(facies);
+    }
+  }
+  if ("subEventInfo" in o) {
+    if (!o.subEventInfo || !Array.isArray(o.subEventInfo))
+      throwError("ColumnInfo", "subEventInfo", "array", o.subEventInfo);
+    for (const event of o.subEventInfo) {
+      assertSubEventInfo(event);
+    }
```
asivath, line 655: these array checks could also be made into a generic function

server/src/parse-datapacks.ts
```diff
@@ -11,12 +11,21 @@ import {
...
+  Range,
+  ColumnHeaderProps,
+  SubRangeInfo,
+  assertSubRangeInfo
 } from "@tsconline/shared";
 import { trimQuotes, trimInvisibleCharacters, grabFilepaths, hasVisibleCharacters } from "./util.js";
 import { createInterface } from "readline";
 const patternForColor = /\d+\/\d+\/\d+/;
 const patternForLineStyle = /solid|dashed|dotted/;
+const patternForAbundance = /TOP|missing|rare|common|frequent|abundant|sample|flood/;
```
asivath, line 27: idk if the use case is the regex should match the entire string or only a part, but if the former wrap in ^ and $

server/__tests__/parse-datapacks.test.ts
asivath: you are right that since it's a testing file it should be as self documenting as possible, but code that is long and with lots of repeated code can be annoying/confusing to read as well. the helper function does not have to be one big function with tons of parameters you could split it up. for example you could have functions like this:
```
function expectMapSizes({ faciesSize, blockSize, eventSize, rangeSize }) {
  expect(faciesMap.size).toBe(faciesSize);
  expect(blockMap.size).toBe(blockSize);
  expect(eventMap.size).toBe(eventSize);
  expect(rangeMap.size).toBe(rangeSize);
}

function expectMapsEqual({ expectedFacies, expectedBlock, expectedEvent, expectedRange }) {
  expect(faciesMap).toEqual(expectedFacies);
  expect(blockMap).toEqual(expectedBlock);
  expect(eventMap).toEqual(expectedEvent);
  expect(rangeMap).toEqual(expectedRange);
}
```

and could be called like this:
```
expectMapSizes({ faciesSize: 2, blockSize: 0, eventSize: 0, rangeSize: 0 });
expectMapsEqual({ expectedFacies: expectedFaciesMap, expectedBlock: expectedBlockMap, expectedEvent: expectedEventMap, expectedRange: expectedRangeMap });
```

shared/src/index.ts
paologum: hmm maybe could do this but if we do it should be in a different pr. it could make things nicer but i think there's other things we should do first. I'll think about maybe doing this next week.

shared/src/index.ts
paologum: hmm i guess we could although i don't believe it's needed it could save some space. if i decide to do this it'll be in another pr but i'm leaning no

asivath (approved):

PR #55, opened 2024-03-16
Tsc 260 fix plugins make websocket if necessary

asivath:
Fixed Fastify plugins by doing a couple of things:

- Updated Fastify and Fastify-static to compatible versions
- Updated Typescript to 5.2.2 (the version Fastify is on) across all workspaces. Issue was caused by server being at 4.9.3.
- Packages were inconsistent across workspaces, so I added the syncpack package that automatically fixes these version issues

Other things I did:

- Added a clean script that deletes the dist folder to make sure that it builds completely from scratch
- Cleaned up all the package.jsons (moved all @types/, typescript, etc. to devDependencies)


package.json
```diff
@@ -1,37 +1,12 @@
...
-    "test": "jest -t",
-    "test-clear-cache": "jest --clearCache"
-  },
   "dependencies": {
-    "@types/node": "^20.11.24",
-    "fastify": "^4.13.0",
-    "fs": "^0.0.1-security",
-    "nvm": "^0.0.4",
-    "ts-node": "^10.9.2",
-    "typescript": "^5.0.4"
```
paologum, line 30: ok so you're removing these and you verify it all works okay with out like `fs` and `nvm`?
or is this syncpack

app/package.json
```diff
@@ -40,10 +31,20 @@
     "simplebar-react": "^3.2.4"
   },
   "devDependencies": {
+    "@types/color": "^3.0.6",
     "@types/react": "^18.0.27",
     "@types/react-dom": "^18.0.10",
     "@vitejs/plugin-react": "^3.1.0",
-    "typescript": "^4.9.3",
+    "typescript": "5.2.2",
```
paologum, line 38: you probably already checked but just verify all files are still fine on the dev side because i know when i change my typescript to not use the workspace but the ide's version of typescript, intellisense picks up a lot of problems. It might be because of the discrepancy between workspace and ide but still check

shared/package.json
```diff
@@ -1,23 +1,23 @@
...
-  "type": "module",
-  "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "build": "tsc -b",
-    "dev": "tsc -b -w"
+  "version": "1.0.0",
+  "author": "",
+  "dependencies": {
+    "@fastify/cors": "^9.0.1"
+  },
```
paologum, line 8: you might be able to just remove this because i don't think we use cors on the shared side

paologum (approved):
looks good just resolve my comments

package.json
asivath: I removed those packages myself. fs is already provided by Node and the one on npm is actually a placeholder package. Same with nvm, it's a placeholder package.

app/package.json
asivath: for the ide we need to make sure that people use the workspace version because the ide uses a version of typescript that will break everything. I initially tried using its version but it causes a lot of problems.

shared/package.json
asivath: ok ill remove it

PR #56, opened 2024-03-16
fixed age popup not showing

asivath:
image
image

Removed generateChart from general-actions.ts and made Home.tsx use initiateChartGeneration


paologum (changes requested):
this is more of a general comment but the way the popup shows right now feels really weird

can you format it similar to this

You can use the same buttons but add a thin border to the popup and format the buttons like that 

Also don't have the Age-Flag detected, just ask the simple Do you want to use the default age

image

asivath:
image
image

asivath:
image
image

paologum (approved):
looks great

PR #57, opened 2024-03-19
TSC-264-Limit-choices-for-top-base-based-on-the-other-top-base

SejalKumar03:
image
image


app/src/state/actions/column-actions.ts
```diff
@@ -64,6 +64,57 @@ export const setcolumnSelected = action((name: string) => {
...
+
+  const columnsToShow = searchResults.map(result => result.column.name);
+  
+  updateColumnVisibility(columnsToShow);
+
+  return searchResults;
+};
+
+
+
```
paologum, line 117: or this

app/src/SettingsTabs/Column.tsx
```diff
@@ -136,7 +138,8 @@ const ColumnAccordion: React.FC<ColumnAccordionProps> = observer(
                   key={childName}
                   details={childDetails}
                   expandedAccordions={expandedAccordions}
-                  accordionClicked={accordionClicked}
+                  accordionClicked={accordionClicked} 
+                  visibleColumns={[]}
```
paologum, line 142: same

app/src/SettingsTabs/Column.tsx
```diff
@@ -218,11 +254,22 @@ export const Column = observer(function Column() {
...
+  return <SearchComponent />;
 });
+
+function setSearchResults(arg0: {
+  // if there are no children, don't make an accordion
+  column: ColumnInfo; parentHierarchy: ColumnInfo[];
+}[]) {
+  throw new Error("Function not implemented.");
+}
+
```
paologum, line 275: same

app/src/SettingsTabs/Time.tsx
```diff
@@ -58,11 +64,22 @@ export const Time = observer(function Time() {
...
+                }
+                const correspondingBaseStage = state.geologicalBaseStageAges.find((item) => item.value >= age);
+                if (correspondingBaseStage) {
+                  actions.setSelectedBaseStage(correspondingBaseStage.key);
+                  actions.setBaseStageAge(Math.max(correspondingBaseStage.value, age));
+                }
+              }
+              else {
+                const errorMessage = "Invalid Top Age input. Please enter a valid age less than Base Age.";
                 actions.pushError(errorMessage);
```
paologum, line 83: should be context, you might need to rebase if this isn't throwing an error

app/src/SettingsTabs/Time.tsx
```diff
@@ -58,11 +64,22 @@ export const Time = observer(function Time() {
...
+                  actions.setSelectedTopStage(correspondingTopStage.key);
+                } else {
+                  actions.setSelectedTopStage("");
+                }
+                const correspondingBaseStage = state.geologicalBaseStageAges.find((item) => item.value >= age);
+                if (correspondingBaseStage) {
+                  actions.setSelectedBaseStage(correspondingBaseStage.key);
+                  actions.setBaseStageAge(Math.max(correspondingBaseStage.value, age));
+                }
+              }
```
paologum, line 80: same here as top above

app/src/SettingsTabs/Time.tsx
```diff
@@ -169,6 +194,11 @@ export const Time = observer(function Time() {
             control={<TSCCheckbox />}
             label="Hide block labels based on priority"
           />
+          <FormControlLabel
+            name="use-suggested-age-spans"
+            control={<TSCCheckbox />}
+            label="Hide block labels based on priority"
+          />
```
paologum, line 201: hmmm what is this doesn't this already exist

shared/dist/index.d.ts
```diff
@@ -215,6 +215,7 @@ export type ColumnInfo = {
     subFaciesInfo?: SubFaciesInfo[];
     minAge: number;
     maxAge: number;
+    show: boolean
```
paologum, line 218: not this pr

paologum (changes requested):

app/src/SettingsTabs/Time.tsx
```diff
@@ -37,12 +37,14 @@ export const Time = observer(function Time() {
...
-                actions.pushError(ErrorCodes.TOP_STAGE_AGE_INVALID);
+              actions.setSelectedTopStage(selectedValue);
+              actions.setTopStageAge(selectedAge);
+              if (selectedAge > state.settings.baseStageAge) {
+                const correspondingBaseStage = state.geologicalBaseStageAges.find((item) => item.value >= selectedAge);
+                if (correspondingBaseStage) {
+                  actions.setSelectedBaseStage(correspondingBaseStage.key);
+                  actions.setBaseStageAge(Math.max(correspondingBaseStage.value, selectedAge));
+                }
               }
```
paologum, line 48: i was thinking about this and i think it'd make more sense to just make one method `actions.setTopStageAge` that does all this (sets both the text and number fields in state)

that way here and anywhere else you can just call the `actions` method with a number and it will throw the proper errors/set properly

app/src/SettingsTabs/Time.tsx
```diff
@@ -59,12 +61,22 @@ export const Time = observer(function Time() {
...
+                  actions.setSelectedTopStage(correspondingTopStage.key);
+                } else {
+                  actions.setSelectedTopStage("");
+                }
+                const correspondingBaseStage = state.geologicalBaseStageAges.find((item) => item.value >= age);
+                if (correspondingBaseStage) {
+                  actions.setSelectedBaseStage(correspondingBaseStage.key);
+                  actions.setBaseStageAge(Math.max(correspondingBaseStage.value, age));
+                }
+              }
```
paologum, line 77: basically you can just call the method from above

app/src/SettingsTabs/Time.tsx
paologum: for example you should just call `setTopStageAge` and it should set the text field only if there is a value EQUAL to the value you passed into the method and then it should also set the `baseStage` appropriately (number) and then if the number that `baseStage` increases to has an equivalent entry then it should set the text field too, otherwise the textfield should be blank

currently in your implementation it checks `>= selectedAge` which it should be `== selectedAge`

app/src/SettingsTabs/Time.tsx
```diff
@@ -103,11 +117,17 @@ export const Time = observer(function Time() {
...
                 actions.removeError(ErrorCodes.BASE_STAGE_AGE_INVALID);
-              } else {
+                if (correspondingBaseStage) {
+                  actions.setSelectedBaseStage(correspondingBaseStage.key);
+                } else {
+                  actions.setSelectedBaseStage("");
+                }
+              } 
+              else {
                 actions.pushError(ErrorCodes.BASE_STAGE_AGE_INVALID);
```
paologum, line 131: same as top (in the context of making one call to `setBaseAge` that changes both text field and number) but in this case we don't want to change `topAge`

shared/dist/index.d.ts
paologum, line 1: remove this file, if you rebase it should throw an error in the pipeline that this file exists

paologum (changes requested):

app/src/SettingsTabs/Time.tsx
```diff
@@ -37,13 +37,8 @@ export const Time = observer(function Time() {
...
               const selectedAge = selectedAgeItem?.value || 0;
-              if (selectedAge >= 0 && selectedAge <= state.settings.baseStageAge) {
-                actions.setSelectedTopStage(selectedValue);
-                actions.setTopStageAge(selectedAge);
-                actions.removeError(ErrorCodes.TOP_STAGE_AGE_INVALID);
-              } else {
-                actions.pushError(ErrorCodes.TOP_STAGE_AGE_INVALID);
-              }
+              actions.setSelectedTopStage(selectedValue);
+              actions.setTopStageAge(selectedAge);
```
paologum, line 41: you don't use `selectedAgeItem` also `setSelectedTopStage` should be in `setTopStageAge` ONLY if there is one that exists, otherwise `""` or whatever nothing is i lk forget

app/src/SettingsTabs/Time.tsx
```diff
@@ -59,12 +54,11 @@ export const Time = observer(function Time() {
             value={state.settings.topStageAge.toString()}
             onChange={(event) => {
               const age = parseFloat(event.target.value);
-              if (!isNaN(age) && age >= 0 && age <= state.settings.baseStageAge) {
-                actions.setSelectedTopStage("");
+              if (!isNaN(age)) {
                 actions.setTopStageAge(age);
+              }
+              else {
                 actions.removeError(ErrorCodes.TOP_STAGE_AGE_INVALID);
```
paologum, line 61: you should `removeError` on `!isNaN` and otherwise `pushError`

app/src/SettingsTabs/Time.tsx
```diff
@@ -81,19 +75,21 @@ export const Time = observer(function Time() {
...
               const selectedAgeItem = state.geologicalBaseStageAges.find((item) => item.key === selectedValue);
               const selectedAge = selectedAgeItem?.value || 0;
-              if (selectedAge >= 0 && selectedAge >= state.settings.topStageAge) {
+              if (selectedAge > state.settings.topStageAge) {
                 actions.setSelectedBaseStage(selectedValue);
                 actions.setBaseStageAge(selectedAge);
                 actions.removeError(ErrorCodes.BASE_STAGE_AGE_INVALID);
               } else {
                 actions.pushError(ErrorCodes.BASE_STAGE_AGE_INVALID);
               }
```
paologum, line 84: you should have these errors and all the methods, and the `if` checks within the `setBaseStageAge` method

paologum: on success inside `setBaseStageAge` it should always `removeError`

app/src/SettingsTabs/Time.tsx
```diff
@@ -103,11 +99,10 @@ export const Time = observer(function Time() {
...
-              if (!isNaN(age) && age >= 0 && state.settings.topStageAge <= age) {
-                actions.setSelectedBaseStage("");
+              if (!isNaN(age) && age >= state.settings.topStageAge) {
                 actions.setBaseStageAge(age);
-                actions.removeError(ErrorCodes.BASE_STAGE_AGE_INVALID);
-              } else {
+              } 
+              else {
                 actions.pushError(ErrorCodes.BASE_STAGE_AGE_INVALID);
               }
```
paologum, line 107: `actions.removeError` is removed here but it should always be removed at the end of the `setBaseStageAge` method (technically not here so that's fine but when it sets it properly in the `actions.setBaseStageAge` method it should remove it there).

app/src/state/actions/general-actions.ts
```diff
@@ -1,5 +1,6 @@
 import { action, runInAction } from "mobx";
 import { TimescaleItem } from "@tsconline/shared";
+import { actions } from "../index";
```
paologum, line 3: don't import actions

app/src/state/actions/general-actions.ts
```diff
@@ -590,9 +591,28 @@ export const setChartHash = action("setChartHash", (charthash: string) => {
...
+      actions.setSelectedTopStage(correspondingTopStage.key);
+    } else {
+      actions.setSelectedTopStage("");
+    }
+
+    const correspondingBaseStage = state.geologicalBaseStageAges.find((item) => item.value >= age);
+    if (correspondingBaseStage) {
+      actions.setSelectedBaseStage(correspondingBaseStage.key);
+      actions.setBaseStageAge(Math.max(correspondingBaseStage.value, age));
+    }
```
paologum, line 605: with this, you are basically making the `topStageAge` and `baseStageAge` equal, we don't want that. we want the gap that existed between them initially `baseStageAge - topStageAge` to be preserved

paologum (changes requested):
do you mind rebasing so that the main changes don't show on your pr. it just makes it easier to read haha

PR #58, opened 2024-03-20
implement all column types

paologum:
sequence/trend, chron, transect, blank, freehand, graph/point all implemented

now processes header and min/max age for these


JacquiFreeElectron (commented):
looks great

server/src/parse-datapacks.ts
```diff
@@ -103,18 +125,52 @@ export async function parseDatapacks(decryptFilePath: string, files: string[]):
...
         // if the parent is not a child
         if (!isChild.has(parent)) {
-          recursive("Root", parent, children, columnInfoArray, allEntries, faciesMap, blocksMap, eventMap, rangeMap);
+          recursive(
+            "Root",
+            parent,
+            children,
+            columnInfoArray,
+            allEntries,
+            faciesMap,
```
jay-jhk-lee, line 163: Do you have to pass all the maps every recursive call? Could there be a way to abstract it out?

asivath (approved):

server/src/parse-datapacks.ts
paologum: not really unless i change this whole file to be an object/class that has it's own properties that are the maps but that would mean changing a good amount of this file and other files to refactor things around which would warrant it being it's own task (which i'll add on the backlog). I could see this as a future refactoring of the `parse-datapacks.ts` file but for now I think this is alright.

it would more or less be referenced like a python constructor with like `self.allEntires` `self.faciesMap` and you would just change that in each instance which makes sense so yes this is a good idea

jay-jhk-lee:
Looks good!

PR #61, opened 2024-03-21
Tsc 253 upgrade to mui 5

asivath:
Removed MUI packages that are used for MUI 4 as well other packages that are not used. All the changes in the files were done by an automatic tool called codemod to upgrade from MUI 4 to 5.


app/src/SettingsTabs/Legend.tsx
```diff
@@ -189,7 +190,7 @@ const FilterMenu: React.FC<FilterMenuProps> = observer(
     const { state } = useContext(context);
     return (
       <>
-        <IconButton ref={menuRef} className="filter-button" {...anchorProps}>
+        <IconButton ref={menuRef} className="filter-button" {...anchorProps} size="large">
```
paologum, line 193: are you manually adding these? i'm interested to know why

app/src/App.tsx
```diff
@@ -21,48 +21,50 @@ export default observer(function App() {
...
-        <Route path="/about" element={<About />} />
-      </Routes>
-      {Array.from(state.errors.errorAlerts.entries()).map(([context, error], index) => (
-        <TSCError
-          key={context}
-          errorContext={context}
-          message={error.errorText}
-          index={index}
-          count={error.errorCount}
+    <StyledEngineProvider injectFirst>
```
paologum, line 24: what exactly does this do

paologum (approved):
just some questions i have, no changes

app/src/SettingsTabs/Legend.tsx
asivath: this was done automatically but it's because mui 5 changes some padding stuff and size large is the same size as mui 4

app/src/App.tsx
asivath: mui 5 is moving away from JSS for Emotion, but this gives backwards compatibility for JSS. It basically tells the styling engine to let JSS styles override MUI styles

asivath: they do recommend moving away from JSS to newer components should probably not use it

PR #62, opened 2024-03-27
upload datapack framework

paologum:
based on a username that we hash, upload a datapack to that directory and create a json cache that the app can service when logging in new users.
Two main routes `/user-datapacks/:username/` and `/upload/:username`

First route grabs any datapacks that have been parsed in the users directory
Second route uploads a datapack, decrypts, parses and throws any error (with cleanup) with only one file.

TODO: sunset files and login functionality


server/src/routes.ts
```diff
@@ -1,15 +1,177 @@
...
+  const datapackDir = path.join(userDir, "datapacks");
+  const decryptDir = path.join(userDir, "decrypted");
+  const filepath = path.join(datapackDir, filename);
+  const decryptedFilepathDir = path.join(decryptDir, filenameWithoutExtension);
+  const mapPackIndexFilepath = path.join(userDir, "MapPackIndex.json");
+  const datapackIndexFilepath = path.join(userDir, "DatapackIndex.json");
+  if (!/^(\.dpk|\.txt|\.map|\.mdpk)$/.test(ext)) {
+    reply.status(415).send({ error: "Invalid file type" });
+    return;
+  }
```
asivath, line 81: should also check MIME type since extensions can easily be faked

server/src/routes.ts
```diff
@@ -1,15 +1,177 @@
...
+  if (!fs.existsSync(datapackDir)) {
+    fs.mkdirSync(datapackDir, { recursive: true });
+  }
+  const fileStream = file.file;
+  console.log("Uploading file: ", filename);
+  pump(fileStream, fs.createWriteStream(filepath));
+  try {
+    // must wait for the file to be written before decrypting
+    await new Promise<void>((resolve, reject) => {
+      pump(fileStream, fs.createWriteStream(filepath), (err) => {
```
asivath, line 91: two pumps?

server/src/routes.ts
```diff
@@ -1,15 +1,177 @@
...
+  } catch (e) {
+    resetUploadDirectory(filepath, decryptedFilepathDir);
+    reply.status(500).send({ error: "Failed to save file with error: " + e });
+    return;
+  }
+  if (file.file.truncated) {
+    resetUploadDirectory(filepath, decryptedFilepathDir);
+    reply.status(413).send({ error: "File too large" });
+    return;
+  }
```
asivath, line 108: this should happen before uploading

server/src/routes.ts
```diff
@@ -1,15 +1,177 @@
...
+  try {
+    const cmd =
+      `java -jar ${assetconfigs.decryptionJar} ` +
+      // Decrypting these datapacks:
+      `-d "${filepath}" ` +
+      // Tell it where to send the datapacks
+      `-dest ${decryptDir} `;
+    console.log("Calling Java decrypt.jar: ", cmd);
+    execSync(cmd, { stdio: "inherit" });
```
asivath, line 117: why not use async/await? Same with the other file operations

server/src/routes.ts
```diff
@@ -1,15 +1,177 @@
...
+  if (fs.existsSync(path.join(userDir, "DatapackIndex.json"))) {
+    Object.assign(datapackIndex, JSON.parse(fs.readFileSync(path.join(userDir, "DatapackIndex.json")).toString()));
+  }
+  const mapPackIndex: MapPackIndex = {};
+  if (fs.existsSync(path.join(userDir, "MapPackIndex.json"))) {
+    Object.assign(mapPackIndex, JSON.parse(fs.readFileSync(path.join(userDir, "MapPackIndex.json")).toString()));
+  }
+  const indexResponse = { datapackIndex, mapPackIndex };
+  assertIndexResponse(indexResponse);
+  reply.status(200).send(indexResponse);
```
asivath, line 50: error handling for readFileSync or JSON.parse faling

server/src/routes.ts
paologum: this is how `@fastify/multipart` is implemented, i tried looking into finding the size before but haven't seen anything that throws when the `fileSize` indicated in the options is exceeded until this bit of code catches it. what it does is read the file in to the limit and stops and sets truncate to true. we read the file through a buffer so i don't think that should cause much of a problem with overloading requests.

JacquiFreeElectron (commented):
Looks good

server/src/routes.ts
paologum: i can only really verify `application/octet-stream` and `text/plain` but i added them

asivath (approved):
looks good

PR #63, opened 2024-03-27
Tsc 229 style popup and show popup

asivath:
Added popups and also allowed panning/zooming on the SVG. The reason the SVG needs to be embedded into the DOM rather than within an object tag is because of the CORS policies. If it's in an object tag and because the chart comes from a different source than the app, the policy will block a lot of its interactivity. Directly embedding is a bit unsafe so I verify where the SVG comes from and also sanitize the SVG.

https://github.com/earthhistoryviz/tsconline/assets/94872897/a081dd55-a9cb-47ed-850d-0bcd2af9c813

https://github.com/earthhistoryviz/tsconline/assets/94872897/e993dd9f-20d4-4cf8-aa81-9239c72b5aaa


app/src/components/TSCSvgComponent.tsx
```diff
@@ -0,0 +1,117 @@
...
+        setSvgContent(sanitizedSVG);
+      })
+      .catch(() => {
+        actions.pushError(ErrorCodes.INVALID_SVG_RESPONSE);
+        setIsValidPath(false);
+      });
+  }, [path]);
+
+  useEffect(() => {
+    if (!state.settings.mouseOverPopupsEnabled) return;
```
paologum, line 47: does this still have the weird functionality where it has to be clicked twice to turn on between calls

app/src/components/TSCSvgComponent.tsx
```diff
@@ -0,0 +1,117 @@
...
+    let url: string = "http://localhost:3000";
+    if (import.meta.env.VITE_SERVER_URL) {
+      url = import.meta.env.VITE_SERVER_URL;
+    }
+    if (!path.startsWith(url)) {
+      actions.pushError(ErrorCodes.INVALID_SVG_PATH);
+      setIsValidPath(false);
+      return;
+    }
+    fetch(path)
```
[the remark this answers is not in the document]
paologum: change the error as well to a general error code for invalid paths

app/src/components/TSCPopupManager.tsx
```diff
@@ -0,0 +1,54 @@
...
+      }
+    };
+
+    window.addEventListener("message", handleMessage);
+
+    return () => window.removeEventListener("message", handleMessage);
+  }, []);
+
+  if (!isValidPath) {
+    return <div>Invalid SVG Path. Please check the path and try again.</div>;
```
paologum, line 39: make this `Typography`

app/src/components/TSCSvgComponent.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    return () => {
+      container.removeEventListener("mouseover", handleSvgEvent);
+      container.removeEventListener("mouseout", handleSvgEvent);
+      container.removeEventListener("click", handleSvgEvent);
+    };
+  }, [svgContent, state.settings.mouseOverPopupsEnabled]);
+
+  if (!isValidPath) {
+    return <div>Invalid SVG Path. Please check the path and try again.</div>;
```
paologum, line 113: same as above

paologum (changes requested):
some comments

app/src/components/TSCSvgComponent.tsx
asivath: ya nothing resets them between calls

paologum: can you fully describe what's happening or should happen because i'm a little confused. a video of you doing it would also be helpful. i think i might know but i'm not sure

asivath: https://github.com/earthhistoryviz/tsconline/assets/94872897/c090c18f-209f-4ee5-9337-c653d8af86e2

if you want to turn the popups back off you need to click twice

asivath:
image

app/src/components/TSCSvgComponent.tsx
paologum: can you add in the pr line 140 in `Time.tsx` to add `checked={state....mouseOverPopup}` in `TSCCheckbox`

server/src/routes.ts
```diff
@@ -191,6 +191,25 @@ export const uploadDatapack = async function uploadDatapack(
...
+    await access(imagePath);
+    const image = await readFile(imagePath);
+    reply.send(image);
+  } catch (e) {
+    const error = e as NodeJS.ErrnoException;
+    if (error.code === "ENOENT") {
+      reply.status(404).send({ error: "Image not found" });
+    } else {
+      reply.status(500).send({ error: "An error occurred" });
+    }
```
paologum, line 209: in the future, you will also have to check the user images in their respective `upload` directory.
you don't have to do that until the framework is fully setup with login (comment only)

server/src/routes.ts
```diff
@@ -191,6 +191,25 @@ export const uploadDatapack = async function uploadDatapack(
   reply.status(200).send({ message: "File uploaded" });
 };
 
+export const fetchImage = async function (
+  request: FastifyRequest<{ Params: { datapackName: string; imageName: string } }>,
+  reply: FastifyReply
+) {
+  const imagePath = `${assetconfigs.decryptionDirectory}/${request.params.datapackName}/datapack-images/${request.params.imageName}`;
```
paologum, line 198: you should use `path.join` here just in case

app/src/components/TSCSvgComponent.tsx
```diff
@@ -0,0 +1,83 @@
...
+              }
+              break;
+            }
+          }
+          break;
+        }
+        currentElement = currentElement.parentElement;
+      }
+    };
+
```
paologum, line 70: besides the event listener settings, can't this exist as it's own function? still in the component but outside of the `useEffect`? both `chartContent` and `state.settings.mousOverPopupsEnabled` will change due to observer and reload the component properly so if you just put the return stuff inside the handler it should be the same and be a little more abstracted

paologum (changes requested):

paologum:
> image

can you make the close an x in the top right?

asivath:
image

app/src/components/TSCPopupManager.tsx
```diff
@@ -0,0 +1,97 @@
...
+              await new Promise((resolve, reject) => {
+                reader.onloadend = resolve;
+                reader.onerror = reject;
+              });
+              return { fullMatch, dataUrl: reader.result };
+            }
+          } catch (error) {
+            console.error("Error fetching or processing image:", error);
+            return { fullMatch, dataUrl: "" };
+          }
```
paologum, line 41: do you think we shouldn't `pushError` here? or in `actions.fetchImage`?

paologum (approved):
one question/comment

app/src/components/TSCPopupManager.tsx
asivath: should we show errors to the user that they can't do anything about?

asivath: acutally user datapacks can have images right?

paologum: yeah they can/should, but also we show server errors which technically they can't do anything about

PR #64, opened 2024-03-28
manually append facies columns

paologum:
this is the comment about the "hack"
```
/**
 * Since we hash by name only to allow consistency between facies maps and
 * the column page, a generic like Facies Label will cause errors.
 * The solution @Paolo came up with is to prepend the name of the parent
 * and change before the conversion to xml. The downside is we must check
 * every ColumnInfo object which may cause problems with time consistency.
 * However, this is asyncronous, which makes it less likely to cause problems.
 * @param column
 */
```
Screenshot 2024-03-27 at 9 09 48 PM

java file ^

https://github.com/earthhistoryviz/tsconline/assets/89664317/5ca55d3e-1c7b-4835-a9b1-bd649467f08b


lbostre (approved):
Looks good!

asivath (approved):

PR #67, opened 2024-03-30
Tsc 236 add general snackbar

JacquiFreeElectron:
Screenshot 2024-03-29 at 11 59 10 PM
Screenshot 2024-03-29 at 11 59 21 PM
Screenshot 2024-03-29 at 11 58 20 PM


app/src/App.css
```diff
@@ -1,12 +1,14 @@
-.MuiAlert-filled.success-alert {
+.MuiAlert-filled.snackbar {
   align-items: center;
-  width: 100%;
+  /*width: 100%;*/
```
paologum, line 3: remove this comment if you don't need it

app/src/assets/icons/info-icon.json
paologum, line 1: just curious what website did you grab this from

app/src/state/actions/general-actions.ts
```diff
@@ -558,6 +549,28 @@ export const pushError = action("pushError", (context: ErrorCodes) => {
...
+  for (const snackbar of state.snackbars) {
+    if (snackbar.snackbarText === text) {
+      snackbar.snackbarCount += 1;
+      return;
+    }
+  }
+  if (severity != "success" && severity != "info") {
+    console.log("severity of the snackbar must be either 'success' or 'info'");
+    return;
+  }
```
paologum, line 566: instead of checking the severity with an if, just make the parameter `severity: "success" | "info"`

app/src/state/actions/general-actions.ts
```diff
@@ -651,9 +664,9 @@ export const setBaseStageAge = action("setBaseStageAge", (age: number) => {
 export const settingsXML = action("settingsXML", (xml: string) => {
   state.settingsXML = xml;
 });
-export const setOpenSnackbar = action("setOpenSnackbar", (show: boolean) => {
+/* export const setOpenSnackbar = action("setOpenSnackbar", (show: boolean) => {
   state.openSnackbar = show;
-});
+}); */
```
paologum, line 669: remove

app/src/state/actions/util-actions.ts
```diff
@@ -19,3 +19,11 @@ export function displayServerError<T>(response: T | null, context: ErrorCodes, m
...
   }
 }
+
+/**
+ * Display snackbar
+ * @param message the message to be shown
+ */
+export function displaySnackbar(message: string, severity: "success" | "info") {
+  pushSnackbar(message, severity);
+}
```
paologum, line 29: no need for this, the `displayError` you're probably basing this on is specifically for server responses that throw errors where it requires a specific method to print out the server response. you can just use `pushSnackbar`

app/src/state/actions/general-actions.ts
```diff
@@ -558,6 +549,28 @@ export const pushError = action("pushError", (context: ErrorCodes) => {
...
+  if (severity != "success" && severity != "info") {
+    console.log("severity of the snackbar must be either 'success' or 'info'");
+    return;
+  }
+  state.snackbars.push({
+    snackbarText: text,
+    snackbarCount: 1,
+    severity: severity
+  });
+  state.openSnackbar = true;
```
paologum, line 572: you can deprecate `state.openSnackbar` and just make all snackbars that are pushed default to `show=true`. if they exist, they should always be shown

app/src/components/TSCSnackbar.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+  const bgColor = severity === "info" ? theme.palette.snackbarAlert.light : theme.palette.snackbarAlert.main;
+  function handleCloseSnackbar(_event: React.SyntheticEvent | Event, reason?: string) {
+    if (reason === "clickaway") return;
+
+    actions.removeSnackbar(text);
+  }
+  return (
+    <Snackbar
+      open={state.openSnackbar}
```
paologum, line 34: should just be true

paologum (changes requested):
last thing, can you show me/upload what happens when the text overflows in the alert?
also the photo provided has the text slightly miscentered. You might need to `justify-content: center, align-items: center` for the text to be aligned

app/src/assets/icons/info-icon.json
JacquiFreeElectron: LottieFiles

app/src/state/actions/general-actions.ts
JacquiFreeElectron: I tried this way, but there is always one snackbar that never goes away. I can try it again, but if it's not working, can I keep `state.openSnackb`ar?

paologum: hmm that technically means that there exists a snackbar within the array that sticks around. i would try debugging and stuff because the array should not exist/have any contents after 5 seconds of an info/success bar getting pushed.

JacquiFreeElectron: okk

JacquiFreeElectron:
Screenshot 2024-04-01 at 12 27 44 PM
Screenshot 2024-04-01 at 12 27 14 PM

JacquiFreeElectron:
https://github.com/earthhistoryviz/tsconline/assets/104879321/e20651a5-cd51-47e1-b590-3cd7e2cbfc9b

app/src/components/TSCSnackbar.css
```diff
@@ -0,0 +1,13 @@
+:root {
+  --snackbar-width: 300px;
+  --snackbar-height: 60px;
+}
+
+.snackabr {
```
paologum, line 6: hmm this should probably be snackbar

app/src/components/TSCSnackbar.tsx
```diff
@@ -0,0 +1,60 @@
...
+      autoHideDuration={5000}
+      TransitionComponent={Slide}>
+      <Alert
+        style={{
+          backgroundColor: bgColor
+        }}
+        severity={severity}
+        variant="outlined"
+        color={severity}
+        className="snackabr"
```
paologum, line 49: same

app/src/App.css
```diff
@@ -1,12 +1,9 @@
...
+.MuiAlert-root.snackbar .MuiAlert-icon {
   display: flex;
   align-items: center;
 }
-.MuiAlert-root.success-alert .MuiAlert-message {
+
+.MuiAlert-root.snackbar .MuiAlert-message {
   display: flex;
   align-items: center;
   flex-grow: 1;
```
paologum, line 9: probably move this to TSCSnackbar.css since the actual snackbar class is there. also snackabr was used so these weren't in effect so just verify it's still ok

paologum (changes requested):

app/src/components/TSCSnackbar.css
JacquiFreeElectron: 😂right

paologum (approved):
looks great :D

PR #68, opened 2024-03-31
ui for datapack form upload

paologum:
https://github.com/earthhistoryviz/tsconline/assets/89664317/a7a0ae16-5aeb-458f-8174-7e23933babb9


app/src/components/TSCDatapackUploadForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+                actions.pushError(ErrorCodes.NO_DATAPACK_FILE_FOUND);
+                return;
+              }
+              if (!datapackName || !datapackDescription) {
+                actions.pushError(ErrorCodes.UNFINISHED_DATAPACK_UPLOAD_FORM);
+                return;
+              }
+              actions.removeError(ErrorCodes.NO_DATAPACK_FILE_FOUND);
+              actions.removeError(ErrorCodes.UNFINISHED_DATAPACK_UPLOAD_FORM);
+              actions.uploadDatapack(datapackFile!, "username");
```
asivath, line 109: there should be user feedback if upload is successful like the successful chart generation lottie. also clearing the form is another thing

asivath (commented):
just one comment

app/src/components/TSCDatapackUploadForm.tsx
paologum: jacqui's pr isn't merged yet so that's gonna be added in later. what do you mean by clearing the form

asivath: settings the fields empty

paologum: i mean they can exit and come back? i don't necessarily think we need a button to clear itt

asivath: you dont need a different button just clear it when they submit. my first instinct when seeing that the text wasn't cleared was that something was wrong with the form data since usually when you submit a form successfully it clears the fields for the next submission. exiting and coming back seems like a workaround for clearing it. if you think otherwise you can just resolve this.

paologum: i think, that the upload can get cancelled in many ways once they say finish & upload, so it shouldn't clear because it's asynchronous and the app shouldn't wait for a successful response from the server to clear the fields. I'll add a clear/reset button so it's easier to upload multiple data packs in a row however. I might also make multiple file uploads possible but that seems very complicated so it'll be on the backlog.

with that in mind lmk if that works with you. (it'll be more clear that it went through successfully when jacqui's pr gets through)

paologum:
https://github.com/earthhistoryviz/tsconline/assets/89664317/dee8f7a5-b0af-4f58-8dda-28f42c3685d2

asivath (approved):

PR #69, opened 2024-04-01
Tsc 126 column width

lbostre:
- added column menu option to change width on chart generation

867ca8e60b8da8b70d00e29c31057e97

DEMO
https://github.com/earthhistoryviz/tsconline/assets/29764221/98e7eb58-4689-4f06-9b72-db4a9f8f6636


app/src/SettingsTabs/EditWidthField.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { state, actions } = useContext(context);
+  const editWidth =
+    state.settingsTabs.columnSelected === null
+      ? 0
+      : state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.width;
+  const [width, setWidth] = useState<number>(editWidth ?? 0);
+  return (
+    <div>
+      <Typography style={{ padding: "5px" }}>Edit Width</Typography>
+      <div style={{ display: "flex", flexDirection: "row" }}>
```
paologum, line 22: can you move this to css

app/src/SettingsTabs/EditWidthField.tsx
```diff
@@ -0,0 +1,48 @@
...
+const WidthTextField = ({ ...props }: TextFieldProps) => (
+  <TextField {...props} hiddenLabel variant="filled" size="small" />
+);
+
+export const EditWidthField = observer(() => {
+  const { state, actions } = useContext(context);
+  const editWidth =
+    state.settingsTabs.columnSelected === null
+      ? 0
+      : state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.width;
```
paologum, line 17: i think this would make more sense from an abstractability standpoint to just pass the width in as a parameter of `EditWidthField`. You can do the check somewhere in the parent component which makes this more understandable/possibly usable elsewhere

app/src/state/actions/column-actions.ts
```diff
@@ -58,6 +58,23 @@ export const updateEditName = action((newName: string) => {
...
+    return;
+  }
+  if (!state.settingsTabs.columnHashMap.has(state.settingsTabs.columnSelected)) {
+    console.log(
+      "WARNING: tried to access",
+      state.settingsTabs.columnSelected,
+      "in state.settingsTabs.columnHashMap, but map does not this key"
+    );
+  }
+  if (isNaN(newWidth)) return;
```
paologum, line 73: if `updateWidth` accepts only `newWidth: number` this isn't needed

app/src/SettingsTabs/EditWidthField.tsx
```diff
@@ -0,0 +1,48 @@
...
+          }}
+          style={{ height: "20px" }}
+        />
+        <div>
+          <Button
+            color="secondary"
+            variant="contained"
+            onClick={() => {
+              actions.updateWidth(width);
+            }}>
```
paologum, line 41: i know numeric format should handle everything properly, but i would add the `isNaN` here to make sure and to also `pushError` if it isn't.

paologum (changes requested):
lowkey forgot to review this so sorry it was late

app/src/SettingsTabs/EditWidthField.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <Button
+            color="secondary"
+            variant="contained"
+            onClick={() => {
+              if (isNaN(inputWidth)) pushError(ErrorCodes.INVALID_WIDTH_TYPE);
+              actions.updateWidth(inputWidth);
+            }}>
+            Confirm
+          </Button>
+        </div>
```
paologum, line 45: the more i think about this the more i think that we shouldn't have a confirm button and it should just update on change of `NumericFormat`

if the user never presses confirm all the columns will have weird numbers they might not expect and also more importantly the java file doesn't have a confirm

you shouldn't need the error anymore and if the user doesn't have any number or the field is empty, the width should set to 20 (the java file does this)

app/src/state/actions/column-actions.ts
```diff
@@ -58,6 +58,22 @@ export const updateEditName = action((newName: string) => {
...
+    console.log(
+      "WARNING: tried to access",
+      state.settingsTabs.columnSelected,
+      "in state.settingsTabs.columnHashMap, but map does not this key"
+    );
+  }
+  state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.width = newWidth;
+  return;
+});
+
```
paologum, line 76: i feel like this makes too many assumptions about the structure, and if we every decide to change how we do `columnSelected` or what not it would require a lot of overhead changes that we would need to sort out.

so instead you should just pass in the object and update the width (2 params). it should also check if the column has the width object and console error if it doesn't have a width

paologum (changes requested):
update a new picture when you remove the confirm button

app/src/SettingsTabs/ColumnMenu.tsx
```diff
@@ -87,6 +88,12 @@ export const ColumnMenu = observer(() => {
       >
         {state.settingsTabs.columnSelected && <EditNameField />}
         {state.settingsTabs.columnSelected && <FontMenu />}
+        {state.settingsTabs.columnSelected &&
+          state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.width && (
+            <EditWidthField
+              width={state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.width ?? 0}
+            />
+          )}
```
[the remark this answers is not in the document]
lbostre: For some reason, if I change the code to `state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected) && ..`. rather than the ```
state.settingsTabs.columnSelected &&
          state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.width &&
```
the width value is not passed in properly.

File: app/src/state/actions/column-actions.ts [hunk 8]
```diff
@@ -58,6 +58,31 @@ export const updateEditName = action((newName: string) => {
...
+    console.error(
+      "ERROR: Tried to update width for column",
+      state.settingsTabs.columnSelected,
+      "but width attribute is missing."
+    );
+    return;
+  }
+  selectedColumn.width = newWidth;
+  return;
+});
```
paologum, line 84: this should just do `columnObject.width = newWidth`

app/src/SettingsTabs/EditWidthField.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const EditWidthField: React.FC<{
+  width: number | undefined;
+}> = observer(({ width = 20 }) => {
+  const { state, actions } = useContext(context);
+  const [inputWidth, setInputWidth] = useState<number>(width);
+  let columnObject: ColumnInfo | undefined;
+  if (state.settingsTabs.columnSelected != null) {
+    columnObject = state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected);
+  }
```
paologum, line 21: if it doesn't exist just return null, `columnObject` should always exist if this edit component is rendered

app/src/SettingsTabs/EditWidthField.tsx
```diff
@@ -0,0 +1,43 @@
...
+        onValueChange={(values) => {
+          const floatValue = values.floatValue;
+          if (floatValue === undefined && columnObject) {
+            actions.updateWidth(columnObject, 20);
+            return;
+          }
+          if (floatValue !== undefined) {
+            setInputWidth(floatValue);
+            if (columnObject) actions.updateWidth(columnObject, floatValue);
+          }
```
paologum, line 37: is there a reason you specifically use undefined vs just using truthy values

app/src/state/parse-settings.ts
```diff
@@ -409,6 +409,20 @@ function generateColumnXml(presetColumn: ColumnInfoTSC, stateColumn: ColumnInfo
...
+        let useEditWidth = false;
+        if (stateColumn && stateColumn !== undefined) {
+          if (stateColumn.width !== undefined) {
+            xml += `${indent}<setting name="width">${stateColumn.width}</setting>\n`;
+            useEditWidth = true;
+          }
+        }
+        if (!useEditWidth) {
+          xml += `${indent}<setting name="width">${presetColumn[key]}</setting>\n`;
+        }
```
paologum, line 423: can you just use `stateColumn`? or does it need `presetColumn`?

our `ColumnInfo` state object is more correct so i feel like we don't need `presetColumn` unless i'm missing something

app/src/util/error-codes.ts
```diff
@@ -19,7 +19,8 @@ export enum ErrorCodes {
   UNRECOGNIZED_DATAPACK_FILE = "UNRECOGNIZED_DATAPACK_FILE",
   UNFINISHED_DATAPACK_UPLOAD_FORM = "UNFINISHED_DATAPACK_UPLOAD_FORM",
   DATAPACK_ALREADY_EXISTS = "DATAPACK_ALREADY_EXISTS",
-  NO_DATAPACKS_SELECTED = "NO_DATAPACKS_SELECTED"
+  NO_DATAPACKS_SELECTED = "NO_DATAPACKS_SELECTED",
+  INVALID_WIDTH_TYPE = "INVALID_WIDTH_TYPE"
```
paologum, line 23: you don't use this anymore so you should remove

app/src/util/error-codes.ts
```diff
@@ -45,5 +46,6 @@ export const ErrorMessages = {
   [ErrorCodes.UNFINISHED_DATAPACK_UPLOAD_FORM]:
     "Please finish the datapack upload form before attempting to upload the file.",
   [ErrorCodes.DATAPACK_ALREADY_EXISTS]: "Datapack already exists. Please upload a new datapack file.",
-  [ErrorCodes.NO_DATAPACKS_SELECTED]: "No datapacks selected. Please select at least one datapack to generate."
+  [ErrorCodes.NO_DATAPACKS_SELECTED]: "No datapacks selected. Please select at least one datapack to generate.",
+  [ErrorCodes.INVALID_WIDTH_TYPE]: "Invalid width type. Width should be a number."
```
paologum, line 50: same

app/src/SettingsTabs/EditWidthField.tsx
paologum: also if there is no width, it should return null since it can be undefined, you can add that as a check here or when you render it in the `ColumnMenu`. I would pick one or the other for everything though, so either use all return nulls on conditions or use the truthy values in `ColumnMenu` for the column and width objects

paologum (changes requested):

app/src/state/parse-settings.ts
lbostre: The stateColumn sometimes has the width undefined, while the presetColumn always has a width. If I were to just use stateColumn, I'm not sure how to the undefined case because I don't think there's a universal default width.

paologum: i think if you don't even set/add that line if it's undefined it should still work. I would try that and generate all charts to make sure it works though but I'm like 99% sure the width for metacolumns (non-existing width columns) should not matter

app/src/state/parse-settings.ts
```diff
@@ -409,6 +408,20 @@ function generateColumnXml(presetColumn: ColumnInfoTSC, stateColumn: ColumnInfo
...
```
paologum, line 422: i'm confused wouldn't this just be 
```
if (stateColumn && stateColumn.width) {
xml += `${indent}<setting name="width">${stateColumn.width}</setting>\n`;
}
```
i'm not sure what editWidth is off the top of my head so lmk

shared/src/index.ts
```diff
@@ -176,7 +178,8 @@ export type ColumnInfo = {
   maxAge: number;
   enableTitle: boolean;
   rgb: RGB;
-  width: number;
+  width?: number;
```
paologum, line 181: is there a reason width and units are on `FontLabelOptions`?

app/src/state/actions/column-actions.ts
```diff
@@ -58,24 +58,24 @@ export const updateEditName = action((newName: string) => {
   return;
 });
 
+export const updateWidth = action((columnObject: ColumnInfo, newWidth: number) => {
+  columnObject.width = newWidth;
+  return;
```
paologum, line 63: no need for a return

paologum (changes requested):
is there a reason my commits are like on top of yours 😭  it's lowk hard to read the pr with mine and your changes mixed together so maybe after this, try rebasing on `origin/main` again and see if the commits normalize bc it's kinda confusing. i'm not sure why it would layer our changes together tho

app/src/state/parse-settings.ts
lbostre: I changed it to that, but I'm still running into chart generation issues...I'm pretty sure that the issues are unrelated to this tho because they weren't working before changing this too

shared/src/index.ts
paologum: oh what the heck ok i was tweaking, this is fine haha

app/src/state/parse-settings.ts
```diff
@@ -383,6 +383,13 @@ function generateColumnXml(presetColumn: ColumnInfoTSC, stateColumn: ColumnInfo,
         } else {
           xml += `${indent}<setting name="${xmlKey}"/>\n`;
         }
+      } else if (key === "customColor") {
+      } else if (key === "width") {
+        if (stateColumn && stateColumn.width) {
+          xml += `${indent}<setting name="width">${stateColumn.width}</setting>\n`;
+        }
+      } else if (key === "orientation") {
+        xml += `${indent}<setting name="${xmlKey}" orientation="${presetColumn[key as keyof ColumnInfoTSC]}"/>\n`;
```
paologum, line 392: more of a question, but is there a reason you added `customColor` with nothing and `orientation` as part of the pr? it probably works but `orientation` is already checked for on line 412 below and `customColor` i'm kinda confused about

lbostre: oh nah i can remove that I think thats probably from me messing up a rebase

app/src/state/parse-settings.ts
```diff
@@ -383,6 +383,10 @@ function generateColumnXml(presetColumn: ColumnInfoTSC, stateColumn: ColumnInfo,
         } else {
           xml += `${indent}<setting name="${xmlKey}"/>\n`;
         }
+      } else if (key === "width") {
+        if (stateColumn && stateColumn.width) {
+          xml += `${indent}<setting name="width">${isNaN(stateColumn.width) ? 20 : stateColumn.width}</setting>\n`;
+        }
```
paologum, line 389: one small thing, we don't want the `stateColumn` to still be `isNaN` if we are actually successfully generating a chart. we want it to be actually valid so if `isNaN` then set the `stateColumn.width`. i'm not sure the specifics on whether you can actually change it here, but if you can't change the `stateColumn.width` here the other solution is

in `changeManuallyAddedColumns` in `generate-chart-actions.ts`, add a check on the width that sets it if it is `NaN` or 0

also yeah you don't have a check for 0 here so add that too

paologum (changes requested):

paologum (approved):
looks good ! :D

PR #71, opened 2024-04-04
sunsetting uploaded files

paologum:
server will delete user datapacks that have exceeded 2 weeks of being on the server.
Future Thoughts:
1. Might need to put a limit on user datapacks because someone might be able to just upload however many they want which could overload the server
2. Will need to brainstorm ways to tell the user their datapack has been sunset since we can't push requests/information to the app side from the server without a request.


server/src/file-metadata-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+ * @param destination
+ * @returns
+ */
+export function loadFileMetadata(destination: string) {
+  if (!existsSync(destination)) {
+    writeFileSync(destination, "{}");
+  }
+  const metadata = readFileSync(destination, "utf-8");
+  return JSON.parse(metadata);
+}
```
asivath, line 19: why synchronous?

server/src/file-metadata-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+  assertFileMetadataIndex(metadata);
+  metadata[filepath] = {
+    fileName,
+    uploadedAt: new Date().toISOString(),
+    decryptedFilepath,
+    mapPackIndexFilepath,
+    datapackIndexFilepath
+  };
+  writeFileSync(fileMetadataFilepath, JSON.stringify(metadata));
+}
```
asivath, line 45: same

server/src/file-metadata-handler.ts
```diff
@@ -0,0 +1,81 @@
...
+        const datapackIndex = JSON.parse(await readFile(metadata[file]!.datapackIndexFilepath, "utf-8"));
+        assertDatapackIndex(datapackIndex);
+        const mapPackIndex = JSON.parse(await readFile(metadata[file]!.mapPackIndexFilepath, "utf-8"));
+        assertMapPackIndex(mapPackIndex);
+        await rm(metadata[file]!.decryptedFilepath, { recursive: true, force: true });
+        unlink(file, (err) => {
+          if (err) {
+            console.error(`Error deleting file ${file} with error: ${err}`);
+          }
+        });
```
asivath, line 69: why use both rm and unlink instead of just one?

asivath (commented):
1. Should definitely have a limit not sure how much
2. If we have emails associated with users we could send an email about deletion. We could also give a notification when they log in but that would depend on how long their session lasts.

server/src/file-metadata-handler.ts
paologum: the synchronous file reading here in my opinion has very negligible stall time but i can change it

server/src/file-metadata-handler.ts
paologum: file is one file, rm is a directory but i guess i can use rm for both

server/src/file-metadata-handler.ts
asivath: yea in isolation but the more and more synchronous operations that happen means more blocking of the event loop

asivath (approved):

PR #73, opened 2024-04-04
changed structure of column info and bug fixes

jay-jhk-lee:
separated chart root and chart title in columninfo, changed affected functions, fixed chart title name not being modified


app/src/SettingsTabs/Column.tsx
```diff
@@ -216,7 +216,8 @@ export const Column = observer(function Column() {
           </TSCButton>
           {state.settingsTabs.columns && (
             <ColumnAccordion
-              details={state.settingsTabs.columns}
+              //takes in one chart title, change when multiple chart titles are involved
+              details={state.settingsTabs.columns.children[0]}
```
paologum, line 220: you might as well make this a for loop or get all the children of `columns` instead of just the 0th index in prep for multiple

paologum (changes requested):

PR #81, opened 2024-04-12
Tsc 122 extra information display

lbostre:
Displays extra information in column settings menu.

909525a79e209b44e8020242d4eacc6b


app/src/SettingsTabs/ColumnMenu.tsx
```diff
@@ -83,6 +87,7 @@ export const ColumnMenu = observer(() => {
       <div id="ColumnMenuContent" className="column-menu-content">
         {state.settingsTabs.columnSelected && <EditNameField />}
         {state.settingsTabs.columnSelected && <FontMenu />}
+        {state.settingsTabs.columnSelected && info !== "" && <InfoBox />}
```
paologum, line 90: this can just be `info !== ""`

app/src/SettingsTabs/InfoBox.css
```diff
@@ -0,0 +1,18 @@
...
+  border-radius: 5px;
+  margin-top: 5px;
+  padding: 4px;
+  max-height: 200px;
+}
+
+.scroll-bar {
+  max-height: 200px;
+  /*padding-right: 10px;*/
+}
```
paologum, line 18: remove comments if unneeded

app/src/SettingsTabs/InfoBox.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const info =
+    state.settingsTabs.columnSelected === null
+      ? ""
+      : state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.popup;
+  return (
+    <div className="container">
+      <Typography style={{ fontWeight: "bold" }}>Information and References</Typography>
+      <Box className="info-box">
+        <StyledScrollbar className="scroll-bar">
+          <div dangerouslySetInnerHTML={{ __html: info.substring(1, info.length - 1).replace('""', '"') }} />
```
paologum, line 19: does this assume that info is always in quotes? i don't think that's the case. you probably should make a `trimQuotes` function similar to the one in `server/src/util.ts` that checks if it's in quotes then trims it. Also shouldn't it be `.replaceAll`

app/src/SettingsTabs/InfoBox.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Box, Typography } from "@mui/material";
+import "./InfoBox.css";
+import { StyledScrollbar } from "../components";
+
+export const InfoBox = observer(() => {
+  const { state } = useContext(context);
+  const info =
+    state.settingsTabs.columnSelected === null
+      ? ""
+      : state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.popup;
```
paologum, line 13: this is kinda redundant, just pass it in through the params

app/src/SettingsTabs/ColumnMenu.tsx
```diff
@@ -46,7 +47,10 @@ const EditNameField = observer(() => {
 export const ColumnMenu = observer(() => {
   const { state } = useContext(context);
   const [openMenu, setOpenMenu] = useState(false);
-
+  const info =
+    state.settingsTabs.columnSelected === null
+      ? ""
+      : state.settingsTabs.columnHashMap.get(state.settingsTabs.columnSelected)!.popup;
```
paologum, line 53: you should also make sure `state.settingsTabs.columnHashMap.get(...)` isn't null because if so the ! will make bad things happen

paologum (changes requested):
looks great so far :D

app/src/SettingsTabs/InfoBox.tsx
```diff
@@ -0,0 +1,25 @@
...
+import "./InfoBox.css";
+import { StyledScrollbar } from "../components";
+
+export const InfoBox: React.FC<{ info: string }> = observer(({ info }) => {
+  function trimQuotes(input: string): string {
+    if (input.startsWith('"') && input.endsWith('"')) {
+      return input.slice(1, -1);
+    }
+    return input;
+  }
```
paologum, line 13: you should move this to `util.ts` on the app side because other components of the app could use this function.

app/src/SettingsTabs/ColumnMenu.tsx
```diff
@@ -85,6 +89,7 @@ export const ColumnMenu = observer(() => {
       <div id="ColumnMenuContent" className="column-menu-content">
         {column && <EditNameField />}
         {column && <FontMenu column={column} />}
+        {info !== "" && <InfoBox info={info ?? ""} />}
```
paologum, line 92: if you do `{!!info && <InfoBox info={info} />}` `{info !== "" && <InfoBox info={info ?? ""} />}`
i think that would be a lot cleaner.

empty strings are falsy so we use that to shorten syntax

paologum (changes requested):

paologum (approved):
looks good!

PR #82, opened 2024-04-12
allow for datapacks from user directories to be loaded

paologum:
doesn't actually generate the chart, but the datapacks are created correctly
the settings parser needs to create the settings xml solely from the `ColumnInfo` object for it to work
also made some minor changes like abstracting the `fetchSettingsXml` on the app side and moving url params to the body when fetching a chart

### DIFF
--- a/app/src/state/actions/general-actions.ts
+++ b/app/src/state/actions/general-actions.ts
@@ -229,41 +229,21 @@ const setChartSettings = action("setChartSettings", (settings: ChartSettingsInfo
  */
 export const setDatapackConfig = action(
   "setChart",
-  async (datapacks: string[], settingsPath: string): Promise<boolean> => {
+  async (datapacks: string[], settingsPath?: string): Promise<boolean> => {
     let datapackAgeInfo: DatapackAgeInfo = {
       datapackContainsSuggAge: false
     };
     let mapInfo: MapInfo = {};
     let mapHierarchy: MapHierarchy = {};
     let columnInfo: ColumnInfo;
+    let chartSettings: ChartInfoTSC = {};
     try {
-      // Grab the settings for this chart if there are any:
-      //TODO: only get default settings file if its a preset
       if (settingsPath && settingsPath.length > 0) {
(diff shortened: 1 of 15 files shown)


app/src/state/actions/general-actions.ts
```diff
@@ -352,8 +345,8 @@ export const setDatapackConfig = action(
     state.settingsTabs.columns = columnInfo;
     state.mapState.mapInfo = mapInfo;
     state.config.datapacks = datapacks;
-    state.config.settingsPath = settingsPath;
-    initializeColumnHashMap(state.settingsTabs.columns);
+    state.settingsTSC = chartSettings;
+    initializeColumnHashMap(columnInfo);
```
jay-jhk-lee, line 349: this should be state.settingsTabs.columns, not columnInfo

jay-jhk-lee: inside initializeColumnHashMap

paologum: they are essentially the same thing, but because `columnInfo` is the actual object being passed in i prefer it over the state object that could be something else if order is switched around accidentally in the future

jay-jhk-lee: They aren't the same thing. If you pass in columnInfo, you can't change the settings for the columns chart title and Ma. If you pass in state.column, then changing the settings for them work. I'm assuming it's something to do with us predefining chart title and Ma, so maybe it works if they weren't defined initially but added afterwards.

jay-jhk-lee: for example, for AfricaBights editing the name for chart title or Ma doesn't work but it works for central africa cenozoic

paologum: wait that doesn't make any sense because we call `state.settingsTabs.columns = columnInfo;` two lines above

jay-jhk-lee: I'm not sure of the root cause, but when I tested it that was how it was

asivath (approved):

PR #83, opened 2024-04-13
font options should be recursively set in the parsing of datapacks, chron multi column capability, subTransect fix

paologum:
depending on the column type, the font options that should show for the font menu are now correctly set
the java program tends to give all font options when given a column type that isn't recognized, but i did some trial and error to find out what actually matters in those columns and whittled it down. this could be wrong so i added todos for future reference. Depending if it is a meta column or an actual leaf column, different font menus will be shown.

also chron columns consist of multiple columns so i added functionality for that. 

also added `subTransect` onto `ColumnInfo` since it wasn't there before (bug)
Screenshot 2024-04-13 at 5 50 22 PM
Screenshot 2024-04-13 at 5 50 43 PM
Screenshot 2024-04-13 at 5 51 07 PM

### DIFF
--- a/app/src/SettingsTabs/ColumnMenu.tsx
+++ b/app/src/SettingsTabs/ColumnMenu.tsx
@@ -82,7 +82,7 @@ export const ColumnMenu = observer(() => {
       </div>
       <div id="ColumnMenuContent" className="column-menu-content">
         {state.settingsTabs.columnSelected && <EditNameField />}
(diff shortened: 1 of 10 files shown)


lbostre (approved):
Looks good!

jay-jhk-lee (approved):
looks good 👍

asivath (approved):

PR #84, opened 2024-04-15
changing structure of datapacks to accommodate different charts with different units (other fixes as well)

paologum:
now each datapack will have a unit identifier when a datapack gets parsed which allows it to merge cleanly.
datapacks will now only have one `ColumnInfo` object that represents the "Chart" with all the actual columns as its children.
now if there are different types of units, you can toggle between them in the time tab

other units besides Ma will have their chart title set to their units

other fixes:
vertical scale fixed to multiply by a factor of 30
changed fontmenu structuring to clean things up and added an `on` variable to `fontsInfo` to toggle things

things i found that have bugs (new tasks):
* `datapackAgeInfo` has to account for if the user has changed settings or not, if they have changed then don't ask for a popup
* also since there can be multiple datapacks, there must be `datapackAgeInfo` for each unit type
* font inheritance is done in place so we have to manually add inheriting fonts down from parent to child and if the setting gets turned off on the parent, to revert back. if the user has changed a font in any of the children of the inheriting parent, don't change that child

https://github.com/earthhistoryviz/tsconline/assets/89664317/bedbd510-2692-4dcf-b72f-4a34b44ba31d

Screenshot 2024-04-15 at 3 28 34 AM


jay-jhk-lee (commented):
Looks Good

jay-jhk-lee (approved):

paologum:
Screenshot 2024-04-15 at 7 30 51 PM

asivath (approved):

PR #85, opened 2024-04-16
Tsc 267 create login frontend

asivath:
https://github.com/earthhistoryviz/tsconline/assets/94872897/a6a3017b-9a1b-4c22-82c4-21e37a77a96f


app/src/Login.tsx
```diff
@@ -0,0 +1,115 @@
...
+        headers: {
+          "Content-Type": "application/json"
+        },
+        body: JSON.stringify(formData)
+      });
+      if (login.ok) {
+        actions.sessionCheck();
+        if (!state.isLoggedIn) {
+          actions.pushError(ErrorCodes.UNABLE_TO_LOGIN);
+        }
```
paologum, line 43: you should remove error if logged in correctly

app/src/main.tsx
```diff
@@ -3,13 +3,16 @@ import ReactDOM from "react-dom/client";
...
 
 ReactDOM.createRoot(document.getElementById("root") as HTMLElement).render(
-  <React.StrictMode>
-    <context.Provider value={initialContext}>
-      <BrowserRouter>
-        <App />
-      </BrowserRouter>
-    </context.Provider>
-  </React.StrictMode>
+  <GoogleOAuthProvider clientId={import.meta.env.VITE_GOOGLE_CLIENT_ID}>
```
paologum, line 9: does local deving need this env variable?

app/src/Login.tsx
```diff
@@ -0,0 +1,115 @@
...
+      password: data.get("password")
+    };
+    try {
+      const login = await fetcher("auth/login", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json"
+        },
+        body: JSON.stringify(formData)
+      });
```
paologum, line 38: is this a route we have?

app/src/state/actions/general-actions.ts
```diff
@@ -619,6 +619,40 @@ export const fetchImage = action("fetchImage", async (datapackName: string, imag
   return image;
 });
 
+export const logout = action("logout", async () => {
+  try {
+    const response = await fetcher("/auth/logout", {
+      method: "POST",
+      credentials: "include"
+    });
+    if (response.ok) {
```
paologum, line 628: same question here, i'm just wondering how this works mostly

app/src/state/actions/general-actions.ts
```diff
@@ -619,6 +619,40 @@ export const fetchImage = action("fetchImage", async (datapackName: string, imag
...
+      setIsLoggedIn(false);
+    }
+  } catch (error) {
+    console.error("Failed to logout:", error);
+  }
```
paologum, line 633: should be a display error code right?

app/src/NavBar.tsx
```diff
@@ -8,13 +8,15 @@ import HomeIcon from "@mui/icons-material/Home";
...
 import { TSCButton, TSCTabs } from "./components";
+import AccountCircleIcon from "@mui/icons-material/AccountCircle";
 
 import "./NavBar.css";
 
 export const NavBar = observer(function Navbar() {
   const theme = useTheme();
   const { state, actions } = useContext(context);
   const navigate = useNavigate();
+  actions.sessionCheck();
```
paologum, line 19: why do you need to run a session check here? can you explain what this session check will do so i can understand this better

paologum (changes requested):
can you sign in so i can see the ui for after signing in successfully. just like hard code it for the demo so i can see the navbar sign out stuff and also the signout flow.

app/src/main.tsx
asivath: no it's not needed unless you want to use the google login

app/src/Login.tsx
asivath: they exist in the backend branch I just put them here since they're for the frontend

app/src/NavBar.tsx
asivath: i forgot I put that there it was for debugging. its meant to check if a user has a valid session, so probably the two places it should be is after we get a ok response from the login endpoint and some element that will always get rendered so that we can check is a user is logged in after reopening the page, maybe App.tsx?

app/src/state/actions/general-actions.ts
asivath: same in another branch just invalidates a session

asivath:
https://github.com/earthhistoryviz/tsconline/assets/94872897/e8564fbc-0a30-4d54-8c3c-b4428e29dd6c

app/src/main.tsx
asivath: actually since this does not need to be private I could just hardcode it so anyone should be able to use it

app/src/NavBar.tsx
paologum: i think it should get called on `initialize()` and on successful login based on what you said. although do you need to check the session on startup of the app? right now there is no state that saves information between refreshes so you would just be logged out if you refresh at the moment. unless the way you implemented it will keep information through refreshes (i haven't seen that code so i'm assuming)

also we shouldn't put the session check in a raw element imo

app/src/App.tsx
```diff
@@ -11,14 +11,20 @@ import theme from "./theme";
...
+import { Login } from "./Login";
+import { SignUp } from "./SignUp";
+import { AccountRecovery } from "./AccountRecovery";
 import { TSCPopupDialog, TSCError, TSCSnackbar } from "./components";
+import { CssBaseline } from "@mui/material";
 
 export default observer(function App() {
   const { state, actions } = useContext(context);
   const navigate = useNavigate();
+  actions.sessionCheck();
```
paologum, line 23: my previous response, i think this will session check when we don't really need to on every state change

app/src/Login.tsx
```diff
@@ -0,0 +1,122 @@
...
+        }
+      } else {
+        console.error("Error trying to log in: " + login.status);
+        actions.pushError(ErrorCodes.UNABLE_TO_LOGIN);
+      }
+    } catch (error) {
+      console.error("Error:", error);
+      actions.pushError(ErrorCodes.UNABLE_TO_LOGIN);
+    }
+  };
```
paologum, line 59: i forgot to mention, but for debugging purposes, you should use `displayServerError` and pass it the payload from the server and the error generated in the catch. Additionally on any error on the server in your backend code, it should return a json with `{ error: <message> }`

app/src/Login.tsx
```diff
@@ -0,0 +1,122 @@
...
+            <Typography variant="caption" sx={{ color: "black" }}>
+              or
+            </Typography>
+          </Box>
+          <Box className="divider-line"></Box>
+        </Box>
+        <GoogleLogin
+          onSuccess={() => console.log("Logged in with Google")}
+          ux_mode="redirect"
+          login_uri={`${import.meta.env.VITE_SERVER_URL || "https://localhost:3000"}/auth/oauth`}
```
paologum, line 115: is this not `devSafeUrl`?

app/src/NavBar.tsx
```diff
@@ -62,6 +63,39 @@ export const NavBar = observer(function Navbar() {
...
         <TSCButton onClick={() => actions.initiateChartGeneration(navigate)}>Generate Chart</TSCButton>
+        {state.isLoggedIn ? (
+          <Tab
+            className="login-tab"
+            value={5}
+            label="Sign out"
+            icon={<AccountCircleIcon />}
+            onClick={() => {
+              actions.logout();
+              navigate("/");
```
paologum, line 74: i don't know if logging out should navigate away from where they are, bc the difference is they shouldn't be on the login screen anymore

app/src/SignUp.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  return (
+    <Container component="main" maxWidth="xs">
+      <Box
+        sx={{
+          marginTop: 8,
+          display: "flex",
+          flexDirection: "column",
+          alignItems: "center"
+        }}>
```
paologum, line 36: css

app/src/main.tsx
paologum: did you wanna hardcode this still?

paologum (changes requested):

paologum:
oh yeah also, i'm assuming that you don't have a demo of the google auth because that doesn't work until backend? but if it does work can you add a demo of that

app/src/NavBar.tsx
asivath: yea I'll put it in initialize(). the session gets stored in a cookie so it will persist between refreshes

asivath:
yeah the google login will try to redirect and pass stuff to the backend and will just fail

app/src/Login.tsx
```diff
@@ -42,19 +43,17 @@ export const Login: React.FC = observer(() => {
...
       } else {
-        console.error("Error trying to log in: " + login.status);
-        actions.pushError(ErrorCodes.UNABLE_TO_LOGIN);
+        displayServerError(login, ErrorCodes.UNABLE_TO_LOGIN, "Error trying to log in");
       }
     } catch (error) {
-      console.error("Error:", error);
-      actions.pushError(ErrorCodes.UNABLE_TO_LOGIN);
+      displayServerError(error, ErrorCodes.UNABLE_TO_LOGIN, "Error trying to log in");
     }
```
paologum, line 57: the error messages should just be the ErrorMessages[ErrorCodes.UNABLE_TO_LOGIN]

app/src/state/initialize.ts
```diff
@@ -3,6 +3,7 @@ import { actions } from "./index";
 export async function initialize() {
   // If we're running in dev mode (yarn dev), then the app is not served from the same URL
   // as the server hosts the /charts endpoint.  So, we'll hard-code that for ourselves here.
+  actions.sessionCheck();
```
paologum, line 6: idk what will require us to ask for cookie permissions but i doubt this is one of those situations right

paologum (changes requested):

app/src/state/initialize.ts
asivath: should that be a part of this pr

paologum: mmm no i think this does a lot already so you can just make that it's own thing. since we aren't technically publicly released or whatever, it should be alright.

i would say that is one of your next priorities though

paologum (approved):

PR #86, opened 2024-04-17
Translate columnInfo to type TSC, parse type TSC to xml, changes for multiple datapacks now work

jay-jhk-lee:


app/src/state/parse-settings.ts
```diff
@@ -367,121 +336,270 @@ function generateFontsXml(fonts: any, colName: string, stateColumn: any, indent:
...
+  for (let i = settings.topAge.length - 1; i >= 0; i--) {
+    xml += `${indent}<setting name="topAge" source="text" unit="${settings.topAge[i].unit}">\n`;
+    xml += `${indent}    <setting name="text">${settings.topAge[i].text}</setting>\n`;
+    xml += `${indent}</setting>\n`;
+    xml += `${indent}<setting name="baseAge" source="text" unit="${settings.baseAge[i].unit}">\n`;
+    xml += `${indent}    <setting name="text">${settings.baseAge[i].text}</setting>\n`;
+    xml += `${indent}</setting>\n`;
+    xml += `${indent}<setting name="unitsPerMY" unit="${settings.unitsPerMY[i].unit}">${settings.unitsPerMY[i].text}</setting>\n`;
+    xml += `${indent}<setting name="skipEmptyColumns" unit="${settings.skipEmptyColumns[i].unit}">${settings.skipEmptyColumns[i].text}</setting>\n`;
+  }
```
paologum, line 496: why did this change off of 
```
function generateSettingsXml(stateSettings: ChartSettings, indent: string): string {
  let xml = "";
  for (const unit in stateSettings.timeSettings) {
    const timeSettings = stateSettings.timeSettings[unit];
    xml += `${indent}<setting name="topAge" source="text" unit="${unit}">\n`;
    xml += `${indent}    <setting name="text">${timeSettings.topStageAge}</setting>\n`;
    xml += `${indent}</setting>\n`;
    xml += `${indent}<setting name="baseAge" source="text" unit="${unit}">\n`;
    xml += `${indent}    <setting name="text">${timeSettings.baseStageAge}</setting>\n`;
    xml += `${indent}</setting>\n`;
    xml += `${indent}<setting name="unitsPerMY" unit="${unit}">${timeSettings.unitsPerMY * 30}</setting>\n`;
    xml += `${indent}<setting name="skipEmptyColumns">${timeSettings.skipEmptyColumns}</setting>\n`;
  }
  xml += `${indent}<setting name="variableColors">UNESCO</setting>\n`;
  xml += `${indent}<setting name="negativeChk">false</setting>\n`;
  xml += `${indent}<setting name="doPopups">${stateSettings.mouseOverPopupsEnabled}</setting>\n`;
  xml += `${indent}<setting name="enEventColBG">${stateSettings.enableColumnBackground}</setting>\n`;
  xml += `${indent}<setting name="enChartLegend">${stateSettings.enableChartLegend}</setting>\n`;
  xml += `${indent}<setting name="enPriority">${stateSettings.enablePriority}</setting>\n`;
  xml += `${indent}<setting name="enHideBlockLabel">${stateSettings.enableHideBlockLabel}</setting>\n`;
  return xml;
```

i don't see a reason to change what we previously had to convert back into `ChartSettingsTSC` since how it currently is is completely fine and doesn't require any change. it's also a lot more readable if we take the state object and build the xml file that way. since we already have to pull from `ChartSettingsTSC` into the state, changing back into `ChartSettingsTSC` seems redundant

it would also reduce the code in not having a conversion function `translateSettings`

we could in the future cut off the middle man and have them be the same thing but at the very least i don't see the need to have `translateSettings`

app/src/state/actions/generate-chart-actions.ts
```diff
@@ -41,13 +41,12 @@ export const fetchChartFromServer = action("fetchChartFromServer", async (naviga
...
   generalActions.setChartContent("");
-  //let xmlSettings = jsonToXml(state.settingsJSON); // Convert JSON to XML using jsonToXml function
-  // console.log("XML Settings:", xmlSettings); // Log the XML settings to the console
   let body;
   try {
     const columnCopy: ColumnInfo = JSON.parse(JSON.stringify(state.settingsTabs.columns));
     changeManuallyAddedColumns(columnCopy);
-    const xmlSettings = jsonToXml(state.settingsTSC, columnCopy, state.settings);
+    const xmlSettings = jsonToXml(columnCopy, state.settings);
+    console.log(xmlSettings);
```
paologum, line 49: remove :D

app/src/state/parse-settings.ts
```diff
@@ -367,121 +336,270 @@ function generateFontsXml(fonts: any, colName: string, stateColumn: any, indent:
...
+  //freehand column
+  else if (state.subFreehandInfo) {
+    column._id = "class datastore.FreehandColumn:" + replaceSpecialChars(state.name, 0);
+  }
+  //sequence column
+  else if (state.subSequenceInfo) {
+    column = JSON.parse(JSON.stringify(defaultSequenceColumnInfoTSC));
+    column._id = "class datastore.SequenceColumn:" + replaceSpecialChars(state.name, 0);
+  } else if (state.subTransectInfo) {
+    column._id = "class datastore.TransectColumn:" + replaceSpecialChars(state.name, 0);
```
paologum, line 400: i think with my pr this can be like three lines where you have `class datastore.${columnDisplayType}: + replace...`

app/src/state/parse-settings.ts
```diff
@@ -367,121 +336,270 @@ function generateFontsXml(fonts: any, colName: string, stateColumn: any, indent:
...
+      //get first word of units
+      if (state.name.indexOf(" ") === -1) {
+        column._id = "class datastore.RulerColumn:" + state.name;
+      } else {
+        column._id = "class datastore.RulerColumn:" + state.name.substring(0, state.name.indexOf(" "));
+      }
+    } else {
+      column._id = "class datastore.BlankColumn:" + state.name;
+    }
+  }
```
paologum, line 436: same here 😢  i probably shoulda told you i was working on it today. i have all these on the column now so you don't need any if's imo from my understanding

app/src/state/parse-settings.ts
```diff
@@ -367,121 +336,270 @@ function generateFontsXml(fonts: any, colName: string, stateColumn: any, indent:
...
+    if (state.name.includes("Chart Title in ")) {
+      let delimiter = " ",
+        start = 3,
+        tokens = state.name.split(delimiter).slice(start),
+        result = tokens.join(delimiter);
+      column._id = "class datastore.RootColumn:" + result;
+    }
+    //edge case for default datapack
+    else if (state.name === "TimeScale Creator GTS2020 chart") {
+      column._id = "class datastore.RootColumn:Chart Title";
```
paologum, line 448: i set chart titles to rootcolumns as well so we don't have to worry about weird edge cases with different titles

app/src/state/parse-settings.ts
```diff
@@ -367,121 +336,270 @@ function generateFontsXml(fonts: any, colName: string, stateColumn: any, indent:
...
+    if (key in fonts) {
+      const inheritable = fonts[key as keyof FontsInfo].inheritable;
+      if (JSON.stringify(newStateColumn[key]) === JSON.stringify(defInfo[key])) {
+        xml += `${indent}<font function="${key}" inheritable="${inheritable}"/>\n`;
+      } else {
+        xml += `${indent}<font function="${key}" inheritable="${inheritable}">`;
+        for (let fKey in newStateColumn[key]) {
+          if (JSON.stringify(newStateColumn[key][fKey]) !== JSON.stringify(defInfo[key][fKey])) {
+            if (fKey === "fontFace") {
+              xml += `font-family: ${newStateColumn[key][fKey]};`;
```
paologum, line 523: i changed some of the fonts as well so that they're more readable and direct instead of super generalized. like directly accessing the property instead of iterating over the keys. that's the direction i want the parser to head towards imo because i think the generalized concept of what we have made so far/the people before us is soooo hard to integrate for new developers.

app/src/state/parse-settings.ts
```diff
@@ -367,121 +336,270 @@ function generateFontsXml(fonts: any, colName: string, stateColumn: any, indent:
...
+        xml += `${indent}<column id="${replaceSpecialChars(column.children[i]._id, 0)}">\n`;
+        xml += columnInfoTSCToXml(column.children[i], `${indent}    `);
+        xml += `${indent}</column>\n`;
       }
+    } else if (key === "justification") {
+      xml += `${indent}<setting justification="${column[key as keyof ColumnInfoTSC]}" name="justification"/>\n`;
+    } else if (key === "orientation") {
+      xml += `${indent}<setting name="orientation" orientation="${column[key as keyof ColumnInfoTSC]}"/>\n`;
+    } else {
+      xml += `${indent}<setting name="${key}">${column[key as keyof ColumnInfoTSC]}</setting>\n`;
```
paologum, line 582: same here, i think if we have less if statements and more direct coding, life would be a lot easier in knowing where everything goes. so for example instead of 
```
else if (key === "justification") {
      xml += `${indent}<setting justification="${column[key as keyof ColumnInfoTSC]}" name="justification"/>\n`;
    }
```
we have
```
xml += `${indent}<setting justification="${column.justification}" name="justification"/>\n`;
```

imo it feels a lot cleaner and understandable. lmk what you think

app/src/state/parse-settings.ts
```diff
@@ -367,121 +336,270 @@ function generateFontsXml(fonts: any, colName: string, stateColumn: any, indent:
...
-  xml += generateColumnXml(settingsTSC["class datastore.RootColumn:Chart Root"]!, columnSettings, "        ");
+  xml += columnInfoTSCToXml(settingsTSC["class datastore.RootColumn:Chart Root"]!, "        ");
   xml += "    </column>\n";
   xml += "</TSCreator>\n";
   return xml;
 }
+
+export function jsonToXml(state: ColumnInfo, settings: ChartSettings): string {
+  return ChartInfoTSCToXml(columnInfoToSettingsTSC(state, settings));
+}
```
paologum, line 605: is this method necessary? isn't it just `ChartInfoTSCToXml`

server/hs_err_pid13487.log
paologum, line 1: remove all hs_err.logs

server/public/presets/001-TSC2020-TEST/settings_copy.tsc
paologum, line 1: this isn't needed right

shared/src/constants.ts
```diff
@@ -206,7 +206,8 @@ export const defaultColumnBasicInfoTSC: ColumnBasicInfoTSC = {
 
 export const defaultEventColumnInfoTSC: EventColumnInfoTSC = {
   ...defaultColumnBasicInfoTSC,
-  type: ""
+  type: "events",
+  rangeSort: "first occurrence"
 };
```
paologum, line 211: default events are 150 width and off

shared/src/constants.ts
```diff
@@ -187,13 +187,13 @@ export const defaultColumnBasicInfoTSC: ColumnBasicInfoTSC = {
   title: "",
   useNamedColor: false,
   placeHolder: false,
-  drawTitle: false,
+  drawTitle: true,
   drawAgeLabel: false,
   drawUncertaintyLabel: false,
   isSelected: false,
```
paologum, line 193: i think columns are default on right?

shared/src/settings-types.ts
```diff
@@ -1,7 +1,7 @@
 import { FontsInfo, assertFontsInfo, throwError } from "./index.js";
 
 export type ChartInfoTSC = {
-  settings?: ChartSettingsInfoTSC;
+  settings: ChartSettingsInfoTSC;
   "class datastore.RootColumn:Chart Root"?: ColumnInfoTSC;
```
paologum, line 5: root should always exist, at least for us because we manually will always append root right

paologum (changes requested):
some small tings n dat 🕺

app/src/state/parse-settings.ts
jay-jhk-lee: My initial idea was that there would be two parts to parsing: columnInfo + ChartSettings -> ChartInfoTSC then ChartInfoTSC->xml. I wanted to have a function that strictly takes in a ChartInfoTSC object and creates a xml file. Since ChartInfoTSC has ChartSettingsInfoTSC and not ChartSettings, I needed to translate it. I think I did that because the settings didn't have all the options when I began. Since it does now, I'll change the ChartInfoTSC type to have ChartSettings instead of ChartInfoTSC and use your function. I do like having a middle man from columnInfo to ColumnInfoTSC since there's a lot of stuff that isn't needed to generate the xml in columninfo.

server/public/presets/001-TSC2020-TEST/settings_copy.tsc
jay-jhk-lee: yeah

app/src/state/parse-settings.ts
jay-jhk-lee: sounds good 👌

app/src/state/parse-settings.ts
jay-jhk-lee: I changed it in my new pr :D

app/src/state/parse-settings.ts
jay-jhk-lee: the current problem with doing `column.justification` is since the ColumnInfoTSC type isn't guaranteed to have the justification property (only RulerColumn has it), typescript doesn't let me do that. So I did a workaround with `as keyof`. I do agree with you tho

app/src/state/parse-settings.ts
paologum: i think though over time in the long run, `ChartInfoTSC` will have to get phased out/merged into what `ColumnInfo` is  because there is technically no need for two type of the same data structure.

app/src/state/parse-settings.ts
paologum: ok i guess that's fine but can you cast the key when/near its definition so that you don't cast it three times

app/src/state/parse-settings.ts
```diff
@@ -308,6 +304,64 @@ function generateSettingsXml(stateSettings: ChartSettings, indent: string): stri
   xml += `${indent}<setting name="enHideBlockLabel">${stateSettings.enableHideBlockLabel}</setting>\n`;
   return xml;
 }
+
+export function translateColumn(state: ColumnInfo): ColumnInfoTSC {
```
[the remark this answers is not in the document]
paologum: also as i said in a previous reply, this is something we're gonna phase out at some point this summer because this is redundant

app/src/state/parse-settings.ts
```diff
@@ -342,123 +396,59 @@ function generateFontsXml(indent: string, fontsInfo?: FontsInfo): string {
...
+        xml += `${indent}<setting name="${key}" standardized="${column[key].standardized}" 
+        useNamed="${column[key].useNamed}">${column[key].text}</setting>\n`;
+      } else if ("useNamed" in column[key]) {
+        xml += `${indent}<setting name="${key}" useNamed="${column[key].useNamed}">${column[key].text}</setting>\n`;
+      } else if ("standardized" in column[key]) {
+        xml += `${indent}<setting name="${key}" useNamed="${column[key].standardized}">${column[key].text}</setting>\n`;
+      } else if (column[key].text.length > 0) {
+        xml += `${indent}<setting name="${key}">${column[key].text}</setting>\n`;
+      } else if (column[key].text.length > 0) {
+        xml += `${indent}<setting name="${key}">${column[key].text}</setting>\n`;
```
paologum, line 418: same thing twice

app/src/state/parse-settings.ts
```diff
@@ -342,123 +396,59 @@ function generateFontsXml(indent: string, fontsInfo?: FontsInfo): string {
...
+    if (key === "title") {
+      xml += `${indent}<setting name="title">${column[key]}</setting>\n`;
+    } else if (key === "backgroundColor" || key === "customColor") {
+      if ("standardized" in column[key] && "useNamed" in column[key]) {
+        xml += `${indent}<setting name="${key}" standardized="${column[key].standardized}" 
+        useNamed="${column[key].useNamed}">${column[key].text}</setting>\n`;
+      } else if ("useNamed" in column[key]) {
+        xml += `${indent}<setting name="${key}" useNamed="${column[key].useNamed}">${column[key].text}</setting>\n`;
+      } else if ("standardized" in column[key]) {
+        xml += `${indent}<setting name="${key}" useNamed="${column[key].standardized}">${column[key].text}</setting>\n`;
```
paologum, line 414: tbh these cases will never occur. the only thing that will happen is it's not default, or we have an rgb value

`useNamed` and `standardized` will never happen (at the moment) so ultimately you can just have

xml += `${indent}<setting name="${key}">${column[key].text}</setting>\n`;

or you can hard code `useNamed` to false since that's how it is for every column. i don't see a need to included any of this code if we're not going to be using it (i don't even think we should have this tbh)

paologum (changes requested):

app/src/state/parse-settings.ts
```diff
@@ -308,6 +304,64 @@ function generateSettingsXml(stateSettings: ChartSettings, indent: string): stri
...
+      break;
+    case "Point":
+      column = JSON.parse(JSON.stringify(defaultPointColumnInfoTSC));
+  }
+  switch (state.columnDisplayType) {
+    case "RootColumn":
+      if (state.name.substring(0, 14) === "Chart Title in") {
+        column._id = `class datastore.${state.columnDisplayType}:` + state.name.substring(15, state.name.length);
+      } else column._id = `class datastore.${state.columnDisplayType}:` + state.name;
+      break;
```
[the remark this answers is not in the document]
jay-jhk-lee: I guess we could set it to editName for chart root since it doesn't change, but the edit name for the chart title can change so it should be set to `state.name`. Basically, editName is equivalent to title and name is equivalent to id. To keep it consistent, we should set it to name regardless.

app/src/state/parse-settings.ts
jay-jhk-lee: I do see that `useNamed` and `standardized` will never currently, but the the goal of this function was to translate any ColumnInfoTSC into xml. The ColumnInfoTSC type can have those values, so I think we should cover all the cases. The function also covers the case you've mentioned too.`useNamed` is not always false, I've looked at settings files that have both true and false. I think we should have this since it does exist in settings.tsc.

app/src/state/parse-settings.ts
```diff
@@ -308,6 +304,64 @@ function generateSettingsXml(stateSettings: ChartSettings, indent: string): stri
...
+    case "MetaColumn":
+    case "BlockSeriesMetaColumn":
+      column._id = `class datastore.${state.columnDisplayType}:` + state.name;
+      break;
+    default:
+      column._id = `class datastore.${state.columnDisplayType}Column:` + state.name;
+  }
+  column.title = replaceSpecialChars(state.editName, 1);
```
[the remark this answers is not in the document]
jay-jhk-lee: I agree, it's a bit ambiguous. I think instead of`convertToSpecialChars`, `escapeSpecialChars` is a bit better since that's what the process is called. lmk what you think

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
jay-jhk-lee: If you are referring to it being redundant in the case of parsing from ColumnInfo to xml, I agree. But I think it's valuable to have something that connects ColumnInfo to ColumnInfoTSC (unless you want to phase out the TSC types, but I think we should keep them). That could be in the form of functions for parsing ColumnInfo to xml, but I think it's easier to understand if we have a function that translates it directly.

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
paologum: actually i just checked, what matters is that the name is different only. The id has nothing to do with display. so you can just do `column._id = `class datastore.${state.columnDisplayType}:` + state.name;`

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
paologum: i was doing some searching and the "special chars" we are replacing are specific to html so i agree escape would be good so lets call it `escapeHtmlChars`

and if you could, could you make the parameters `(text: string, type: "min" | "all" = "min")` instead of type 0 vs 1 for clarity as well?

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
paologum: for now it's fine we can worry/discuss it when the time comes. (technically though all of tsc would be available in a regular columninfo variable, with maybe some shortcuts for things we don't support)

app/src/state/parse-settings.ts
paologum: the problem with that is that it will introduce weird cases we might not now the source of. to keep the main from having potential bugs that we are not aware of, we should keep the most amount of code limited to what we know is the case. I know it's not always the case of `useNamed = false` but nothing on our app will use any sort of named or standardized colors meaning that if they change it, they might produce weird use cases since that field will be true or standardized. we will definitely revisit it when we want to work on supporting it but for being proactive and limiting our range of features as we have been doing, i think this is the best approach

jay-jhk-lee: I see your point. Since we don't have anything on the app that changes those fields, it will never be filled. If/when we do add functionality to change those fields, it might introduce unexpected outputs. You want to be sure that it only uses the background color since that's what we have implemented. I'll comment the code out instead of removing it since that is the correct output if those fields are defined so we can use it later if applicable.

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
jay-jhk-lee: For the columnDIsplayType other than root, meta, and blockseries, they don't have the "Column" suffix so I have to add it. For Chart title, some chart titles have the "Chart Title in" prefix which is manually added so I have to remove it.

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
jay-jhk-lee: What do you mean by `type: "min" | "all" = "min"`? I'll change it to `type: "att" | "text"` for now

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
paologum: the code above is just `RootColumn` where I'm saying the id "doesn't matter". I'm saying you don't need to remove the prefix because all that matters is this root column does not have the same id as another root column

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
paologum: sure that works, what i was saying is that you have two different ways of escaping, one with all the escaping and one with minimum amount of escaping

app/src/state/parse-settings.ts
paologum: sure that works :D

app/src/state/actions/generate-chart-actions.ts
```diff
@@ -61,14 +62,13 @@ export const fetchChartFromServer = action("fetchChartFromServer", async (naviga
...
-  //let xmlSettings = jsonToXml(state.settingsJSON); // Convert JSON to XML using jsonToXml function
-  // console.log("XML Settings:", xmlSettings); // Log the XML settings to the console
   let body;
   try {
     if (state.settingsTabs.columns !== undefined) normalizeColumnProperties(state.settingsTabs.columns);
     const columnCopy: ColumnInfo = JSON.parse(JSON.stringify(state.settingsTabs.columns));
     changeManuallyAddedColumns(columnCopy);
-    const xmlSettings = jsonToXml(state.settingsTSC, columnCopy, state.settings);
+    const chartSettingsCopy: ChartSettings = JSON.parse(JSON.stringify(state.settings));
+    const xmlSettings = jsonToXml(columnCopy, chartSettingsCopy);
```
paologum, line 71: just curious why does this need to be a copy?

app/src/state/parse-settings.ts
```diff
@@ -249,37 +250,30 @@ export function xmlToJson(xml: string): ChartInfoTSC {
...
-function replaceSpecialChars(text: string, type: number): string {
+function escapeHtmlChars(text: string, type: "att" | "text"): string {
   text = text.replaceAll("&", "&amp;");
   text = text.replaceAll("<", " &lt; ");
-  if (type == 0) {
+  if (type == "att") {
     text = text.replaceAll('"', "&quot;");
     text = text.replaceAll("'", "&apos;");
     text = text.replaceAll(">", " &gt; ");
   }
```
paologum, line 270: can you make the type actually `attribute` instead of `att` so it's more clear

app/src/state/parse-settings.ts
```diff
@@ -342,124 +394,75 @@ function generateFontsXml(indent: string, fontsInfo?: FontsInfo): string {
...
+      //   } else
+      //     xml += `${indent}<setting name="${key}" standardized="${column[key].standardized}"
+      //   />\n`;
+      // } else
+      if (column[key].text.length > 0) {
+        xml += `${indent}<setting name="${key}" useNamed="false">${column[key].text}</setting>\n`;
       } else {
-        xml += `${indent}<setting name="${xmlKey}">${presetColumn[key as keyof ColumnInfoTSC]}</setting>\n`;
+        xml += `${indent}<setting name="${key}"/>\n`;
       }
```
paologum, line 433: technically for all xmls i've seen (and how rohan implemented on main), `backgorundColor` and `customColor` works like 

```
 else if (key === "backgroundColor") {
        if (stateColumn.rgb.r == 255 && stateColumn.rgb.g == 255 && stateColumn.rgb.b == 255) {
          xml += `${indent}<setting name="${xmlKey}"/>\n`;
        } else {
          xml += `${indent}<setting name="${xmlKey}" useNamed="false">rgb(${stateColumn.rgb.r},${stateColumn.rgb.g},${stateColumn.rgb.b})</setting>\n`;
        }
      } else if (key == "customColor") {
        xml += `${indent}<setting name="${xmlKey}" useNamed="false">rgb(${stateColumn.rgb.r},${stateColumn.rgb.g},${stateColumn.rgb.b})</setting>\n`;
      } 
```

app/src/state/parse-settings.ts
```diff
@@ -342,124 +394,75 @@ function generateFontsXml(indent: string, fontsInfo?: FontsInfo): string {
...
+        xml += columnInfoTSCToXml(column.children[i], `${indent}    `);
+        xml += `${indent}</column>\n`;
+      }
+    } else if (key === "justification") {
+      assertRulerColumnInfoTSC(column);
+      xml += `${indent}<setting justification="${column[key]}" name="justification"/>\n`;
+    } else if (key === "orientation") {
+      assertZoneColumnInfoTSC(column);
+      xml += `${indent}<setting name="orientation" orientation="${column[key]}"/>\n`;
+    } else {
```
paologum, line 450: don't these asserts mean you can now access `column.orientation` and `column.justification` now

shared/src/settings-types.ts
```diff
@@ -186,6 +187,7 @@ export function assertZoneColumnInfoTSC(o: any): asserts o is ZoneColumnInfoTSC
 
 export function assertEventColumnInfoTSC(o: any): asserts o is EventColumnInfoTSC {
   if (typeof o.type !== "string") throwError("EventColumnInfoTSC", "type", "string", o.type);
+  if (typeof o.rangeSort !== "string") throwError("EventColumnInfoTSC", "rangeSort", "string", o.type);
 }
 export function assertSequenceColumnInfoTSC(o: any): asserts o is SequenceColumnInfoTSC {
```
paologum, line 192: technically for this and all other values of `ColumnTSC` types that extend `ColumnBasicInfoTSC`, they should also call `assertColumnBasicInfoTSC`

paologum (changes requested):

app/src/state/actions/generate-chart-actions.ts
jay-jhk-lee: I don't think it has to be a copy, but since the column was made into a copy, I made a chart settings copy too. They're both states, so I wanted to keep it consistent.

app/src/state/parse-settings.ts
jay-jhk-lee: yeah

app/src/state/parse-settings.ts
jay-jhk-lee: the rgb text was taken care of the translate function before, but I'll standardize it to the rgb type for the text and do the same thing here

app/src/state/parse-settings.ts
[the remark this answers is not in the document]
jay-jhk-lee: I understand, but the chart title is also classified as the root column, and I directly use the id for the xml parsing. We do need to remove the `Chart Title in` prefix because If a settings file has "RootColumn:Chart title in ka(1000)" id instead of "RootColumn:ka (1000)" for example then the file doesn't recognize the id and overwrites it with the defaults.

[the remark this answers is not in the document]
paologum: what i'm saying is that i checked and it does not need to "recognize" the id. the id means nothing, it just differentiates the different column hierarchies. it works as it should regardless of the type of id in this specific case. the id must be different that is all. besides, the case you are aritculating would only occur if we read back in the columnInfo settingsXml we generated (which we have never done thus far) and in that case the column variable would have to be the same, meaning that it would need to have `Chart title in` for it to recognize the correct column in our website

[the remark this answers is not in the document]
paologum: ah wait i see, ok but you should still use just the
```
column._id = `class datastore.${state.columnDisplayType}:` + state.name;
```
and change it in `manuallyAddedColumns` in `generate-chart-actions` (i think it's in that file) and make the name correct there instead of here where you check if the name is like that because that's technically like a normalizing function

app/src/state/actions/generate-chart-actions.ts
paologum: sure i guess, the only reason that column was made a copy is that i change the manually added columns to their original names that would technically be duplicates in the hash map, so they have to be copied, then changed

app/src/state/parse-settings.ts
paologum: you can make it either way through checking if it's white through the string or through an rgb value, either way is fine, but the main thing is i think `backgroundColor` and `customColor` checks should be seperate, and `backgroundColor` should be empty when white to keep in line with the settings xml

app/src/state/parse-settings.ts
```diff
@@ -342,124 +412,81 @@ function generateFontsXml(indent: string, fontsInfo?: FontsInfo): string {
...
+        xml += `${indent}<column id="${escapeHtmlChars(column.children[i]._id, "attribute")}">\n`;
+        xml += columnInfoTSCToXml(column.children[i], `${indent}    `);
+        xml += `${indent}</column>\n`;
+      }
+    } else if (key === "justification") {
+      assertRulerColumnInfoTSC(column);
+      xml += `${indent}<setting justification="${column[key]}" name="justification"/>\n`;
+    } else if (key === "orientation") {
+      assertZoneColumnInfoTSC(column);
+      xml += `${indent}<setting name="orientation" orientation="${column[key]}"/>\n`;
```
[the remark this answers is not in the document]
jay-jhk-lee: yeah I'll change it

app/src/state/parse-settings.ts
```diff
@@ -308,6 +319,65 @@ function generateSettingsXml(stateSettings: ChartSettings, indent: string): stri
...
+      column._id = `class datastore.${state.columnDisplayType}:` + state.name;
+      break;
+    default:
+      column._id = `class datastore.${state.columnDisplayType}Column:` + state.name;
+  }
+  column.title = escapeHtmlChars(state.editName, "text");
+  column.isSelected = state.on;
+  column.drawTitle = state.enableTitle;
+  column.fonts = state.fontsInfo;
+  if (state.width) column.width = state.width;
```
paologum, line 358: you should make the `ColumnTSC` type have width be optional, because leyton and i made the `columninfo` variable have optional width so it wouldn't be put in the xml as a tag. right now the width is going to be the default if it's a parent which could be mislead, so i think it should be optional and not be put in the xml if it's not there

paologum (changes requested):

app/src/state/parse-settings.ts
jay-jhk-lee: Since ColumnTSC is supposed to be a direct representation of the settings file, I think making the width property optional on it is not a good idea since every column in a settings file has it even though the column itself might not use it (like the parent columns). I'll set the default to -1 which will clearly indicate that the width hasn't been set for this. The generated settings file from the java program will autofill the width either way. Lmk what you think

app/src/state/parse-settings.ts
```diff
@@ -342,124 +411,87 @@ function generateFontsXml(indent: string, fontsInfo?: FontsInfo): string {
...
-      } else if (key === "drawTitle") {
-        xml += `${indent}<setting name="${xmlKey}">${stateColumn.enableTitle}</setting>\n`;
+    } else if (key === "customColor") {
+      xml += `${indent}<setting name="customColor" useNamed="false">rgb(${column.customColor.text.r},${column.customColor.text.g},${column.customColor.text.b})</setting>\n`;
+    } else if (key === "width") {
+      if (column.width) {
+        xml += `${indent}<setting name="width">${column.width}</setting>\n`;
       } else {
-        xml += `${indent}<setting name="${xmlKey}">${presetColumn[key as keyof ColumnInfoTSC]}</setting>\n`;
+        xml += `${indent}<setting name="width"></setting>\n`;
```
paologum, line 461: i think it'd be better to just not have the tag, i've verified it works without the tag, so only add the tag if the width exists

app/src/state/parse-settings.ts
```diff
@@ -308,6 +319,64 @@ function generateSettingsXml(stateSettings: ChartSettings, indent: string): stri
...
+      column._id = `class datastore.${state.columnDisplayType}:` + state.name;
+      break;
+    default:
+      column._id = `class datastore.${state.columnDisplayType}Column:` + state.name;
+  }
+  column.title = escapeHtmlChars(state.editName, "text");
+  column.isSelected = state.on;
+  column.drawTitle = state.enableTitle;
+  column.fonts = state.fontsInfo;
+  column.width = column.width ? column.width : undefined;
```
paologum, line 358: this can be just `column.width = column.width` right?

paologum (changes requested):

paologum (approved):

PR #88, opened 2024-04-17
 font inheritance, enable title, more error checking, column types, lone columns

paologum:
- fixed font changes with inheritance and backend restructuring
- font inheritance works, (doesn't show the children inheriting on the front end but changes xml to chart propogation)
- added enable title checkbox
- removed some general parse setting assumptions
- added error if no columns are selected on generation
- added error if all charts have unit ranges of 0 (top: 0, base: 0)
- added column types for display (Zone, MetaColumn, etc)
- refactored a lot of the code to now parse datapacks that are LONE which means they do not have a parent-child hierarchy

https://github.com/earthhistoryviz/tsconline/assets/89664317/249ac527-046f-4bb5-8143-6e8287c59d73


jay-jhk-lee (approved):
Looks Good!

asivath (approved):

PR #89, opened 2024-04-19
TSC-276-description objects

rupp3:
Changed DatapackParsingPack and activeDatapack. Still getting a type error in routes.ts when the loadIndexes function is called, not exactly sure what to put in replacement of [filename]. Previously had all of the components of DatapackParsingPack being set to an empty string or array and file set to filename but after rebase and yarn build it gave more errors so I currently have it blank.


rupp3:
Changed the parse-datapacks test file and the json file that holds the keys to include file, description, title, and size. I made each parameter in the json file to be equal to empty strings but I don't think that's correct? I tried replacing the empty strings with what is in activeDatapacks for the Africa Bight datapack in both the test file and the json file but either way failed. Have screenshots: 
Screenshot 2024-04-25 004136
Screenshot 2024-04-25 004350
Screenshot 2024-04-25 003918

app/vite.config.ts.timestamp-1712938232587-38897d5850a4f.mjs
paologum, line 1: what is this file?

server/src/index.ts
```diff
@@ -42,13 +42,18 @@ try {
...
-    (datapack) => '"' + assetconfigs.datapacksDirectory + "/" + datapack + '"'
-  );
+  const datapacks = assetconfigs.activeDatapacks.map((datapack) => ({
+    title: datapack.title,
+    description: datapack.description,
+    file: datapack.file,
+    size: datapack.size
+  }));
+
+  const datapackPaths = datapacks.map((datapack) => '"' + assetconfigs.datapacksDirectory + "/" + datapack.file + '"');
```
paologum, line 52: i think you can just do this in one statement, you don't need to create `datapacks`, you can just make the datapack array in one.

server/src/load-packs.ts
```diff
@@ -31,29 +32,29 @@ export async function loadIndexes(
...
   mapPackIndex: MapPackIndex,
   decryptionDirectory: string,
-  datapacks: string[]
+  datapacks: DatapackParsingPack[]
 ) {
-  console.log(`\nParsing datapacks: ${datapacks}\n`);
   for (const datapack of datapacks) {
-    await parseDatapacks(datapack, decryptionDirectory)
+    console.log(`\nParsing datapacks: ${datapack.file}\n`);
+    await parseDatapacks(datapack.file, datapack.title, datapack.description, datapack.size, decryptionDirectory)
```
paologum, line 39: instead of adding each individual field, it would be better in the long run to just make the parameter the datapack object

server/src/load-packs.ts
```diff
@@ -31,29 +32,29 @@ export async function loadIndexes(
   datapackIndex: DatapackIndex,
   mapPackIndex: MapPackIndex,
   decryptionDirectory: string,
-  datapacks: string[]
+  datapacks: DatapackParsingPack[]
 ) {
-  console.log(`\nParsing datapacks: ${datapacks}\n`);
```
paologum, line 36: you should keep this

server/src/routes.ts
```diff
@@ -167,19 +193,7 @@ export const uploadDatapack = async function uploadDatapack(
...
-    } catch (e) {
-      console.error(e);
-      await resetUploadDirectory(filepath, decryptedFilepathDir);
-      reply.status(500).send({ error: "Failed to parse MapPackIndex.json" });
-      return;
-    }
-  }
-  await loadIndexes(datapackIndex, mapPackIndex, decryptDir.replaceAll("\\", "/"), [filename]);
+  await loadIndexes(datapackIndex, mapPackIndex, decryptDir.replaceAll("\\", "/"), [filePack]);
   if (!datapackIndex[filename]) {
```
paologum, line 197: why is this removed

server/src/load-packs.ts
```diff
@@ -31,29 +32,29 @@ export async function loadIndexes(
   datapackIndex: DatapackIndex,
   mapPackIndex: MapPackIndex,
   decryptionDirectory: string,
-  datapacks: string[]
+  datapacks: DatapackParsingPack[]
```
paologum, line 35: this shouldn't be `DatapackParsingPack` because the objective of this function is to create `DatapackParsingPack`'s so instead you should take the datapack description items of the object and put it into it's own object that can be fed into this method

server/src/routes.ts
```diff
@@ -90,6 +91,31 @@ export const uploadDatapack = async function uploadDatapack(
...
+      width: 0,
+      units: ""
+    },
+    datapackAgeInfo: { datapackContainsSuggAge: false },
+    ageUnits: "",
+    description: "",
+    title: "",
+    file: filename,
+    size: ""
+  };
```
paologum, line 118: that way with the above comment, you don't make this slightly confusing object that has a `columnInfo` object that doesn't make sense

server/src/types.ts
```diff
@@ -1,8 +1,8 @@
-import { throwError } from "@tsconline/shared";
+import { throwError, DatapackParsingPack } from "@tsconline/shared";
 
 export type AssetConfig = {
   activeJar: string;
-  activeDatapacks: string[];
+  activeDatapacks: DatapackParsingPack[];
```
paologum, line 5: this would change to whatever that object is called

shared/src/index.ts
```diff
@@ -16,6 +16,10 @@ export type DatapackParsingPack = {
   columnInfo: ColumnInfo;
   datapackAgeInfo: DatapackAgeInfo;
   ageUnits: string;
+  description: string;
+  title: string;
+  file: string;
+  size: string;
```
paologum, line 22: these would be that object

paologum (changes requested):

app/vite.config.ts.timestamp-1712938232587-38897d5850a4f.mjs
rupp3: honestly no clue, did not know it was there

server/src/routes.ts
rupp3: i changed filename to filePack bc of having loadIndexes accepting a DatapackParsingPack as the last arg

app/src/SettingsTabs/Datapack.tsx
```diff
@@ -52,7 +52,7 @@ export const Datapacks = observer(function Datapacks() {
                   </td>
                   <td className="info-cell">
                     <div>
-                      <Tooltip title="Description" arrow placement="right">
+                      <Tooltip title={"state.datapackInfo[index].description"} arrow placement="right">
```
paologum, line 55: i don't this displays the description if it's in quotes right

server/__tests__/parse-datapacks.test.ts
```diff
@@ -61,13 +62,13 @@ import {
   ColumnHeaderProps
 } from "@tsconline/shared";
 const key = JSON.parse(readFileSync("server/__tests__/__data__/column-keys.json").toString());
-
+const datapack: DatapackDescriptionInfo = { description: "", title: "", file: "", size: "" };
```
paologum, line 65: this should be in the describe block below instead of the top of the test

server/src/routes.ts
```diff
@@ -167,20 +172,8 @@ export const uploadDatapack = async function uploadDatapack(
...
-      assertMapPackIndex(mapPackIndex);
-    } catch (e) {
-      console.error(e);
-      await resetUploadDirectory(filepath, decryptedFilepathDir);
-      reply.status(500).send({ error: "Failed to parse MapPackIndex.json" });
-      return;
-    }
-  }
-  await loadIndexes(datapackIndex, mapPackIndex, decryptDir.replaceAll("\\", "/"), [filename]);
-  if (!datapackIndex[filename]) {
```
paologum, line 183: is there a reason you removed the `fs.existsSync(mapPackIndexFilepath)` check

app/src/SettingsTabs/Datapack.tsx
paologum: oh wait i get why, but technically this pr should have this fixed in the end for sure

shared/src/index.ts
```diff
@@ -18,6 +18,13 @@ export type DatapackParsingPack = {
   ageUnits: string;
 };
 
+export type DatapackDescriptionInfo = {
+  description: string;
+  title: string;
+  file: string;
+  size: string;
```
paologum, line 25: so technically, you need the `DatapackDescriptionInfo` object for the information we keep in the json file, so that is good. But we also need to have it represented in `DatapackParsingPack`. Which means you should have this object inside `DatapackParsingPack` so we can access it on the app side.

server/src/load-packs.ts
```diff
@@ -112,11 +114,16 @@ export async function loadFaciesPatterns() {
...
-  datapacks: string[] = assetconfigs.activeDatapacks,
+  datapacks: DatapackDescriptionInfo[] = assetconfigs.activeDatapacks,
   decryptionDirectory: string = assetconfigs.decryptionDirectory
 ) {
-  const imagePaths = await grabFilepaths(datapacks, decryptionDirectory, "MapImages");
+  const imagePaths = await grabFilepaths(
+    datapacks.map((datapack) => datapack.file),
+    decryptionDirectory,
+    "MapImages"
+  );
```
paologum, line 126: i think you should just leave this as `string[]` and instead when you call `await grabMapImages` you do the map to filepath stuff there

server/src/parse-datapacks.ts
```diff
@@ -116,10 +117,13 @@ export function spliceArrayAtFirstSpecialMatch(array: string[]): ParsedColumnEnt
  * @param files the files to be parsed
  * @returns
  */
-export async function parseDatapacks(file: string, decryptFilePath: string): Promise<DatapackParsingPack | null> {
-  const decryptPaths = await grabFilepaths([file], decryptFilePath, "datapacks");
+export async function parseDatapacks(
+  datapackInfo: DatapackDescriptionInfo,
+  decryptFilePath: string
+): Promise<DatapackParsingPack | null> {
+  const decryptPaths = await grabFilepaths([datapackInfo.file], decryptFilePath, "datapacks");
```
paologum, line 124: here you pass in `datapackInfo` but you don't do anything with it in this method

server/src/types.ts
```diff
@@ -72,7 +72,11 @@ export function assertAssetConfig(o: any): asserts o is AssetConfig {
   if (typeof o.uploadDirectory !== "string") throw 'AssetConfig must have a "uploadDirectory" string';
   if (!o.activeDatapacks || !Array.isArray(o.activeDatapacks)) throw 'AssetConfig must have an "activeJar" string';
   for (const [index, ad] of o.activeDatapacks.entries()) {
-    if (typeof ad !== "string") throw "AssetConfig activeDatapacks item " + index + " must be a string";
+    if (typeof ad !== "object") throw "AssetConfig activeDatapacks item " + index + " must be an object";
+    for (const key in ad) {
+      if (typeof ad[key] !== "string")
+        throw "AssetConfig activeDatapacks item " + index + ", property" + key + "must be a string";
+    }
```
paologum, line 79: wouldn't you just call your method `assertDatapackDescriptionInfo` here

paologum (changes requested):

server/src/routes.ts
rupp3: i'm thinking that when i rebase it removes things because i didn't remove that myself/on purpose

paologum: okok well if you rebase it'll have a conflict with this that you should change to the main version

server/src/parse-datapacks.ts
rupp3: datapackInfo is passed in to access the files to be called in grabFilepaths(). I just replaced files with datapackInfo.file

paologum: you can probably just keep it to `files: string` and just pass the property itself instead of the whole object for better readability

server/src/routes.ts
```diff
@@ -339,7 +340,7 @@ export const fetchChart = async function fetchChart(request: FastifyRequest, rep
   const userDatapackNames = userDatapackFilepaths.map((datapack) => path.basename(datapack));
   const datapacks = [];
   for (const datapack of chartrequest.datapacks) {
-    if (assetconfigs.activeDatapacks.includes(datapack)) {
+    if (assetconfigs.activeDatapacks.map((datapack) => datapack.file).includes(datapack)) {
```
paologum, line 343: this is gonna execute map multiple times for every datapack, you should just map once

server/__tests__/parse-datapacks.test.ts
```diff
@@ -76,15 +77,15 @@ describe("general parse-datapacks tests", () => {
    * Checks both datapack ages and columnInfo values
    */
   it("should parse general datapack with all column types", async () => {
-    const datapacks = await parseDatapacks("", "parse-datapacks-test-2.txt");
+    const datapacks = await parseDatapacks(datapack, "parse-datapacks-test-2.txt");
```
paologum, line 80: these should change back

server/src/parse-datapacks.ts
```diff
@@ -251,7 +257,7 @@ export async function parseDatapacks(file: string, decryptFilePath: string): Pro
     units: ageUnits,
     columnDisplayType: "RootColumn"
   };
-  return { columnInfo: chartColumn, datapackAgeInfo, ageUnits };
+  return { columnInfo: chartColumn, datapackAgeInfo, ageUnits, file, description };
```
paologum, line 260: if you're going to add all the description stuff here, you should pass the `DatapackDescription` object in and add each param here, just make sure that `DatapackParsingPack` has all the corresponding items

paologum (commented):
thought i'd add some more comments

server/src/parse-datapacks.ts
rupp3: should i even be putting the description stuff in parseDatapacks? it feels like it does not need to be there but i am not sure where to put it. same goes for the app side, i still cant access the description to display on the frontend

server/src/parse-datapacks.ts
```diff
@@ -114,10 +114,10 @@ export function spliceArrayAtFirstSpecialMatch(array: string[]): ParsedColumnEnt
  * @param files the files to be parsed
  * @returns
  */
-export async function parseDatapacks(file: string, decryptFilePath: string): Promise<DatapackParsingPack | null> {
-  const decryptPaths = await grabFilepaths([file], decryptFilePath, "datapacks");
+export async function parseDatapacks(files: string, decryptFilePath: string): Promise<DatapackParsingPack | null> {
+  const decryptPaths = await grabFilepaths([files], decryptFilePath, "datapacks");
```
paologum, line 118: you can't append the required fields if you just send in the `files` here. i said something a long time ago about setting this to files so that's my fault but this should be your object so you can add the necessary parts to `DatapackParsingPack`

server/assets/config.json
paologum, line 1: this is missing vertebrates and human culture

shared/src/index.ts
```diff
@@ -30,6 +30,17 @@ export type DatapackParsingPack = {
...
+};
+
+export type DatapackDescriptionInfo = {
+  description: string;
+  title: string;
+  file: string;
+  size: string;
```
paologum, line 43: technically you aren't using `DatapackDescriptionInfo` on `app`, so this should be in `types.ts`

shared/src/index.ts
```diff
@@ -671,6 +682,15 @@ export function assertDatapackParsingPack(o: any): asserts o is DatapackParsingP
...
+
+export function assertDatapackDescriptionInfo(o: any): asserts o is DatapackDescriptionInfo {
+  if (!o || typeof o !== "object") throw new Error("DatapackDescriptionInfo must be a non-null object");
+  if (typeof o.description !== "string") throw new Error("DatapackDescriptionInfo description must be of type string");
+  if (typeof o.title !== "string") throw new Error("DatapackDescriptionInfo title must be of type string");
+  if (typeof o.file !== "string") throw new Error("DatapackDecriptionInfo file must be of type string");
+  if (typeof o.size !== "string") throw new Error("DatapackDescriptionInfo size must be of type string");
+}
+
 export function assertDatapackIndex(o: any): asserts o is DatapackIndex {
```
paologum, line 694: same

server/assets/config.json
paologum, line 1: still missing vertebrates and human culture

server/src/load-packs.ts
```diff
@@ -31,32 +31,39 @@ export async function loadIndexes(
...
   decryptionDirectory: string,
-  datapacks: string[]
+  datapacks: DatapackDescriptionInfo[]
 ) {
   console.log(`\nParsing datapacks: ${datapacks}\n`);
   for (const datapack of datapacks) {
-    await parseDatapacks(datapack, decryptionDirectory)
+    console.log(`\nParsing datapack file: ${datapack.file}\n`);
+    await parseDatapacks(
+      { file: datapack.file, description: datapack.description, title: datapack.title, size: datapack.size },
```
paologum, line 40: i think you can just pass in `datapack`

server/src/parse-datapacks.ts
```diff
@@ -132,6 +136,11 @@ export async function parseDatapacks(file: string, decryptFilePath: string): Pro
   const freehandMap: Map<string, Freehand> = new Map();
   const blankMap: Map<string, ColumnHeaderProps> = new Map();
   const loneColumns: ColumnInfo[] = [];
+  const description: string = datapackInfo.description;
+  const file: string = datapackInfo.file;
+  const title: string = datapackInfo.title;
+  const size: string = datapackInfo.size;
```
paologum, line 142: you don't need to create temp variables for these, you can just go straight to the return

server/src/routes.ts
```diff
@@ -149,6 +150,12 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
   const decryptedFilepathDir = path.join(decryptDir, filenameWithoutExtension);
   const mapPackIndexFilepath = path.join(userDir, "MapPackIndex.json");
   const datapackIndexFilepath = path.join(userDir, "DatapackIndex.json");
+  const datapackInfo: DatapackDescriptionInfo = {
+    file: filename,
+    description: "",
+    title: "",
+    size: ""
+  };
   async function errorHandler(message: string, errorStatus: number, e?: unknown) {
```
paologum, line 159: since you created the new fields, the app has some of these forms already built on the front end so can you send those and input them here

server/src/routes.ts
```diff
@@ -386,9 +393,10 @@ export const fetchChart = async function fetchChart(request: FastifyRequest, rep
   const userDatapackNames = userDatapackFilepaths.map((datapack) => path.basename(datapack));
   const datapacks = [];
   const userDatapacks = [];
+  const datapackFiles = assetconfigs.activeDatapacks.map((datapack) => datapack.file);
 
   for (const datapack of chartrequest.datapacks) {
-    if (assetconfigs.activeDatapacks.includes(datapack)) {
+    if (datapackFiles.includes(datapack)) {
```
paologum, line 399: for ease of readability can you name it `serverDatapacks`

paologum (changes requested):
this is looking really good! try to rebase for your next push

server/src/routes.ts
rupp3: not 100% sure what I need to send. I've been using DatapackIndex to access each field but it has not been working here. where on the front end are they getting built?

server/__tests__/parse-datapacks.test.ts
```diff
@@ -133,13 +133,15 @@ import {
   Freehand,
   ColumnHeaderProps
 } from "@tsconline/shared";
+import { DatapackDescriptionInfo } from "../src/types";
 const key = JSON.parse(readFileSync("server/__tests__/__data__/column-keys.json").toString());
 describe("general parse-datapacks tests", () => {
   /**
    * Parses the general Africa Bight map pack
    */
+  const datapack: DatapackDescriptionInfo = { description: "", title: "", file: "", size: "" };
```
paologum, line 142: you should make the strings for the test variables not empty and make them different just from a good test standpoint

server/src/routes.ts
```diff
@@ -290,6 +290,7 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
   const decryptedFilepathDir = path.join(decryptDir, filenameWithoutExtension);
   const mapPackIndexFilepath = path.join(userDir, "MapPackIndex.json");
   const datapackIndexFilepath = path.join(userDir, "DatapackIndex.json");
+  //const serverdatapackInfo: DatapackIndex = {};
```
paologum, line 293: take out?

server/src/parse-datapacks.ts
```diff
@@ -282,9 +284,22 @@ export async function parseDatapacks(
...
   };
   setShowLabels(chartColumn);
-
-  const datapackParsingPack = { columnInfo: chartColumn, ageUnits, defaultChronostrat, formatVersion, isUserDatapack };
-
+  try {
+    assertDatapackDescriptionInfo(datapackInfo);
+  } catch (e) {
+    console.log(`Error ${e} found while processing DatapackDescriptionInfo`);
+  }
```
paologum, line 291: i don't think you need this because you  pass in the value as the type `DatapackDescriptionInfo` already

server/src/routes.ts
paologum: well this function is concerned with the upload of  a datapack meaning it doesn't exist on the `DatapackIndex` until this successfully completes. Therefore meaning it should have the proper fields set in the `DatapackParsingPack` to complete successfully.

package.json
```diff
@@ -6,6 +6,7 @@
   "devDependencies": {
     "@types/eslint": "^8",
     "@types/jest": "^29.5.12",
+    "@types/lodash": "^4.17.1",
```
paologum, line 9: did you add this type? if not i would run `yarn remove @types/lodash` in the root directory

server/__tests__/parse-datapacks.test.ts
```diff
@@ -153,15 +153,21 @@ import {
...
+    description: "Africa Bight Map",
+    file: "AfricaBight.map",
+    size: "200.98 KB"
+  };
   it("should parse africa general datapack", async () => {
-    const datapacks = await parseDatapacks("", "parse-datapacks-test-1.txt");
-    // console.log(JSON.stringify(datapacks, null, 2))
+    const datapacks = await parseDatapacks(datapack1, "parse-datapacks-test-1.txt");
     expect(datapacks).toEqual(key["general-parse-datapacks-test-1-key"]);
   });
```
paologum, line 172: i would put `datapack1` in the `it` test and call it a better name like `africaGeneralDescriptionInfo` or alternatively don't use any variable and pass it in raw

server/__tests__/parse-datapacks.test.ts
paologum: you should just name the strings like `"description"`, etc just for 100% verification

server/src/database.ts
```diff
@@ -69,6 +69,8 @@ await db.schema
   .addColumn("verifyOrReset", "text", (col) => col.notNull())
   .execute();
 
+//************** COMMENTED OUT DUE TO MIGRATIONS FILE PATH ERROR DELETE BEFORE MERGE***********
+/*
```
paologum, line 73: no comment

server/src/database.ts
```diff
@@ -1,7 +1,7 @@
 import { Database, User, NewUser, UpdatedUser, Verification, NewVerification } from "./types.js";
 import BetterSqlite3 from "better-sqlite3";
 import { Kysely, SqliteDialect } from "kysely";
-import { exec } from "child_process";
+//import { exec } from "child_process";
```
paologum, line 4: no comment

server/src/database.ts
```diff
@@ -80,7 +82,7 @@ await new Promise<void>((resolve, reject) => {
     if (stderr) console.error(stderr);
     resolve();
   });
-});
+});*/
```
paologum, line 85: no comment

server/src/load-packs.ts
```diff
@@ -32,33 +32,42 @@ export async function loadIndexes(
...
         }
         assertDatapackParsingPack(datapackParsingPack);
-        datapackIndex[datapack] = datapackParsingPack;
-        console.log(chalk.green(`Successfully parsed ${datapack}`));
+        datapackIndex[datapack.file] = {
+          ...datapackParsingPack,
+          description: datapack.description,
+          title: datapack.title,
+          size: datapack.size
+        };
```
paologum, line 52: `datapackParsingPack` should already have those fields properly set

server/src/load-packs.ts
```diff
@@ -114,8 +123,9 @@ export async function loadFaciesPatterns() {
  * Finds all map images and puts them in the public directory
  * For access from fastify server servicing
  */
+//changed datapacks to DatapackDescriptionInfo[] to match the type in the function
 export async function grabMapImages(
-  datapacks: string[] = assetconfigs.activeDatapacks,
+  datapacks: string[] = assetconfigs.activeDatapacks.map((datapack) => datapack.file),
```
paologum, line 128: should remove the comment since it doesn't apply

server/src/parse-datapacks.ts
```diff
@@ -258,8 +259,17 @@ export async function parseDatapacks(
...
-
+  const datapackParsingPack = {
+    columnInfo: chartColumn,
+    ageUnits,
+    defaultChronostrat,
+    formatVersion,
+    description: datapackInfo.description,
+    title: datapackInfo.title,
+    file: datapackInfo.file,
+    size: datapackInfo.size,
```
paologum, line 270: if you wanted you could also do `...datapackInfo` i think but this is arbitrary so you don't have to change it, just thought i'd comment

server/src/routes.ts
```diff
@@ -289,6 +290,12 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
   const decryptedFilepathDir = path.join(decryptDir, filenameWithoutExtension);
   const mapPackIndexFilepath = path.join(userDir, "MapPackIndex.json");
   const datapackIndexFilepath = path.join(userDir, "DatapackIndex.json");
+  const datapackInfo: DatapackDescriptionInfo = {
+    file: filename,
+    description: "",
+    title: filenameWithoutExtension,
+    size: file.file.bytesRead.toString()
+  };
```
paologum, line 298: this pr should also send/receive the user's inputed title and description from the frontend

server/src/types.ts
```diff
@@ -137,9 +152,5 @@ export function assertAssetConfig(o: any): asserts o is AssetConfig {
   if (typeof o.colors !== "string") throw 'AssetConfig must have a "colors" string';
   if (typeof o.fileMetadata !== "string") throw 'AssetConfig must have a "fileMetadata" string';
   if (typeof o.uploadDirectory !== "string") throw 'AssetConfig must have a "uploadDirectory" string';
-  if (!o.activeDatapacks || !Array.isArray(o.activeDatapacks)) throw 'AssetConfig must have an "activeJar" string';
-  for (const [index, ad] of o.activeDatapacks.entries()) {
-    if (typeof ad !== "string") throw "AssetConfig activeDatapacks item " + index + " must be a string";
-  }
   if (typeof o.timescaleFilepath !== "string") throw 'AssetConfig must have a "timescaleFilepath" string';
```
paologum, line 155: don't you need this check to make sure there are `activeDatapacks`?

paologum (changes requested):

package.json
rupp3: was previously getting errors for not finding lodash but works fine without now

paologum:
can you upload a demo of you uploading and working with the datapack page after a file is uploaded

server/src/routes.ts
```diff
@@ -264,21 +270,45 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  let tempFilePath: string;
+  const fields: Record<string, string> = {};
+  let uploadedFile: MultipartFile | undefined;
+
+  for await (const part of parts) {
+    if (part.type === "file") {
+      uploadedFile = part;
+      tempFilePath = path.join(os.tmpdir(), uploadedFile.filename);
+      const tempFileStream = fs.createWriteStream(tempFilePath);
+      await pump(uploadedFile.file, tempFileStream); //need to ensure the file is written
```
paologum, line 283: i'm not sure why we need to write it in a temp directory?
why don't you just set `uploadedFile` to the part and then let the code written before handle `uploadedFile` (the way it was `await pump` ing before instead of the overhead of temporary directories

server/src/routes.ts
```diff
@@ -264,21 +270,45 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  const description = fields.description;
+
+  if (!uploadedFile) {
     reply.status(404).send({ error: "No file uploaded" });
+    console.log("No file uploaded, console log");
+    return;
+  }
+  if (!name || !description) {
+    reply.status(404).send({ error: "No name or description uploaded" });
+    console.log("No name or description uploaded");
```
paologum, line 299: no logs

server/src/routes.ts
```diff
@@ -315,7 +349,15 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
-  if (file.file.truncated) {
+  try {
+    const state = await fs.promises.stat(filepath);
+    console.log("File size: ", state.size);
+  } catch (e) {
+    console.error("Error getting file size: ", e);
+    console.log("File not found after save, error:", e);
+    return;
+  }
+  if (uploadedFile.file.truncated) {
```
paologum, line 360: i would check size after we see if it's too large, also remove all logs and errors, and just return the correct response with the correct error codes (probably should look up the error codes)

server/src/routes.ts
```diff
@@ -376,14 +419,16 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  await loadIndexes(datapackIndex, mapPackIndex, decryptDir.replaceAll("\\", "/"), [datapackInfo], true);
   if (!datapackIndex[filename]) {
     await errorHandler("Failed to load decrypted datapack", 500);
     return;
   }
   try {
     await writeFile(datapackIndexFilepath, JSON.stringify(datapackIndex));
+    console.log("Wrote datapack index to file", datapackIndexFilepath);
     await writeFile(mapPackIndexFilepath, JSON.stringify(mapPackIndex));
+    console.log("Wrote map pack index to file");
```
paologum, line 431: remove logs

server/src/routes.ts
```diff
@@ -397,11 +442,13 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
       mapPackIndexFilepath,
       datapackIndexFilepath
     );
+    console.log("metadata to file");
   } catch (e) {
     await errorHandler("Failed to load and write metadata for file", 500, e);
     return;
   }
   reply.status(200).send({ message: "File uploaded" });
+  console.log("File uploaded to server");
```
paologum, line 451: remove logs

server/src/types.ts
```diff
@@ -140,7 +155,15 @@ export function assertAssetConfig(o: any): asserts o is AssetConfig {
...
-    if (typeof ad !== "string") throw "AssetConfig activeDatapacks item " + index + " must be a string";
+    if (typeof ad !== "object")
+      throw "AssetConfig activeDatapacks item " + index + " must be a valid DatapackDescriptionInfo object";
+    if (
+      typeof ad.description !== "string" ||
+      typeof ad.title !== "string" ||
+      typeof ad.file !== "string" ||
+      typeof ad.size !== "string"
+    )
+      throw "AssetConfig activeDatapacks description, title, file, or size item " + index + " must be a valid string";
```
paologum, line 166: i would make each of these their own `if` statement

paologum (changes requested):

paologum:
i would also upload a demo here

server/src/database.ts
```diff
@@ -83,7 +83,7 @@ export async function initializeDatabase() {
       if (stderr) console.error(stderr);
       resolve();
     });
-  });
+  });*/
```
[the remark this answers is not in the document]
rupp3: NOTE: uncommenting at very end bc this makes my server not run

server/src/load-packs.ts
```diff
@@ -31,33 +31,37 @@ export async function loadIndexes(
   datapackIndex: DatapackIndex,
   mapPackIndex: MapPackIndex,
   decryptionDirectory: string,
-  datapacks: string[],
+  datapacks: DatapackDescriptionInfo[],
   userUploaded?: boolean
 ) {
   console.log(`\nParsing datapacks: ${datapacks}\n`);
   for (const datapack of datapacks) {
+    console.log(`\nParsing datapack file: ${datapack.file}\n`);
```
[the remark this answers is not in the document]
rupp3: output of line 37 is 
Parsing datapacks: [object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object] so i am going to remove the datapacks variable as well, just keeping the string

server/src/routes.ts
rupp3: when i had it just line 280 and the original pump, the server would hang in the if block. apparently need to write the file somewhere first in order to use it so i made a temp and then had it rewritten to the correct filepath

server/src/routes.ts
rupp3: i can just delete it. only used it to check if the file existed when i was trying the pump stuff. or i can make an error for an empty file being uploaded if that's relevant?

rupp3:
Demo video
https://github.com/earthhistoryviz/tsconline/assets/54484966/b97e7389-4f44-44ad-ab1d-9989263ec019

paologum:
from the demo, you can see that the description being smaller than one line makes the description resize. you should make that container take up the whole parent container. (not the whole page but just the container it is in)

app/src/components/datapack_display/TSCDatapackCard.tsx
```diff
@@ -18,6 +18,10 @@ export const TSCDatapackCard: React.FC<TSCDatapackCardProps> = ({ name, datapack
   const navigate = useNavigate();
   const [loading, setLoading] = useState(false);
   const defaultImageUrl = devSafeUrl("/datapack-images/default.png");
+  if (!datapack) {
+    console.error("Datapack is undefined");
+    return null;
+  }
```
paologum, line 24: this case should never happen because the prop is always defined

app/src/components/datapack_display/TSCDatapackRow.tsx
```diff
@@ -23,6 +23,10 @@ export const TSCDatapackRow: React.FC<TSCDatapackRowProps> = ({ name, datapack,
   const navigate = useNavigate();
   const theme = useTheme();
   const defaultImageUrl = devSafeUrl("/datapack-images/default.png");
+  if (!datapack) {
+    console.error("Datapack is undefined");
+    return null;
+  }
```
paologum, line 29: same here

server/src/routes.ts
```diff
@@ -62,7 +66,7 @@ export const requestDownload = async function requestDownload(
   request: FastifyRequest<{ Params: { filename: string }; Querystring: { needEncryption?: boolean } }>,
   reply: FastifyReply
 ) {
-  const uuid = request.session.get("uuid");
+  const uuid = "username"; //request.session.get("uuid");
```
paologum, line 69: not supposed to be username

server/src/routes.ts
```diff
@@ -25,6 +25,10 @@ import { loadIndexes } from "./load-packs.js";
 import { updateFileMetadata, writeFileMetadata } from "./file-metadata-handler.js";
 import { datapackIndex as serverDatapackindex, mapPackIndex as serverMapPackIndex } from "./index.js";
 import { glob } from "glob";
+import { DatapackDescriptionInfo } from "./types.js";
+import { MultipartFile } from "@fastify/multipart";
+import * as os from "os";
+//import { stringify } from "querystring";
```
paologum, line 31: comments?

server/src/database.ts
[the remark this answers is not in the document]
paologum: i wouldn't commit these comments, i would change them back and just keep these changes local and not commit them so that we don't accidentally merge undesired changes

server/src/routes.ts
```diff
@@ -225,19 +229,21 @@ export const fetchServerMapPackInfo = async function fetchServerMapPackInfo(
 
 export const fetchUserDatapacks = async function fetchUserDatapacks(request: FastifyRequest, reply: FastifyReply) {
   // for test usage: const uuid = "username";
-  const uuid = request.session.get("uuid");
+  const uuid = "username"; //request.session.get("uuid");
```
paologum, line 232: same

server/src/routes.ts
```diff
@@ -258,27 +264,49 @@ export const fetchUserDatapacks = async function fetchUserDatapacks(request: Fas
 
 // If at some point a delete datapack function is needed, this function needs to be modified for race conditions
 export const uploadDatapack = async function uploadDatapack(request: FastifyRequest, reply: FastifyReply) {
-  const uuid = request.session.get("uuid");
+  const uuid = "username"; //request.session.get("uuid");
```
paologum, line 267: same

server/src/routes.ts
paologum: then instead of making a temp file, just do the actual file saving stuff here

paologum (changes requested):

server/src/routes.ts
paologum: don't you need to set the `size`?

server/src/routes.ts
```diff
@@ -293,17 +321,21 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
     await resetUploadDirectory(filepath, decryptedFilepathDir);
     reply.status(errorStatus).send({ error: message });
   }
+  const datapackInfo: DatapackDescriptionInfo = {
+    file: filename,
+    description: description,
+    title: name,
+    size: uploadedFile.file.bytesRead.toString()
```
paologum, line 328: does this put it in the format of 30mb or like 2kb?
if not, it should because this should be a user readable field

rupp3: output is just a number like "size in datapackInfo: 16965905"

paologum: ok yeah you should convert it properly to a file format string, because that doesen't help the user

server/src/routes.ts
paologum: i was using postman to test a different upload route i'm making and found that

```
 for await (const part of parts) {
    if (part.type === "file") {
      file = part;
    } else if (part.fieldname === "title" && typeof part.value === "string") {
      title = part.value;
    } else if (part.fieldname === "description" && typeof part.value === "string") {
      description = part.value;
    }
  }
```
doesn't hang for me

rupp3: it still hangs in the file if block

server/src/routes.ts
```diff
@@ -264,30 +268,67 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
-    file.mimetype !== "application/octet-stream" &&
-    file.mimetype !== "text/plain" &&
-    file.mimetype !== "application/zip"
+    uploadedFile.mimetype !== "application/octet-stream" &&
+    uploadedFile.mimetype !== "text/plain" &&
+    uploadedFile.mimetype !== "application/zip"
   ) {
-    reply.status(400).send({ error: `Invalid mimetype of uploaded file, received ${file.mimetype}` });
+    reply.status(400).send({ error: `Invalid mimetype of uploaded file, received ${uploadedFile.mimetype}` });
     return;
```
paologum, line 306: on all these errors, you need to clean up the file

server/src/routes.ts
```diff
@@ -264,30 +268,67 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
   }
   // for test usage: const uuid = "username";
-  const file = await request.file();
-  if (!file) {
+  const parts = request.parts();
+  const fields: Record<string, string> = {};
+  let uploadedFile: MultipartFile | undefined;
+  const userDir = path.join(assetconfigs.uploadDirectory, uuid);
+  const datapackDir = path.join(userDir, "datapacks");
+  await mkdirp(datapackDir);
```
paologum, line 276: try catch, also just use `mkdir`

server/src/routes.ts
```diff
@@ -264,30 +268,67 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  for await (const part of parts) {
+    if (part.type === "file") {
+      uploadedFile = part;
+      const filepath = path.join(datapackDir, uploadedFile.filename);
+      const fileStream = fs.createWriteStream(filepath);
+      await pump(uploadedFile.file, fileStream);
```
paologum, line 282: try catch, and cleanup

server/src/routes.ts
```diff
@@ -264,30 +268,67 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
   const datapackIndexFilepath = path.join(userDir, "DatapackIndex.json");
+  let formattedSize;
+  const bytes = uploadedFile.file.bytesRead;
+  if (bytes === 0) formattedSize = "0 Bytes";
+  else {
+    const k = 1024;
+    const sizes = ["Bytes", "KB", "MB", "GB", "TB"];
+    const i = Math.floor(Math.log(bytes) / Math.log(k));
+    formattedSize = parseFloat((bytes / Math.pow(k, i)).toFixed(2)) + " " + sizes[i];
+  }
```
paologum, line 324: if bytes is 0, it should not be a valid file so throw an error

server/src/routes.ts
paologum: just a heads up but i won't approve until all the minor fixes for local deving are committed and pushed

server/src/routes.ts
paologum: also, the `uploadedFile = part` i was talking about earlier only ever works for unencrypted files for some reason, i haven't looked into it more but this is fine now

paologum (changes requested):

app/src/components/datapack_display/TSCDatapackRow.tsx
```diff
@@ -60,13 +61,13 @@ export const TSCDatapackRow: React.FC<TSCDatapackRowProps> = ({ name, datapack,
       <img className={styles.image} src={imageUrl} alt="datapack" onError={() => setImageUrl(defaultImageUrl)} />
       <div className={styles.middle}>
         <Typography className={styles.header} color="textSecondary">
-          {name}
+          {datapack.title}
         </Typography>
         <Typography className={styles.fd} color="textSecondary">
-          Dixon, Dougal, et al. · Created 10/10/2024
+          {datapack.description}
```
paologum, line 67: this is for the author and creation date, so it shouldn't be the description, you can leave this for now

server/src/routes.ts
```diff
@@ -274,58 +278,97 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  const parts = request.parts();
+  const fields: Record<string, string> = {};
+  let uploadedFile: MultipartFile | undefined;
+  let userDir: string;
+  let datapackDir: string;
+  let filepath: string = "";
+  try {
+    userDir = path.join(assetconfigs.uploadDirectory, uuid);
+    datapackDir = path.join(userDir, "datapacks");
+    await mkdir(datapackDir);
```
paologum, line 297: you should make it recursively just to be sure

server/src/routes.ts
```diff
@@ -274,58 +278,97 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+        filepath = path.join(datapackDir, uploadedFile.filename);
+        const fileStream = fs.createWriteStream(filepath);
+        try {
+          await pump(uploadedFile.file, fileStream);
+        } catch (e) {
+          await errorHandler("Failed to upload file with error " + e, 500, e);
+        }
+      } else if (part.type === "field") {
+        const field = part as { fieldname: string; value: string };
+        fields[part.fieldname] = field.value;
```
paologum, line 315: to be robust you should check the type of `fieldname` and `value` in the if as well

server/src/routes.ts
```diff
@@ -274,58 +278,97 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  }
+  const name = fields.name;
+  const description = fields.description;
+
+  if (!uploadedFile) {
+    await errorHandler("No file uploaded", 404);
+    return;
+  }
+  if (!name || !description) {
+    await errorHandler("Name or description not provided", 404);
```
paologum, line 330: should bee 400

server/src/routes.ts
```diff
@@ -274,58 +278,97 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+    for await (const part of parts) {
+      if (part.type === "file") {
+        uploadedFile = part;
+        filepath = path.join(datapackDir, uploadedFile.filename);
+        const fileStream = fs.createWriteStream(filepath);
+        try {
+          await pump(uploadedFile.file, fileStream);
+        } catch (e) {
+          await errorHandler("Failed to upload file with error " + e, 500, e);
+        }
```
paologum, line 312: return if error

server/src/routes.ts
```diff
@@ -274,58 +278,97 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+      } else if (part.type === "field") {
```
paologum, line 313: check truncation here and not on line 371

server/src/routes.ts
```diff
@@ -274,58 +278,97 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
-    });
-  } catch (e) {
-    await errorHandler("Failed to save file with error: " + e, 500, e);
-    return;
-  }
-  if (file.file.truncated) {
+  if (uploadedFile.file.truncated) {
     await errorHandler("File too large", 413);
     return;
   }
```
paologum, line 374: for above

paologum (changes requested):

server/src/routes.ts
```diff
@@ -274,61 +278,102 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  }
+  const name = fields.name;
+  const description = fields.description;
+
+  if (!uploadedFile) {
+    await errorHandler("No file uploaded", 400);
+    return;
+  }
+  if (!name || !description) {
+    await errorHandler("Name or description not provided", 404);
```
paologum, line 335: this should be 400

rupp3:
I showed a demo of my branch in the meeting after you left. Prof Ogg and Aaron suggested having a way to differentiate between the server datapacks and user uploaded datapacks on the front end. Also said to remove the Upload Datapack button if a user is not logged in. I can do that in this branch but I would highkey love to merge this PR first and do those fixes on a different branch

paologum (approved):
looks great! i'll add those in a seperate task so don't worry about that for this pr. you can continue working on the public datapack stuff as well as a priority too and we'll figure out when to do the other items

PR #90, opened 2024-04-24
Tsc 295 general chart fonts change chart root

paologum:
this is the fonts tab in the java file that allows you to change the font options for the whole chart (every possible unit/chart inside chart root)

basically used `FontRows` that's used for individual columns, but locked it if it was `Chart Root`

i've made it so it only displays the font menu rows that apply to the whole chart (not default all)

Screenshot 2024-04-24 at 2 28 15 AM
Screenshot 2024-04-24 at 2 30 55 AM


asivath (approved):
looks good

PR #93, opened 2024-05-04
strip quotations for id, check if settings is populated before applying

jay-jhk-lee:
Some datapacks generated column names with quotations surrounding it which is not how the setting file id is formed, and changed to correctly check settings before applying


app/src/state/parse-settings.ts
```diff
@@ -342,6 +342,10 @@ export function translateColumnInfoToColumnInfoTSC(state: ColumnInfo): ColumnInf
     case "Point":
       column = JSON.parse(JSON.stringify(defaultPointColumnInfoTSC));
   }
+  //strip surrounding quotations for id (ex. Belgium Datapack)
+  if (state.name.charAt(0) === '"' && state.name.charAt(state.name.length - 1) === '"') {
+    state.name = state.name.slice(1, -1);
+  }
```
paologum, line 348: use the trim quotes method in `util.ts`

paologum: also add a todo here to check with Ogg about quote usage, just so i can find this easier in the future if need be

paologum (changes requested):

server/src/parse-datapacks.ts
```diff
@@ -224,14 +224,14 @@ export async function parseDatapacks(file: string, decryptFilePath: string): Pro
     },
     popup: "",
     children: [],
-    parent: chartTitle,
+    parent: chartTitle === "TimeScale Creator GTS2020 chart" ? "Chart Title" : chartTitle,
```
paologum, line 227: can you explain?

paologum: ok i understand now, instead change the name to "Chart Title" and keep the edit name `chartTitle`

should just change for the parent param in `recursive`, 234, and the ruler column above

app/src/state/actions/general-actions.ts
```diff
@@ -330,7 +330,8 @@ export const setDatapackConfig = action(
     }
     resetSettings();
     //TODO: apply presets, temp code for applying just the chart settings
-    if (chartSettings) {
+    //check if chart settings is populated
+    if (Object.keys(chartSettings).length !== 0) {
```
paologum, line 334: i think this would make more sense if you just made `chartSettings` `ChartSettingsTSC | null`, that way you can check if it's truthy and it makes more sense that it doesn't exist rather than a partial

paologum (changes requested):

paologum (approved):

PR #97, opened 2024-05-08
search algorithm for columns

paologum:
picked this up and finished resolving all errors from sejal :D

one error/bug that will be fixed in a future task are directories/parents that are "found" or set to show true with none of their children matching. this will be fixed when we implement a more robust expand variable instead of how we implement expansions currently.

https://github.com/earthhistoryviz/tsconline/assets/89664317/13edf244-0e11-4fa7-b060-d12a0e8cf2c5


app/src/state/actions/column-actions.ts
```diff
@@ -66,13 +66,41 @@ export const updateWidth = action((columnObject: ColumnInfo, newWidth: number) =
...
+        if (parentColumnInfo) {
+          parentColumnInfo.show = true;
+          parentName = parentColumnInfo.parent;
+        } else {
+          break;
+        }
+      }
+    }
+  });
+});
```
asivath, line 102: can be one loop

asivath (commented):
just one thing

app/src/state/actions/column-actions.ts
paologum: if you're referenceing 83 and 87 loops merging, technically I don't think they can because then parents might be set to false when they might include children that contain the search term

asivath (approved):

PR #98, opened 2024-05-09
Tsc 302 allow users to change their email

asivath:
https://www.dropbox.com/scl/fi/d1byydyzewqqyw0rzh25e/reset_email_example.mp4?rlkey=hnv8q3zwcnweqgeyfd9hip9gz&st=gumed8h0&dl=0
This video shows the flow for a user who didn't request their email to be changed. If they meant to then it will follow the normal email verification flow which I didn't show. 

I needed to modify the database to implement this feature so I introduced migrations, which I explain with comments in database.ts and migrations.ts.


server/db/migrate.ts
paologum, line 1: so fun so fun, i'm gonna probably migrate `file metadata` to be in the database because it's gonna have migrating issues as well. that should be fine right

server/db/migrations/2024.05.08-17.59.45-adjust-columns-in-verification-table.ts
```diff
@@ -0,0 +1,35 @@
...
+  await db.schema
+    .createTable("verification_new")
+    .addColumn("id", "integer", (col) => col.autoIncrement().primaryKey())
+    .addColumn("userId", "integer", (col) => col.notNull())
+    .addColumn("token", "text", (col) => col.notNull().unique())
+    .addColumn("expiresAt", "datetime", (col) => col.notNull())
+    .addColumn("verifyOrReset", "text", (col) => col.notNull())
+    .execute();
+  await sql<Verification[]>`INSERT INTO verification_new (userId, token, expiresAt, verifyOrReset)
+                            SELECT userId, token, expiresAt, verifyOrReset FROM verification`.execute(db);
```
paologum, line 15: is this not a Kysely command we can execute

server/src/database.ts
```diff
@@ -1,46 +1,66 @@
...
+  emailVerified: integer, not null, default 0
+  invalidateSession: integer, not null, default 0
+
+verification:
+  id: integer, primary key, auto increment
+  userId: integer, not null
+  token: text, not null, unique
+  expiresAt: datetime, not null
+  reason: text, not null
+*/
```
paologum, line 32: so this schema will never change ever

server/src/database.ts
```diff
@@ -1,46 +1,66 @@
...
+
+exec("cd db && yarn tsx migrate.ts up", (error, stdout, stderr) => {
+  if (error) {
+    console.error(`exec error: ${error}`);
+    return;
+  }
+  stdout && console.log(stdout);
+  stderr && console.error(stderr);
+});
+
```
paologum, line 63: should do this in a promise

server/src/login-routes.ts
```diff
@@ -42,6 +42,150 @@ const sendEmail = async (email: Email) => {
...
+    if (!user) {
+      throw new Error("User not found");
+    }
+    const randomPassword = randomBytes(16).toString("hex");
+    const newEmail: Email = {
+      from: process.env.EMAIL_USER,
+      to: email,
+      subject: "Account Recovery",
+      text: `Your account credentials have been invalidated. To regain access, please reset your password by visiting ${process.env.APP_URL || "http://localhost:5173"}/forgot-password. If you did not request this change or need further assistance, please contact our support team.`
+    };
```
paologum, line 94: just a comment, but your next task should probably be making a general email sender that is just a styled email given some params since we have so many emails at this point

server/src/login-routes.ts
```diff
@@ -42,6 +42,150 @@ const sendEmail = async (email: Email) => {
...
+  try {
+    const verificationRow = (await findVerification({ token }))[0];
+    if (!verificationRow || verificationRow.reason !== "invalidate") {
+      reply.status(404).send({ error: "Token not found" });
+      return;
+    }
+    if (new Date(verificationRow.expiresAt) < new Date()) {
+      await deleteVerification({ token, reason: "invalidate" });
+      reply.status(401).send({ error: "Token expired" });
+      return;
```
paologum, line 81: is there ever a worry that so many people have expired tokens and never reaccess it over long periods of time and there's just a buildup of expired tokens in the db. should we clean that up in that scenario?

server/src/login-routes.ts
```diff
@@ -42,6 +42,150 @@ const sendEmail = async (email: Email) => {
...
+  }
+  reply.send({ authenticated: true });
+};
+
+export const accountRecovery = async function accountRecovery(
+  request: FastifyRequest<{ Body: { token: string; email: string } }>,
+  reply: FastifyReply
+) {
+  const { token, email } = request.body;
+  if (!token || !email) {
```
paologum, line 64: check regex for email

server/src/login-routes.ts
```diff
@@ -42,6 +42,150 @@ const sendEmail = async (email: Email) => {
...
+        subject: "Email Changed",
+        text: "You have requested to change your email to the same email. If you did not request this change, please change your password and contact support."
+      };
+      sendEmail(sameEmail);
+      reply.send({ message: "Email changed" });
+    }
+    if (hashedPassword && !(await compare(password, hashedPassword))) {
+      reply.status(401).send({ error: "Incorrect password" });
+      return;
+    }
```
paologum, line 145: let's say that a google user changes their email to a non-google account. then how do they sign in with no password?

additionally if they switch over to another google account that should be fine then right

server/src/types.ts
```diff
@@ -14,13 +14,14 @@ export interface UserTable {
   uuid: string;
   pictureUrl: string | null;
   emailVerified: number;
+  invalidateSession: number;
```
paologum, line 17: is there no way to make these booleans for better readability?

paologum (changes requested):

app/src/ResetEmail.tsx
```diff
@@ -0,0 +1,116 @@
...
+import "./Login.css";
+
+export const ResetEmail: React.FC = observer(() => {
+  const theme = useTheme();
+  const [loading, setLoading] = useState(false);
+  const [showForm, setShowForm] = useState(true);
+  const { actions } = useContext(context);
+  const navigate = useNavigate();
+
+  const handlelSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
```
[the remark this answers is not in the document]
asivath: crazy i should probably look into a spell checker

app/src/AccountRecovery.tsx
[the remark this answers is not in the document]
asivath: this where users go to if they click the link for invalidating their password

app/src/ResetEmail.tsx
```diff
@@ -0,0 +1,116 @@
...
+        switch (response.status) {
+          case 400:
+            errorCode = ErrorCodes.INVALID_FORM;
+            break;
+          case 401:
+            errorCode = ErrorCodes.NOT_LOGGED_IN;
+            navigate("/login");
+            break;
+          case 409:
+            errorCode = ErrorCodes.EMAIL_ALREADY_EXISTS;
```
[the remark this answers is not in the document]
asivath: yea it used to be a thing but I removed it, forgot to remove this

server/db/migrate.ts
asivath: yes should be fine

server/db/migrations/2024.05.08-17.59.45-adjust-columns-in-verification-table.ts
asivath: yeah i found a way to do it

server/src/database.ts
asivath: yeah the initial table creation commands and any migrations pushed to main should never be changed. that's why I thought we should document the actual schema there after all migrations

server/src/login-routes.ts
asivath: yeah I can clean them up

server/src/login-routes.ts
asivath: technically this email is not entered by the user it comes from the db which has already gone through the check but i'll add it anyway since a user could alter the email in the address or copy it wrong

server/src/login-routes.ts
asivath: we already don't allow google accounts and user/password accounts to mix so should I allow google users to do that? if we want to I could get rid of the password in the form (which is kind of redundant since they have to be signed in to access it) and assign google users a random password they can use to sign in and they can also change their password

app/src/ResetPassword.tsx
```diff
@@ -0,0 +1,181 @@
...
+        <Avatar sx={{ "& .MuiSvgIcon-root": { mr: 0 }, bgcolor: theme.palette.navbar.dark }}>
+          <LockOutlinedIcon sx={{ color: theme.palette.selection.main }} />
+        </Avatar>
+        {loading && <Lottie animationData={loader} autoplay loop width={200} height={200} speed={0.7} />}
+        <Typography variant="h5" sx={{ textAlign: "center", mt: 1 }}>
+          {showPasswordForm || showResendForm
+            ? showResendForm
+              ? "Enter your email to receive a password reset link."
+              : "Please enter your new password."
+            : "Password Recovery"}
```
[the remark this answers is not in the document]
paologum: nvm i lied loader goes above

server/src/types.ts
asivath: i've looked into ways other people have done [it](https://github.com/kysely-org/kysely/issues/123) but since it wasn't super easy I didn't bother. lmk if you think I should use one of those methods

app/src/ResetPassword.tsx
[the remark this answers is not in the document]
asivath: the stall happens from initialize, when I comment out all its actions it loads fine

server/src/login-routes.ts
asivath: i've allowed google users to swtich to user/pass but I have no way of knowing if a users new email is a google account so I'm not sure how to handle that. this could be solved by allowing user/pass and google accounts to mix, so wdyt?

paologum: i think we just don't allow them to go to google accounts after switching to user/pass so what you have sounds good

server/src/types.ts
paologum: ok yeah we can just leave it as it is. also i watched a tiktok that said that we could store enums (like the`reason`) into binary where each bit represents a reason. ie `00` is `password`, `10` is `invalidate, and `01` is `verify`. then we have like a dict mapping for the binary numbers. this makes queries run faster since they aren't checking strings if you query with binary operations like & or |. 

i think this is interesting although something that probably doesn't matter all to much for us since our scale is not on any crazy level and the amount of enums we have for `reason` are so little. thought it was fun though

server/src/login-routes.ts
```diff
@@ -138,32 +138,58 @@ export const resetEmail = async function resetEmail(
...
+        expiresAt: expiresAt.toISOString(),
+        reason: "verify"
+      };
+      await createVerification(newVerification);
+    } else {
+      await updateUser({ userId }, { email: newEmail, emailVerified: 0 });
+      const token = randomBytes(16).toString("hex") + md5(uuid);
+      const expiresAt = new Date();
+      expiresAt.setHours(expiresAt.getHours() + 1);
+      await deleteVerification({ userId: userId, reason: "verify" });
```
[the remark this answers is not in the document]
asivath: true there is no real reason I can think of to keep them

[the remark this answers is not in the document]
asivath: actually except for invalidate tokens since if a attacker send multiple reset emails they could delete the actual users invalidate token

paologum (approved):

PR #99, opened 2024-05-09
backend for curve settings on Point Columns

paologum:
the backend for data mining curves for points, no display yet will be worked on in TSC-320

added util function for assigning common properties between two objects. (different from `Object.assign`).

created tests for assigning common properties and a `util` function i had forgotten about


app/src/state/actions/general-actions.ts
```diff
@@ -188,6 +188,9 @@ export const fetchUserDatapacks = action("fetchUserDatapacks", async () => {
     } catch (e) {
       if (response.status != 404) {
         displayServerError(data, ErrorCodes.INVALID_USER_DATAPACKS, ErrorMessages[ErrorCodes.INVALID_USER_DATAPACKS]);
+      } else {
+        fetchDatapackIndex();
+        fetchMapPackIndex();
```
paologum, line 193: fix for if you signed in but have no datapacks

server/src/util.ts
```diff
@@ -155,3 +155,19 @@ export async function resetUploadDirectory(uploadedFile: string, decryptedUpload
...
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function setCommonProperties(o1: any, o2: any) {
+  for (const key in o1) {
+    if (Object.hasOwnProperty.call(o2, key)) {
+      o1[key] = o2[key];
+    }
+  }
+  return o1;
+}
```
asivath, line 173: could you use generics and partial here instead of any?

shared/src/index.ts
```diff
@@ -201,7 +201,30 @@ export type SubInfo =
   | SubSequenceInfo
   | SubTransectInfo;
 
-export type ColumnSpecificSettings = EventSettings;
+export type ColumnSpecificSettings = EventSettings | PointSettings;
+
+export type PointSettings = {
```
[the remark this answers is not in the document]
paologum: `drawPoints` is false if `pointShape` is `nopoints` so there's no reason to hold both variables

server/src/util.ts
paologum: i'll try it out

asivath (approved):

jay-jhk-lee (approved):

PR #100, opened 2024-05-10
Tsc 310 apply column settings from settings file

jay-jhk-lee:
added the backend for loading settings files. Created a function to recursively go down the info given by the settings file and accessing the respective column info state. Standardized the parameters of the setter functions for columnInfo. Added formatting to the column names when parsing datapack to align with tscreator and the settings file.


paologum (changes requested):

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,48 @@
...
+    column.fontsInfo = cloneDeep(settings.fonts);
+  }
+  const columnName = extractName(settings._id);
+  let curcol: ColumnInfo | undefined = state.settingsTabs.columnHashMap.get(columnName);
+  let blockSeriesColumns: boolean = false;
+  if (curcol === undefined) {
+    if (extractColumnType(settings._id) === "RootColumn") {
+      curcol = state.settingsTabs.columnHashMap.get("Chart Title in " + columnName);
+    }
+  }
```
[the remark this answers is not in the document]
jay-jhk-lee: it only does it if curcol is undefined from the initial map access, so `Chart Root` and `Chart Title` are already covered.

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,48 @@
...
+  return text.substring(text.indexOf(".") + 1, text.indexOf(":"));
+}
+export const applyChartColumnSettings = action("applyChartColumnSettings", (settings: ColumnInfoTSC) => {
+  function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
+    updateEditName(settings.title, column);
+    setEnableTitle(settings.drawTitle, column);
+    if (settings.width) updateWidth(settings.width, column);
+    if (settings.backgroundColor.text) setRGB(settings.backgroundColor.text, column);
+    column.fontsInfo = cloneDeep(settings.fonts);
+  }
```
paologum: one more thing, is there a reason that this is a nested function?

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,48 @@
...
+  let curcol: ColumnInfo | undefined = state.settingsTabs.columnHashMap.get(columnName);
+  let blockSeriesColumns: boolean = false;
+  if (curcol === undefined) {
+    if (extractColumnType(settings._id) === "RootColumn") {
+      curcol = state.settingsTabs.columnHashMap.get("Chart Title in " + columnName);
+    }
+  }
+  if (curcol === undefined) {
+    pushSnackbar("Unknown column name '" + columnName + "' found while loading settings", "warning");
+    console.log("Unknown column name " + columnName + " found while loading settings");
```
paologum, line 46: this would make more sense like
```
if (curcol === undefined) {
    if (extractColumnType(settings._id) === "RootColumn") {
      curcol = state.settingsTabs.columnHashMap.get("Chart Title in " + columnName);
    } else {
      pushSnackbar("Unknown column name '" + columnName + "' found while loading settings", "warning");
    console.log("Unknown column name " + columnName + " found while loading settings");
    }
  } else  setColumnProperties(curcol, settings);
```

easier to read and less dupe

paologum: actually wait let me think for a second

paologum:
```
let curcol: ColumnInfo | undefined = state.settingsTabs.columnHashMap.get(columnName) || state.settingsTabs.columnHashMap.get("Chart Title in " + columnName");
if (!curcol) {
  pushSnackbar stuff
} else setColumnProps
if (BlockSeriesMetaColumn) {
  for loop with diff name
} else {
  normal for loop
}

```

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,64 @@
...
+    if (settings.width) setWidth(settings.width, column);
+    if (settings.backgroundColor.text) setRGB(settings.backgroundColor.text, column);
+    column.fontsInfo = cloneDeep(settings.fonts);
+    switch (extractColumnType(settings._id)) {
+      case "EventColumn":
+        assertEventColumnInfoTSC(settings);
+        if (column.columnSpecificSettings)
+          setEventColumnSettings(column.columnSpecificSettings, { type: settings.type, rangeSort: settings.rangeSort });
+    }
+  }
```
paologum, line 35: can't this be it's own function outs of this (same comment from before but on your most recent commit)

app/src/state/parse-settings.ts
```diff
@@ -122,6 +122,49 @@ function processSettings(settingsNode: Element): ChartSettingsInfoTSC {
...
+    if (maybeChild.nodeType === Node.ELEMENT_NODE) {
+      const child = <Element>maybeChild;
+      let fontProps: string[] = [];
+      if (child.textContent) {
+        fontProps = child.textContent.trim().split(";");
+      }
+      for (let i = 0; i < fontProps.length; i++) {
+        let key = "" as keyof FontsInfo;
+        if (child.getAttribute("function")) {
+          key = child.getAttribute("function")! as keyof FontsInfo;
```
paologum, line 137: you should also check to make sure the key exists on `FontsInfo` as well

app/src/state/parse-settings.ts
```diff
@@ -122,6 +122,49 @@ function processSettings(settingsNode: Element): ChartSettingsInfoTSC {
...
+        } else continue;
+        fonts[key].inheritable = Boolean(child.getAttribute("inheritable"));
```
paologum, line 139: check `child.getAttr` here too, but default this to false if it doesn't find one

app/src/state/parse-settings.ts
```diff
@@ -122,6 +122,49 @@ function processSettings(settingsNode: Element): ChartSettingsInfoTSC {
...
+
+        if (fontProps[i].includes("font-family")) {
+          fonts[child.getAttribute("function")! as keyof FontsInfo].fontFace =
```
paologum, line 142: change to key

app/src/state/parse-settings.ts
```diff
@@ -122,6 +122,49 @@ function processSettings(settingsNode: Element): ChartSettingsInfoTSC {
...
+        }
+        if (fontProps[i].includes("font-weight")) {
+          if (fontProps[i].includes("bold")) {
+            fonts[key].bold = true;
+          }
+        }
+        if (fontProps[i].includes("fill")) {
+          fonts[key].color = fontProps[i].split(": ")[1];
+        }
+      }
```
paologum, line 165: it would be beneficial for us to split `fontProps[i]` early by ": " and call it like `fontPropKey` then check includes on that for less comparisons. also the nested ifs can just be one `if` statement with `&&`

server/src/parse-datapacks.ts
```diff
@@ -1293,6 +1299,21 @@ function recursive(
...
+      if (!Number.isNaN(Number(text.charAt(commaPos - 1)))) {
+        text = text.substring(0, commaPos) + "." + text.substring(commaPos + 1, text.length);
+      }
+      commaPos = text.indexOf(",", commaPos + 1);
+    }
+    return text;
+  }
+  currentColumn = formatColumnName(currentColumn);
   const currentColumnInfo: ColumnInfo = {
     name: trimInvisibleCharacters(currentColumn),
```
paologum, line 1318: take this out so we don't have unecessary nested functions, also format the column name after the trim happens `name: formatColumnName(trimInvisibleCharacters(currentColumn)`

app/src/state/parse-settings.ts
```diff
@@ -122,6 +122,49 @@ function processSettings(settingsNode: Element): ChartSettingsInfoTSC {
...
+        if (child.getAttribute("function")) {
+          key = child.getAttribute("function")! as keyof FontsInfo;
+        } else continue;
+        fonts[key].inheritable = Boolean(child.getAttribute("inheritable"));
+
+        if (fontProps[i].includes("font-family")) {
+          fonts[child.getAttribute("function")! as keyof FontsInfo].fontFace =
+            fontProps[i].split(": ")[1] === "Arial" || "Courier" || "Verdana"
+              ? <"Arial" | "Courier" | "Verdana">fontProps[i].split(": ")[1]
+              : "Arial";
```
paologum, line 145: idk if you wanna do this in this pr but we should change this to a regex in shared like `EventType` just in case we wanna do more fonts

app/src/state/actions/column-actions.ts
jay-jhk-lee: yes, but I think that this function is super specific to this use case so for scope I wanted to leave it inside. I'm good with either tho

paologum: it's more so for the fact of readability than scope

paologum: honestly this could benefit into being turned into an action as well, but that's up to you

server/src/parse-datapacks.ts
jay-jhk-lee: for the first part, same reasoning as before

paologum: this could be a utility function that's generalized as sanitizing a string and also i looked into simpler ways of executing this command and you can just do

```
 return text
    .trim()
    .trimQuotes()
    .replace(/^"(.*)"$/, '$1')
    .replace(/""/g, '"')
    .replace(/(\d),(\d)/g, '$1.$2');
```

for easier readability

PR #101, opened 2024-05-10
Tsc 275 create encryption UI and route

JacquiFreeElectron:
Users can encrypt the datapacks they upload. They can download the encrypted version or simply retrieve the original file. The download button will only show for user-uploaded datapacks.


server/src/routes.ts
```diff
@@ -59,6 +59,104 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+  const { filename } = request.params;
+  const userDir = path.join(assetconfigs.uploadDirectory, uuid);
+  const datapackDir = path.join(userDir, "datapacks");
+  const filepath = path.join(datapackDir, filename);
+  const encryptedFilepathDir = path.join(datapackDir, "encrypted");
+  const checkIfAlreadyEncrypted = path.join(encryptedFilepathDir.replaceAll("\\", "/"), filename);
+  if (fs.existsSync(checkIfAlreadyEncrypted)) {
+    const file = fs.readFileSync(checkIfAlreadyEncrypted);
+    reply.send(file);
+    console.log("file is encrpted when uploaded");
```
jay-jhk-lee, line 82: small spelling error

server/src/routes.ts
```diff
@@ -59,6 +59,104 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+  if (needEncryption === "true") {
+    try {
+      await access(downloadPath);
+      const file = await readFile(downloadPath);
+      const fileContent = file.toString();
+      const lines = fileContent.split("\n");
+
+      if (lines[0] && lines[0].includes("TSCreator Encrypted Datafile")) {
+        reply.send(file);
+        console.log("file is already encrpted before");
```
jay-jhk-lee, line 96: same

server/src/routes.ts
```diff
@@ -59,6 +59,104 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+    return;
+  }
+  //For test usage: const uuid = "username";
+  const { needEncryption } = request.params;
+  const { filename } = request.params;
+  const userDir = path.join(assetconfigs.uploadDirectory, uuid);
+  const datapackDir = path.join(userDir, "datapacks");
+  const filepath = path.join(datapackDir, filename);
+  const encryptedFilepathDir = path.join(datapackDir, "encrypted");
+  const checkIfAlreadyEncrypted = path.join(encryptedFilepathDir.replaceAll("\\", "/"), filename);
```
jay-jhk-lee, line 78: this sounds like a boolean, so I think a better name could be `MaybeEncryptedFilepath`

app/src/SettingsTabs/Datapack.tsx
```diff
@@ -26,6 +31,67 @@ export const Datapacks = observer(function Datapacks() {
...
+      )
+    );
+  };
+  async function getFileURL(needEncryption: boolean, fileName: string) {
+    let fileBlob;
+    if (needEncryption) {
+      fileBlob = await actions.requestDownload(true, fileName);
+    } else {
+      fileBlob = await actions.requestDownload(false, fileName);
+    }
```
paologum, line 57: this can just be `fileBlob = await actions.requestDownload(needEncryption, fileName);

app/src/SettingsTabs/Datapack.tsx
```diff
@@ -26,6 +31,67 @@ export const Datapacks = observer(function Datapacks() {
...
+        <div>
+          <Menu menuButton={<DownloadIcon className="download-icon" />} transition>
+            <MenuItem onClick={() => handleDownload(true, name)}>
+              <Typography>encrypted download</Typography>
+            </MenuItem>
+            <MenuItem onClick={() => handleDownload(false, name)}>
+              <Typography>retrieve original file</Typography>
+            </MenuItem>
+          </Menu>
+        </div>
```
paologum, line 47: Use react fragment, also use pascal case for the options

server/src/routes.ts
```diff
@@ -59,6 +59,104 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+      reply.status(500).send({ error: "An error occurred" });
+    }
+  }
+};
+
+export const loadActiveDatapacks = async function loadActiveDatapacks(request: FastifyRequest, reply: FastifyReply) {
+  const activeDatapacks = assetconfigs.activeDatapacks;
+  reply.status(200).send({ activeDatapacks });
+  return;
+};
```
paologum, line 158: we shouldn't have a route for getting server datapacks only, that's already a route. the way that this should be done is not with a rest api call. you should use a boolean on the `DatapackParsingPack` that is called `isUserDatapack`

app/src/state/actions/general-actions.ts
```diff
@@ -704,6 +704,39 @@ export const fetchImage = action("fetchImage", async (datapackName: string, imag
...
+    method: "GET"
+  });
+  if (!response.ok) {
+    if (response.status === 404) {
+      throw new Error("File not found");
+    } else if (response.status === 500) {
+      throw new Error("Server error");
+    } else {
+      throw new Error("Unknown error");
+    }
```
paologum, line 734: display errors, you shouldn't throw errors on the app side

app/src/SettingsTabs/Datapack.tsx
```diff
@@ -26,6 +31,67 @@ export const Datapacks = observer(function Datapacks() {
...
+
+  async function checkActiveDatapacks(name: string, key: number) {
+    const activeDatapacks = await actions.loadActiveDatapacks();
+    const isSystemDatapack = activeDatapacks.includes(name);
+    if (!isSystemDatapack) {
+      const newButtonState = [...downloadButton];
+      newButtonState[key] = true;
+      setDownloadButton(newButtonState);
+    }
+  }
```
paologum, line 94: no need for a method, it should just be a field on the parsing pack

server/src/index.ts
```diff
@@ -137,7 +137,14 @@ server.get("/presets", async (_request, reply) => {
   reply.send(presets);
 });
 // uploads datapack
-server.post("/upload", routes.uploadDatapack);
+server.post<{ Params: { username: string } }>("/upload", routes.uploadDatapack);
+//download datapack
+server.get<{ Params: { needEncryption: string; filename: string } }>(
+  "/requestDownload/:needEncryption/:filename",
+  routes.requestDownload
```
paologum, line 144: routes should be nouns and not predicates so this route should be `/downloads/:filename` with a query of `encrypted`
filename is still a param but encrypted should be a query
we use query when it's an optional "param"

then you can default to the original file they uploaded and anyone accessing this doesn't need the overhead of knowing whether they want to encrypt or not

server/src/routes.ts
```diff
@@ -59,6 +59,104 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+    } catch (e) {
+      const error = e as NodeJS.ErrnoException;
+      if (error.code === "ENOENT") {
+        reply.status(404).send({ error: "File not found" });
+      } else {
+        reply.status(500).send({ error: "An error occurred" });
+      }
+      return;
+    }
+    await mkdirp(encryptedFilepathDir);
```
paologum, line 108: this can throw an error, should check

server/db/TSC.db
paologum, line 1: should not be committed

server/src/routes.ts
paologum, line 82: use `await access`  (async for everything here, no sync)

server/src/routes.ts
paologum: you should just remove this `console.log`

server/src/routes.ts
paologum: no `console.log`

paologum (changes requested):

server/src/routes.ts
paologum: also, if this directory exists, before, that must mean the file has been encrypted before, meaning that you should check that file and send it. if that file isn't encrypted, delete it and remake it

server/src/routes.ts
```diff
@@ -59,6 +59,104 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+      console.error(e);
+      reply.status(500).send({ error: "Failed to encrypt datapacks with error " + e });
+      return;
+    }
+    downloadPath = path.join(encryptedFilepathDir.replaceAll("\\", "/"), filename);
+  }
+  try {
+    await access(downloadPath);
+    const file = await readFile(downloadPath);
+    reply.send(file);
```
paologum, line 143: should verify it encrypted properly

server/src/routes.ts
JacquiFreeElectron: I think line 79 to 82 checked if a file has been encrypted before. 
Also, if I verify each encryption worked properly by checking the header in line 141 to 143, do I need to check again in here?

server/src/routes.ts
JacquiFreeElectron: Do I need to change `if (fs.existsSync(checkIfAlreadyEncrypted))` too? Cause I don't know how to use access to check if a file exist

server/src/routes.ts
paologum: this is the case where the user has uploaded an unencrypted file, and has requested an encrypted file a second time, meaning it already exists as an encrypted file and we shouldn't encrypt it again

server/src/routes.ts
paologum: `await access` checks if a file exists

server/src/routes.ts
JacquiFreeElectron: yes line 79 to 82 does that. The java program will not be called again

paologum: ah ok but it should verify it is encrypted again on those lines

server/src/routes.ts
paologum: also is there a reason that `encryptedFilepathDir` needs to be replaced? `path` should handle the system path separator logic and we shouldn't replace them unless we are using it to execute a command. it technichally won't do anything either because `path.join` normalizes those back if it was replaced if the system uses back slashes

server/src/routes.ts
JacquiFreeElectron: But `await access` doesn't have a return value. I need a boolean value to decide if I need to return. Or is there a better way?

server/src/routes.ts
JacquiFreeElectron: ok

server/src/routes.ts
paologum: you have an implementation of `await access` on line 89 already?

server/src/routes.ts
JacquiFreeElectron: ok I see. Thanks for telling me this

server/src/routes.ts
JacquiFreeElectron: ok I see what you mean

server/src/routes.ts
```diff
@@ -59,6 +59,104 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+  const checkIfAlreadyEncrypted = path.join(encryptedFilepathDir.replaceAll("\\", "/"), filename);
+  if (fs.existsSync(checkIfAlreadyEncrypted)) {
+    const file = fs.readFileSync(checkIfAlreadyEncrypted);
+    reply.send(file);
+    console.log("file is encrpted when uploaded");
+    return;
+  }
+  let downloadPath = filepath.replaceAll("\\", "/");
+
+  if (needEncryption === "true") {
```
paologum, line 87: one more thing, you need a case for when it's not encrypted, in that case just send the file and don't call java program

JacquiFreeElectron: Yes I have that

paologum: from what i see, it will always run the java program if the user has uploaded an encrypted file

JacquiFreeElectron: No it won't. bc it first check if there is an encrypted file that exist in `encrypted-datapacks` folder, if there exist one, it will return that file without calling the java program. if not, by default the download path is `uuid/datapacks/file.txt`. If `needEncryption` is false, then the download path will never be changed and it will directly download the original file, also the command will not be executed since needEncryption is false. So if you upload an encrypted file the java program will never be called. I just tested it again and it is the case.

JacquiFreeElectron: And if you upload an ecrypted file and request encrypted download, it will also first check the header of the file and if it is encrypted it will return and the java program also won't be called in that case.

paologum: the problem is that you only check `encrypted-datapacks`, the original file could just be encrypted and you don't have any checks for that, so it will run the java program

paologum: hmm wait i see, but technically if you do it this way for both an encrypted file in the encrypted directory, or the uploaded directory you will potentially delete their uploaded directory datapack

we don't want that because that's their file, and the file in the `uploaded` is guaranteed to be valid

server/src/routes.ts
```diff
@@ -59,6 +60,137 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+  if (needEncryption) {
+    try {
+      await access(downloadPath);
+      const file = await readFile(downloadPath);
+      const fileContent = file.toString();
+      const lines = fileContent.split("\n");
+
+      if (lines[0] && lines[0].includes("TSCreator Encrypted Datafile")) {
+        reply.send(file);
+        return;
```
paologum, line 111: since we are using this method of checking a header a lot, you should refactor this to a method that returns a boolean. additionally, since we just need the first line and datapacks can be very big it could slow down the process a lot if we are reading in the whole file, so just create a stream and read the first line in that method

server/src/routes.ts
JacquiFreeElectron: I don't think it will because I don't call rm inside` if (needEncryption)` and it will directly return as long as it finds it is an encrypted file. if it's not then the file has to be encrypted by the java program

JacquiFreeElectron: `rm`is only called when 1. the file is in `encrypted-datapacks` 2. the java program gets called and a encrypted-file is generated. Otherwise it will directly return the target file

server/src/routes.ts
JacquiFreeElectron: ok

server/src/routes.ts
paologum: then there isn't really a need to dynamically set a `downloadPath`, you should just return the file if `!needEncryption` in the very beginning and just run the encryption checks after the java file runs

it's more confusing if you merge the two cases together, when you can just return early on without running any extra code

server/src/routes.ts
```diff
@@ -59,6 +60,137 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+        "Java file was unable to encrypt the file" + filename + ", resulting in an incorrect encryption header.";
+      reply.status(422).send({
+        error: errormsg
+      });
+      return;
+    }
+  } catch (e) {
+    const error = e as NodeJS.ErrnoException;
+    if (error.code === "ENOENT") {
+      reply.status(404).send({ error: "File not found" });
```
paologum, line 187: this should be more specific, because anyone looking at this won't understand what `File not found` means

paologum: It doesn't have to be anything crazy just give a little more context

server/src/routes.ts
```diff
@@ -59,6 +60,137 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+    if (lines[0] && lines[0].includes("TSCreator Encrypted Datafile")) {
+      reply.send(file);
+      return;
+    } else {
+      await rm(maybeEncryptedFilepath, { force: true });
+    }
+  } catch (e) {
+    const error = e as NodeJS.ErrnoException;
+    if (error.code !== "ENOENT") {
+      reply.status(500).send({ error: "An error occurred" });
```
paologum, line 95: for this and all other 500 status codes, you should send the error with it

server/src/index.ts
```diff
@@ -138,7 +138,12 @@ server.get("/presets", async (_request, reply) => {
   reply.send(presets);
 });
 // uploads datapack
-server.post("/upload", routes.uploadDatapack);
+server.post<{ Params: { username: string } }>("/upload", routes.uploadDatapack);
+//download datapack
+server.get<{ Params: { filename: string }; Querystring: { needEncryption: boolean } }>(
+  "/requestDownload/:filename",
```
paologum, line 144: this should just be `/downloads/:filename`

paologum: actually it should be `/downloads/user-datapacks/:filename` because we might download settings xml in the future

server/src/index.ts
```diff
@@ -138,7 +138,12 @@ server.get("/presets", async (_request, reply) => {
   reply.send(presets);
 });
 // uploads datapack
-server.post("/upload", routes.uploadDatapack);
+server.post<{ Params: { username: string } }>("/upload", routes.uploadDatapack);
+//download datapack
```
paologum, line 142: this also should stay `server.post("/upload", routes.uploadDatapack);`

paologum:
you should also upload a demo of every use case

server/src/routes.ts
JacquiFreeElectron: I don't think the cases are merged because rn my workflow does directly return if `!needEncryption` without running any extra code, becasue the huge block inside `if(needEncryption)` won't be executed. I think the part I need to change is that I mistakenly put line 79-82 (which checks if we've encrypted a file before) outside `if(needEncryption)`. I should put that inside `if(needEncryption)` or else it will always return the encrypted file even if the user requests to retrieve the original file. Maybe that's the part where you think that I'm running extra code. Besides that, there's no other code that will be executed.

JacquiFreeElectron: But yeah I can put the immediate return before `if(needEncryption)` if you think that's easier to for other people to read my code

JacquiFreeElectron:
See demo in my dropbox

app/src/SettingsTabs/Datapack.tsx
```diff
@@ -26,6 +28,51 @@ export const Datapacks = observer(function Datapacks() {
...
+        <Fragment>
+          <Menu menuButton={<DownloadIcon className="download-icon" />} transition>
+            <MenuItem onClick={() => handleDownload(true, name)}>
+              <Typography>Encrypted Download</Typography>
+            </MenuItem>
+            <MenuItem onClick={() => handleDownload(false, name)}>
+              <Typography>Retrieve Original File</Typography>
+            </MenuItem>
+          </Menu>
+        </Fragment>
```
paologum, line 43: actually technically since `Menu` is one item, you can just remove the `Fragment` completely i think

app/src/SettingsTabs/Datapack.tsx
```diff
@@ -26,6 +28,51 @@ export const Datapacks = observer(function Datapacks() {
...
+    } catch (error) {
+      actions.pushError(ErrorCodes.INVALID_PATH);
+      return "";
+    }
+  }
+
+  const handleDownload = async (needEncryption: boolean, fileName: string) => {
+    const fileURL = await getFileURL(needEncryption, fileName);
+    const aTag = document.createElement("a");
+    if (typeof fileURL === "string") {
```
paologum, line 66: you should add `fileURL && typeof fileURL === "string"` so they don't download an empty file since you return `""` when `getFileURL` has a bad or empty file

app/src/state/actions/general-actions.ts
```diff
@@ -689,6 +689,33 @@ export const fetchImage = action("fetchImage", async (datapackName: string, imag
...
+  });
+  if (!response.ok) {
+    if (response.status === 404) {
+      displayServerError(null, ErrorCodes.FILE_NOT_FOUND, ErrorMessages[ErrorCodes.FILE_NOT_FOUND]);
+    } else if (response.status === 500) {
+      displayServerError(null, ErrorCodes.SERVER_RESPONSE_ERROR, ErrorMessages[ErrorCodes.SERVER_RESPONSE_ERROR]);
+    } else if (response.status === 422) {
+      displayServerError(
+        null,
+        ErrorCodes.INCORRET_ENCRYPTION_HEADER,
```
paologum, line 710: `INCORRECT`

app/src/state/actions/general-actions.ts
```diff
@@ -689,6 +689,33 @@ export const fetchImage = action("fetchImage", async (datapackName: string, imag
...
+  }
+  const response = await fetcher(route, {
+    method: "GET"
+  });
+  if (!response.ok) {
+    if (response.status === 404) {
+      displayServerError(null, ErrorCodes.FILE_NOT_FOUND, ErrorMessages[ErrorCodes.FILE_NOT_FOUND]);
+    } else if (response.status === 500) {
+      displayServerError(null, ErrorCodes.SERVER_RESPONSE_ERROR, ErrorMessages[ErrorCodes.SERVER_RESPONSE_ERROR]);
+    } else if (response.status === 422) {
```
paologum, line 707: can you make this a switch so it's easier to read

app/src/util/error-codes.ts
```diff
@@ -40,7 +40,9 @@ export enum ErrorCodes {
   UNABLE_TO_RESET_EMAIL = "UNABLE_TO_RESET_EMAIL",
   UNABLE_TO_LOGIN_ACCOUNT_LOCKED = "UNABLE_TO_LOGIN_ACCOUNT_LOCKED",
   UNABLE_TO_RECOVER_ACCOUNT = "UNABLE_TO_RECOVER_ACCOUNT",
-  NOT_LOGGED_IN = "NOT_LOGGED_IN"
+  NOT_LOGGED_IN = "NOT_LOGGED_IN",
+  FILE_NOT_FOUND = "FILE_NOT_FOUND",
```
paologum, line 44: `FILE_NOT_FOUND` is too generic it should be `USER_DATAPACK_FILE_NOT_FOUND_FOR_DOWNLOAD`

app/src/util/error-codes.ts
```diff
@@ -92,5 +94,8 @@ export const ErrorMessages = {
   [ErrorCodes.UNABLE_TO_LOGIN_ACCOUNT_LOCKED]:
     "Your account has been locked. Please check your email for more information or contact support.",
   [ErrorCodes.UNABLE_TO_RECOVER_ACCOUNT]: "Unable to recover account. Please contact support for more information.",
-  [ErrorCodes.NOT_LOGGED_IN]: "You are not logged in. Please log in to access this page."
+  [ErrorCodes.NOT_LOGGED_IN]: "You are not logged in. Please log in to access this page.",
+  [ErrorCodes.FILE_NOT_FOUND]: "The file you want to download was not found.",
```
paologum, line 98: this is too direct, use "The datapack requested was not found on the server. Please try again later or contact our support team"

app/src/util/error-codes.ts
```diff
@@ -92,5 +94,8 @@ export const ErrorMessages = {
...
+  [ErrorCodes.INCORRET_ENCRYPTION_HEADER]:
+    "Java file was unable to encrypt the file you want to download, resulting in an incorrect encryption header."
```
paologum, line 100: we don't need to tell the user that the java file couldn't encrypt, this should be a message that the `displayServerError` handles since it takes in the server response variable. The user should just see "Unable to encrypt the file, please try again later."

server/src/parse-datapacks.ts
```diff
@@ -125,7 +125,11 @@ export function spliceArrayAtFirstSpecialMatch(array: string[]): ParsedColumnEnt
  * @param files the files to be parsed
  * @returns
  */
-export async function parseDatapacks(file: string, decryptFilePath: string): Promise<DatapackParsingPack | null> {
+export async function parseDatapacks(
+  file: string,
+  decryptFilePath: string,
+  isUserDataPack?: boolean
```
paologum, line 131: you should do `isUserDatapack?: boolean = false` or however to default it to false then make `isUserDatapack` a required field on `DatapackParsingPack` for better structure, otherwise you have `undefined` `false` and `true` when we just want two of those values.

server/src/routes.ts
```diff
@@ -59,6 +60,137 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
   reply.status(200).send(datapackInfoChunk);
 };
 
+export const requestDownload = async function requestDownload(
+  request: FastifyRequest<{ Params: { filename: string }; Querystring: { needEncryption: boolean } }>,
```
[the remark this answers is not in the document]
paologum: still needs resolution

server/src/routes.ts
```diff
@@ -59,6 +60,128 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+  const datapackDir = path.join(userDir, "datapacks");
+  const filepath = path.join(datapackDir, filename);
+  const encryptedFilepathDir = path.join(userDir, "encrypted-datapacks");
+  const maybeEncryptedFilepath = path.join(encryptedFilepathDir, filename);
+  if (needEncryption === undefined) {
+    await access(filepath);
+    const file = await readFile(filepath);
+    reply.send(file);
+    return;
+  } else if (needEncryption) {
```
paologum, line 85: so we aren't nesting you can just do `if (!needEncryption)` then you won't need an `else if` because you return

server/src/routes.ts
```diff
@@ -59,6 +60,128 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+            resolve();
+          }
+        });
+      });
+    } catch (e) {
+      console.error(e);
+      reply.status(500).send({ error: "Failed to encrypt datapacks with error " + e });
+      return;
+    }
+    const downloadPath = path.join(encryptedFilepathDir, filename);
```
paologum, line 156: this is a duplicate variable of `maybeEncryptedFile`

server/src/routes.ts
```diff
@@ -59,6 +60,128 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+      await access(filepath);
+      const file = await readFile(filepath);
+      if (await checkHeader(filepath)) {
+        reply.send(file);
+        return;
+      }
+    } catch (e) {
+      const error = e as NodeJS.ErrnoException;
+      if (error.code === "ENOENT") {
+        reply.status(404).send({ error: "File not found" });
```
paologum, line 112: this should be more specific that it can't find the user datapack requested

paologum (changes requested):

server/src/routes.ts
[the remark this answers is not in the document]
paologum: it should be `needEncryption?: boolean`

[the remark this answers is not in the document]
JacquiFreeElectron: Oh do you mean just do `Querystring: { needEncryption?: boolean }`?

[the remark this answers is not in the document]
paologum: yup!

server/src/routes.ts
```diff
@@ -59,6 +60,128 @@ export const fetchServerDatapackInfo = async function fetchServerDatapackInfo(
...
+  const { needEncryption } = request.query;
+  const { filename } = request.params;
+  const userDir = path.join(assetconfigs.uploadDirectory, uuid);
+  const datapackDir = path.join(userDir, "datapacks");
+  const filepath = path.join(datapackDir, filename);
+  const encryptedFilepathDir = path.join(userDir, "encrypted-datapacks");
+  const maybeEncryptedFilepath = path.join(encryptedFilepathDir, filename);
+  if (needEncryption === undefined) {
+    await access(filepath);
+    const file = await readFile(filepath);
```
paologum, line 82: also this should be in a try

app/src/state/actions/general-actions.ts
```diff
@@ -689,6 +689,44 @@ export const fetchImage = action("fetchImage", async (datapackName: string, imag
...
+        );
+        break;
+      }
+      case 500: {
+        displayServerError(null, ErrorCodes.SERVER_RESPONSE_ERROR, ErrorMessages[ErrorCodes.SERVER_RESPONSE_ERROR]);
+        break;
+      }
+      case 422: {
+        displayServerError(
+          null,
```
paologum, line 718: you should send `response` as the first parameter

app/src/util/error-codes.ts
```diff
@@ -40,7 +40,9 @@ export enum ErrorCodes {
   UNABLE_TO_RESET_EMAIL = "UNABLE_TO_RESET_EMAIL",
   UNABLE_TO_LOGIN_ACCOUNT_LOCKED = "UNABLE_TO_LOGIN_ACCOUNT_LOCKED",
   UNABLE_TO_RECOVER_ACCOUNT = "UNABLE_TO_RECOVER_ACCOUNT",
-  NOT_LOGGED_IN = "NOT_LOGGED_IN"
+  NOT_LOGGED_IN = "NOT_LOGGED_IN",
+  USER_DATAPACK_FILE_NOT_FOUND_FOR_DOWNLOAD = "FILE_NOT_FOUND",
```
paologum, line 44: change `FILE_NOT_FOUND`

server/__tests__/__data__/column-keys.json
paologum, line 1: to make this a cleaner merge, can you just copy paste the original file so there isn't 4000 lines of difference and change only the couple of lines you need. the reason for these test files isn't to copy paste every time we change something but to change the small things we add

server/src/index.ts
```diff
@@ -139,6 +139,11 @@ server.get("/presets", async (_request, reply) => {
 });
 // uploads datapack
 server.post("/upload", routes.uploadDatapack);
+//download datapack
+server.get<{ Params: { filename: string }; Querystring: { needEncryption: boolean } }>(
```
paologum, line 143: `needEncryption?:`

server/src/parse-datapacks.ts
```diff
@@ -278,7 +282,8 @@ export async function parseDatapacks(file: string, decryptFilePath: string): Pro
     show: true
   };
   setShowLabels(chartColumn);
-  const datapackParsingPack = { columnInfo: chartColumn, ageUnits, defaultChronostrat, formatVersion };
+  const isUserDatapack = isUserDataPack;
```
paologum, line 285: no need for this variable?

server/src/util.ts
```diff
@@ -171,3 +172,20 @@ export function setCommonProperties<T>(o1: T, o2: Partial<T>): T {
...
+    for await (const line of readline) {
+      isEncrypted = line.includes("TSCreator Encrypted Datafile");
+      break;
+    }
+  } catch (e) {
+    return false;
+  }
+
+  return isEncrypted;
+}
```
paologum, line 191: we should have a small test set for this, but not a part of this pr

paologum (changes requested):

server/src/parse-datapacks.ts
JacquiFreeElectron: I tried to eliminate this variable but the `isUserDatapack` of datapackParsingPack will become undefined. Not sure why. I can try again but I think I need this variable

app/src/state/actions/general-actions.ts
JacquiFreeElectron: Do you mean `displayServerError(response, ErrorCodes.SERVER_RESPONSE_ERROR, ErrorMessages[ErrorCodes.SERVER_RESPONSE_ERROR]);`

app/src/util/error-codes.ts
JacquiFreeElectron: right🤦🏻‍♀️

server/__tests__/__data__/column-keys.json
JacquiFreeElectron: ok

server/src/util.ts
JacquiFreeElectron: sure

server/__tests__/__data__/column-keys.json
JacquiFreeElectron: oh but I think the reason there are 4000 lines of difference is that the lines I modified moved the places of other lines. Probably copy-paste the original file won't fix that?  I can try it tho.

server/src/parse-datapacks.ts
JacquiFreeElectron: I tried again and I have to either keep this variable or do `datapackParsingPack.isUserDatapack = isUserDataPack` after line 286. Otherwise, it will become undefined.

app/src/state/actions/general-actions.ts
paologum: yes

server/src/parse-datapacks.ts
paologum: it's because you're passing it as `isUserDataPack` instead of `isUserDatapack`

JacquiFreeElectron: ahh yes. Changed it

paologum (approved):
nice job!!:)

PR #103, opened 2024-05-12
Tsc 331 style emails

asivath:
PreHeader:
image
Email with button:
image
One more with button:
image
Email without button:
image


paologum:
future note: we should add contact details here or a link to the support page

server/src/login-routes.ts
```diff
@@ -196,10 +190,16 @@ export const resetEmail = async function resetEmail(
...
+      preHeader: "Immediate Action Required",
+      title: "Your Email Address Has Been Changed",
+      message: `We have recorded a change to your email address. If you did not initiate this change, it's crucial to act immediately to secure your account.<br/><br/>
+                Please use the button below to reset your email and password. This link will expire in 24 hours for security reasons. If the link has expired, please contact our support team directly for assistance.`,
+      link: `${process.env.APP_URL || "http://localhost:5173"}/account-recovery?token=${token}&email=${email}`,
+      buttonText: "Reset Email and Password",
+      action: "Account Recovery"
     };
-    sendEmail(invalidateEmail);
+    await sendEmail(invalidateEmail);
```
paologum, line 202: do we need to `await` these now?

server/src/send-email.ts
```diff
@@ -0,0 +1,297 @@
...
+    user: process.env.EMAIL_USER,
+    pass: process.env.EMAIL_PASS
+  }
+});
+
+const HtmlTemplate = (
+  preHeader: string,
+  title: string,
+  message: string,
+  action: string,
```
paologum, line 18: do you need to sanitize these just in case a user email or something has bad stuff in it, or a username?

also for future reliability for other programmers to not call this template with dumb things

paologum (changes requested):

server/src/login-routes.ts
asivath: initially i didn't want to await because sending an email could take a bit but it can throw errors and that should be shown to the user

server/src/send-email.ts
asivath: yeah i'll use dompurify for this

paologum (approved):

PR #104, opened 2024-05-12
Tsc 310 backend for load settings

jay-jhk-lee:
added the backend for loading settings files. Created a function to recursively go down the info given by the settings file and accessing the respective column info state. Standardized the parameters of the setter functions for columnInfo. Added formatting to the column names when parsing datapack to align with tscreator and the settings file.


app/src/SettingsTabs/EditNameField.tsx
```diff
@@ -0,0 +1,42 @@
...
+        />
+        <div className="edit-title-button">
+          <Button
+            color="secondary"
+            variant="contained"
+            onClick={() => {
+              actions.setEditName(editName.current, column);
+            }}>
+            Confirm
+          </Button>
```
paologum, line 37: if you're already changing/moving this
1. the team discussed removing the confirm removal, so can you remove that button and change the event to `onChange` of the Text Field
2. instead of using a `useRef` hook, just make the value `column.editName` (this is sorta the same as 1.)
3. remove the css for the button
4. if you can make this a react fragment ( verify it works as a react fragment )

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,64 @@
...
+  setColumnOn(settings.isSelected, column);
+  if (settings.width) setWidth(settings.width, column);
+  if (settings.backgroundColor.text) setRGB(settings.backgroundColor.text, column);
+  column.fontsInfo = cloneDeep(settings.fonts);
+  switch (extractColumnType(settings._id)) {
+    case "EventColumn":
+      assertEventColumnInfoTSC(settings);
+      assertEventSettings(column.columnSpecificSettings);
+      if (column.columnSpecificSettings)
+        setEventColumnSettings(column.columnSpecificSettings, { type: settings.type, rangeSort: settings.rangeSort });
```
paologum, line 35: you will need to add `PointSettings` because it's on main now

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,64 @@
...
+export const applyChartColumnSettings = action("applyChartColumnSettings", (settings: ColumnInfoTSC) => {
+  const columnName = extractName(settings._id);
+  let curcol: ColumnInfo | undefined =
+    state.settingsTabs.columnHashMap.get(columnName) ||
+    state.settingsTabs.columnHashMap.get("Chart Title in " + columnName);
+  let blockSeriesColumns: boolean = false;
+  if (curcol === undefined) {
+    pushSnackbar("Unknown column name '" + columnName + "' found while loading settings", "warning");
+    //some names reach letter count limit for snackbar, so log it
+    console.log("Unknown column name " + columnName + " found while loading settings");
```
paologum, line 47: There should be a hard limit to the max character count that we can push to a snackbar. Not sure what it is but this should truncate `columnName` to fit that character count. If it's not a constant variable and we just set it somewhere can you move it to `constant.ts`

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,64 @@
...
+    for (let i = 0; i < settings.children.length; i++) {
+      curcol = state.settingsTabs.columnHashMap.get(columnName + " " + extractName(settings.children[i]._id));
+      if (curcol !== undefined) setColumnProperties(curcol, settings.children[i]);
+    }
+  }
+  if (blockSeriesColumns === false) {
+    for (let i = 0; i < settings.children.length; i++) {
+      applyChartColumnSettings(settings.children[i]);
+    }
+  }
```
paologum, line 60: we shouldn't need `blockSeriesColumn` right? we can just do `if... else`? or is there a case im unaware of

app/src/state/parse-settings.ts
```diff
@@ -122,6 +122,49 @@ function processSettings(settingsNode: Element): ChartSettingsInfoTSC {
...
+          key = child.getAttribute("function")! as keyof FontsInfo;
+        } else continue;
+        fonts[key].inheritable = Boolean(child.getAttribute("inheritable"));
+
+        if (fontProps[i].includes("font-family")) {
+          fonts[key].fontFace =
+            fontProps[i].split(": ")[1] === "Arial" || "Courier" || "Verdana"
+              ? <"Arial" | "Courier" | "Verdana">fontProps[i].split(": ")[1]
+              : "Arial";
+        }
```
paologum, line 146: previous comment of 

it would be beneficial for us to split `fontProps[i]` early by ": " and call it like `fontPropKey` then check includes on that for less comparisons. also the nested ifs can just be one if statement with `&&`

paologum (changes requested):

app/src/state/actions/column-actions.ts
jay-jhk-lee: that should work

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,105 @@
...
+          let presetPointShape: PointShape = "nopoints";
+          if (settings.drawPoints === true) {
+            if (settings.pointType === "round") {
+              presetPointShape = "circle";
+            } else if (settings.pointType === "tick") {
+              presetPointShape = "cross";
+            } else {
+              presetPointShape = "rect";
+            }
+          }
```
paologum, line 56: i made a `util` function for this that converts `PointShape` to the other type on my pr, just a comment for how you or i might rebase fyi

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,105 @@
...
+    }
+  }
+  const columnName = extractName(settings._id);
+  let curcol: ColumnInfo | undefined =
+    state.settingsTabs.columnHashMap.get(columnName) ||
+    state.settingsTabs.columnHashMap.get("Chart Title in " + columnName);
+  if (curcol === undefined) {
+    pushSnackbar(
+      "Unknown column name '" + columnName.substring(0, snackbarTextLengthLimit - 1) + "' found while loading settings",
+      "warning"
```
paologum, line 88: don't you have to account for the other string lengths as well?

app/src/state/actions/column-actions.ts
```diff
@@ -1,6 +1,105 @@
...
+    );
+    console.log(columnName);
```
paologum, line 90: remove

app/src/state/actions/column-actions.ts
```diff
@@ -41,29 +140,43 @@ export const toggleSettingsTabColumn = action((name: string) => {
...
+  if (newSettings.scaleStart) pointSettings.scaleStart = newSettings.scaleStart;
+  if (newSettings.scaleStep) pointSettings.scaleStep = newSettings.scaleStep;
+  if (newSettings.fill) pointSettings.fill = newSettings.fill;
+  if (newSettings.pointShape) pointSettings.pointShape = newSettings.pointShape;
+  if (newSettings.lowerRange) pointSettings.lowerRange = newSettings.lowerRange;
+  if (newSettings.upperRange) pointSettings.upperRange = newSettings.upperRange;
+  if (newSettings.smoothed) pointSettings.smoothed = newSettings.smoothed;
+  if (newSettings.minX) pointSettings.minX = newSettings.minX;
+  if (newSettings.maxX) pointSettings.maxX = newSettings.maxX;
+});
```
paologum, line 164: oops, on my pr i change it to `Object.assign` so it's just one line and you should also probably go that way. I realized since we can gurantee that the `newSettings` has all of the `PointSettings` fields, we can just assign since there will be no new keys

server/src/parse-datapacks.ts
```diff
@@ -769,7 +766,9 @@ export async function getColumnTypes(
  * @param tabSeparated
  */
 function setColumnHeaders(column: ColumnHeaderProps, tabSeparated: string[]) {
-  column.name = tabSeparated[0]!;
+  //for formatted names in ColumnInfo object
+  column.name = formatColumnName(tabSeparated[0]!);
+  if (column.name.includes("Nanno Zone")) console.log(column.name);
```
paologum, line 771: remove

server/src/parse-datapacks.ts
```diff
@@ -383,20 +384,16 @@ export async function getAllEntries(
...
-    //if this is not done, then the keys in the state for columns have quotations surrounding it
-    //which is not consistent with the equivalent keys found in the parsed settings json object.
-    //ex "North Belgium -- Oostende, Brussels, Antwerp, Campine, Maastrichen" vs
-    //North Belgium -- Oostende, Brussels, Antwerp, Campine, Maastrichen
+    let parent = line.split("\t:\t")[0];
 
     const childrenstring = line.split("\t:\t")[1];
     if (!parent || !hasVisibleCharacters(parent) || !childrenstring || !hasVisibleCharacters(childrenstring)) continue;
-    // childrenstring = childrenstring!.split("\t\t")[0];
+    parent = formatColumnName(parent);
```
paologum, line 391: will this propagate to the recursive function properly

app/src/state/parse-settings.ts
```diff
@@ -266,7 +297,8 @@ export function xmlToJson(xml: string): ChartInfoTSC {
       settingsTSC["class datastore.RootColumn:Chart Root"]._id = "class datastore.RootColumn:Chart Root";
     }
   }
-  assertChartInfoTSC(settingsTSC);
+
+  //assertChartInfoTSC(settingsTSC);
```
paologum, line 301: you can remove this

paologum (changes requested):

app/src/state/actions/column-actions.ts
jay-jhk-lee: What other string lengths are you referring to?

server/src/parse-datapacks.ts
jay-jhk-lee: This will properly work for the key and values in the `allEntries` mapping

app/src/state/actions/column-actions.ts
paologum: the whole string is `snackbarTextLengthLimit` not just the `columnName`

server/src/parse-datapacks.ts
paologum: ok looks good

paologum (approved):

PR #105, opened 2024-05-14
curve drawing frontend, added portal functionality to color picker

paologum:
- Point Column now draws correctly on graph and settings are correctly calculated
- Color picker now will adjust itself, if portal is requested, to fit viewport and will attach to `document.body` to view on top of modal
- NOTE: line color and draw points don't work in the java file when loading settings file so we'll have to take a look at why it ignores those fields
- changed reset settings to reset to 0 - 10 age range

image

https://github.com/earthhistoryviz/tsconline/assets/89664317/191b96ca-dc74-49f5-a9e7-ff2293615673


jay-jhk-lee (approved):
looks good!

PR #107, opened 2024-05-15
Tsc 336 create diagram for login flow

asivath:
Updated documentation for the login routes


assets/Database-Structure.drawio.png
paologum, line 1: can you add a quick description of what verify, password, and invalidate means in the verification table

assets/Login-Flows.drawio.png
paologum, line 1: what does password verification on the left mean i forgot

paologum, line 1: one thing that confused me was that we say email sent for avoiding attacks on user existence, but on signup don't we tell them that a user already exists with that email? isn't that the same thing and might provoke attacks?

asivath: the first verification is checking that the user inputs are valid, so not empty. the second one is checking if the password is correct. I'm going to replace the first one with "Validate User Inputs".

asivath: it is based on username or email matching, but yes its unsecure. thats why we need captcha/rate limiting. we could also just have a vague error message for everything.

paologum (approved):

PR #108, opened 2024-05-17
TSC-46 Shows error on server when jar is missing

Changk820:
Logs error to console and ends process when the jar file is missing or name is wrong compared to what's listed in the config.json file.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -3,7 +3,7 @@ import cors from "@fastify/cors";
 import fastifyStatic from "@fastify/static";
 import process from "process";
 import { execSync } from "child_process";
-import { deleteDirectory } from "./util.js";
+import { deleteDirectory, checkFileExists } from "./util.js";
 import * as routes from "./routes.js";
 import * as loginRoutes from "./login-routes.js";
 import { DatapackIndex, MapPackIndex, assertIndexResponse } from "@tsconline/shared";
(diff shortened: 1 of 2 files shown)


server/src/index.ts
```diff
@@ -18,6 +18,18 @@ import fastifySecureSession from "@fastify/secure-session";
...
+  try {
+    await readFile(filePath);
+    console.log(`File exists: ${filePath}`);
+    return true;
+  } catch (error) {
+    console.log(`File does not exist: ${filePath}`, error);
+    return false;
+  }
+}
+
```
paologum, line 32: you should probably put this in `util.ts` on the `server` and also we probably don't want to read the whole file since we just want to see if the file exists

here you can use `await access` or `existsSync` 

(if you choose to use `existsSync` we don't need to abstract a new method)

async vs await doesn't really matter in this scenario since this runs once on server startup and also won't have any bad race conditions

server/src/index.ts
```diff
@@ -44,6 +56,21 @@ try {
...
+const [jar1Exists, jar2Exists] = await Promise.all([checkFileExists(jar1Path), checkFileExists(jar2Path)]);
+
+if (!jar1Exists) {
+  console.error("ERROR: Required active JAR file does not exist:", jar1Path);
+  process.exit(1);
+}
+if (!jar2Exists) {
+  console.error("ERROR: Required decrypt JAR file does not exist:", jar2Path);
+  process.exit(1);
+}
```
paologum, line 72: we can probably make this more compact into

`if (await checkFileExists(...))`

also `jar1Path` and `jar2Path` are ambiguous so you should be more descriptive with the names

lastly use `path.join` when dealing with paths. also since this is all relative you don't need `process.cwd()` you can just have the file path to be checked as `assetconfigs.activeJar`

paologum (changes requested):

PR #109, opened 2024-05-20
fix bug where lone column (column with no parent) doesn't have chart settings

paologum:
refactored a good amount of `parse-datapacks.ts` to better reflect the algorithm

added `chalk` to view the logged messages easier

changed a good amount of tests to reflect refactoring

now we don't need a map for every type, we just create a large map of every column, then fill out the relationship recursively

new `datapack-decrypter.jar` that writes everything in UTF-8 for consistency in encoding


JacquiFreeElectron (commented):
👍🏻

PR #112, opened 2024-05-20
Tsc 301 limit password retries use captcha

asivath:
Added recaptcha v3 to auth pages. I also changed the logic in account recovery to directly send the user an email with a password reset token instead of asking them to go to the reset password form and send themselves an email. Recaptcha v3 is invisible and sends a score to the server that represents how likely a user is a bot. Also added rate limiting and logging ips that exceed the limit so we can send a weekly email about those ips.


app/src/SignUp.tsx
```diff
@@ -25,6 +25,13 @@ export const SignUp: React.FC = observer(() => {
   const [loading, setLoading] = useState(false);
   const { actions } = useContext(context);
 
+  useEffect(() => {
+    loadRecaptcha();
+    return () => {
+      removeRecaptcha();
+    };
+  }, []);
```
paologum, line 33: might be a dumb question but why don't we just put the loading and removing reCAPTCHA on the general `app.tsx`
does it need to reload for every component?

server/src/database.ts
```diff
@@ -31,6 +31,11 @@ Database Schema Details (Post-Migration):
   - expiresAt (datetime): Non-nullable, the expiration date/time of the token. Make sure to always use ISO 8601 format. Easy way to get this is by using new Date().toISOString().
   - reason (text): Non-nullable, describes the purpose of the token (e.g., 'email verification', 'password reset').
 
+- ip Table:
+  - id (integer): Primary key, auto-increment.
+  - ip (text): Non-nullable, must be unique, stores the user's IP address.
+  - count (integer): Non-nullable, default is 1, stores the number of times the IP has caused a rate limit violation.
+
```
paologum, line 38: at some point we have to change the `draw.io` stuff so can you send the editable files (not the pngs) in the dropbox so we can edit them easily? or you can just push them to the `gh-pages` either way is fine

server/src/index.ts
```diff
@@ -242,6 +266,36 @@ setInterval(
   },
   1000 * 60 * 60 * 24 * 7
 ); // 1 week
+dotenv.config();
+if (process.env.EMAIL_USER && process.env.EMAIL_PASS && process.env.NODE_ENV === "production") {
+  const emailUser = process.env.EMAIL_USER as string;
+  cron.schedule(
+    "0 0 * * 0", // every Sunday at midnight
```
paologum, line 273: can you send me the credentials so i can login and see these

server/src/index.ts
```diff
@@ -242,6 +266,36 @@ setInterval(
...
+          from: emailUser,
+          to: emailUser,
+          subject: "Timescale Creator IP Notification",
+          preHeader: "Timescale Creator IP Notification",
+          title: "Timescale Creator IP Notification",
+          message: `The following IPs have been rate limited:<br>${ipMessages}`,
+          action: "Ip Notification"
+        };
+        await sendEmail(notificationEmail);
+        await db.deleteFrom("ip").execute();
```
paologum, line 292: so we reset the ips every sunday?

server/src/login-routes.ts
```diff
@@ -480,15 +535,21 @@ export const signup = async function signup(
...
-  if (!username || !password) {
+  const { username, password, recaptchaToken } = request.body;
+  if (!username || !password || !recaptchaToken) {
     reply.status(400).send({ error: "Invalid form" });
     return;
   }
   try {
+    const score = await checkRecaptchaToken(recaptchaToken);
+    // the score is a number between 0 and 1 that indicates the likelihood that the user is a bot
+    if (score < 0.5) {
```
paologum, line 549: for easier changing if we wanna be more lenient, can you make 0.5 a constant from `constant.ts` or at the top of the page

server/src/login-routes.ts
```diff
@@ -480,15 +535,21 @@ export const signup = async function signup(
...
 
 export const login = async function login(
-  request: FastifyRequest<{ Body: { username: string; password: string } }>,
+  request: FastifyRequest<{ Body: { username: string; password: string; recaptchaToken: string } }>,
   reply: FastifyReply
 ) {
-  const { username, password } = request.body;
-  if (!username || !password) {
+  const { username, password, recaptchaToken } = request.body;
+  if (!username || !password || !recaptchaToken) {
```
paologum, line 542: i'm confused, in the `login` component you send `null` for the `recaptchaToken` so wouldn't that mean it never gets past this check

app/src/SignUp.tsx
asivath: we could that would mean having recaptcha on everything instead of just the auth routes which would also include a logo in the bottom right of every page for the recaptcha

server/src/database.ts
asivath: they're already in the assets folder of gh-pages, they can be edited using the draw.io extension in vscode

server/src/login-routes.ts
asivath: the login component doesn't send null it's just that the object passed to handleLogin contains a null recaptchaToken field which is then filled in by handleLogin before sending it to the server

app/src/SignUp.tsx
paologum: ok yeah nvm then we should be ok for now, but we might do that in the future just because i mean technically someone could spam request charts

server/src/database.ts
paologum: ok cool sounds good

app/src/SignUp.tsx
```diff
@@ -61,6 +74,12 @@ export const SignUp: React.FC = observer(() => {
           case 409:
             errorCode = ErrorCodes.UNABLE_TO_SIGNUP_USERNAME_OR_EMAIL;
             break;
+          case 422:
+            errorCode = ErrorCodes.RECAPTCHA_FAILED;
+            break;
+          case 429:
```
paologum, line 80: looks good but i don't see where you send the 429 status being sent, so is this a response that comes with pinging the google api too much?

paologum (approved):

PR #114, opened 2024-05-22
data mining settings

paologum:
changed jar file with following updates

- new xml tags for point and event settings
- event ( `drawExtraColumn` of types `FAD` `LAD` `Combined` `Overlay` ) `Overlay` isn't supported for now as it is a comparison of two columns and slightly unrelated to the data mining
- point (`drawExtraColumn` of types `Frequency` `Maximum Value` `Minimum Value` `Average Value` `Rate of Change` `Overlay`) with same `Overlay` notes
- both have `windowSize` and `stepSize` default to 2 and 1 respectively
- known and pre-existing error -> drawing a point datamining column will overlap into the next column if there exists one

General Changes
- radio group has a clear and header margin customization option
- renamed folders to be consistent
- changed functionality so that changes to columns do not actually change the `DatapackIndex` columns to preserve consistency on the app side (might discuss changing this back but atm this is needed because otherwise columns will be recursively added if you add and remove the same datapack)

https://github.com/earthhistoryviz/tsconline/assets/89664317/302cdd0e-4c8a-48af-be76-e9d63299da8e


jay-jhk-lee (approved):
Looks Good!

PR #117, opened 2024-05-27
support for customization of data mining column on the front and backend and datamining range calculations

paologum:
- does not include reading settings xml to create a new column yet (TSC-370)
- new xml tag in column tag `isDataMiningColumn` will let you know if this was generated from the datapack or by another column

- fix for if there are no column relationships in file and if it just contains "blocks" or "lone columns"
- created datamining that mimics that of the java file to appropriately scale the new columns
- tests for above
- added fields to parsing of events (FAD, LAD, EVENT, EVENTS)
- can clear the field to remove the added column
- selecting any data mining option will create that column and subsequent clicks on other radios will remove and add respectively
- fixed parent relationship bug that it wasn't getting set on multiple datapacks being added together

https://github.com/earthhistoryviz/tsconline/assets/89664317/fd376747-f065-47d4-b4d4-91d38219e460

### DIFF
--- a/app/src/util/data-mining.ts
+++ b/app/src/util/data-mining.ts
@@ -63,7 +63,7 @@ type DataPoint = {
   value: number;
 };
 
-export function computeWindowStatisticsForDataSet(
+export function computeWindowStatisticsForDataPoints(
   data: DataPoint[],
   windowSize: number,
   stat: DataMiningStatisticApproach
--- a/app/src/util/data-mining.ts
+++ b/app/src/util/data-mining.ts
@@ -1,5 +1,6 @@
 import { round } from "lodash";
 import { DataMiningStatisticApproach, WindowStats } from "../types";
+import { normalizeZero } from "./util";
 
 /**
  * This function computes the statistics of a data set in a moving window.
@@ -19,8 +20,9 @@ export function computeWindowStatistics(
     return [];
   }
   data = data.sort((a, b) => a - b);
+  const firstDataPoint = data[0]!;
   const lastDataPoint = data[data.length - 1]!;
-  const windows = Math.ceil(lastDataPoint / windowSize);
+  const windows = Math.ceil((lastDataPoint - firstDataPoint + 1) / windowSize);
   const results: WindowStats[] = [];
   let start = data[0]!; // inclusive
   let end = start + windowSize; // exclusive
@@ -31,8 +33,8 @@ export function computeWindowStatistics(
     if (window.length === 0) {
       results.push({ windowStart: start, windowEnd: end, value: 0 });
     } else {
-      const firstWindowPoint = window[0];
-      const lastWindowPoint = window[window.length - 1];
+      const firstWindowPoint = window[0]!;
+      const lastWindowPoint = window[window.length - 1]!;
       let value = 0;
       switch (stat) {
         case "frequency":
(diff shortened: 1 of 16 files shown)


app/src/util/data-mining.ts
```diff
@@ -0,0 +1,132 @@
...
+        break;
+      case "minimum":
+        value = Math.min(...window);
+        break;
+      case "maximum":
+        value = Math.max(...window);
+        break;
+      case "average":
+        value = window.reduce((a, b) => a + b, 0) / window.length;
+        break;
```
asivath, line 47: should be a if-else for window.length check? if window.length is 0 case average can have a divide by zero error

app/src/util/data-mining.ts
```diff
@@ -0,0 +1,132 @@
...
+      case "rateOfChange":
+        if (!firstWindowPoint || !lastWindowPoint) value = 0;
+        else value = (lastWindowPoint - firstWindowPoint) / firstWindowPoint;
```
asivath, line 50: divide by 0 check?

app/src/util/data-mining.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @param windowSize the size of the window
+ * @param stat The statistic to compute in the window
+ * @returns
+ */
+export function computeWindowStatistics(
+  data: number[],
+  windowSize: number,
+  stat: DataMiningStatisticApproach
+): WindowStats[] {
+  if (data.length === 0 || windowSize === 0 || data.some((d) => isNaN(d))) {
```
asivath, line 18: windowSize <= 0

app/src/util/data-mining.ts
```diff
@@ -0,0 +1,132 @@
...
+        else value = (lastWindowPoint.value - firstWindowPoint.value) / firstWindowPoint.value;
+        break;
+    }
+    results.push({ windowStart: start, windowEnd: end, value: round(value, 2) });
+    if (!lastDataPointAge || end > lastDataPointAge) break;
+    start = end;
+    end = Math.min(end + windowSize, lastDataPointAge);
+  }
+  return results;
+}
```
asivath, line 121: same comments from above

app/src/util/data-mining.ts
paologum: oops yeah i should've put a continue after the if

app/src/util/data-mining.ts
paologum: 0 is falsy so it will go to `value = 0` first

paologum: hmm actually not related but if it's rate of change, then `lastWindowPoint` can still be zero right. i'll change it to that but lmk if that's wrong

asivath: yeah it'd be zero either way anyway

paologum: yes value can be negative

app/src/util/util.ts
```diff
@@ -117,3 +117,7 @@ export function convertTSCColorToRGB(text: string): RGB {
     b: Number(rgb[2])
   };
 }
+
+export function normalizeZero(value: number): number {
+  return value === 0 ? 0 : value;
+}
```
asivath, line 123: what is this for doesn't it just return the same value?

paologum: typescript is dumb so `-0 != 0` so if the rate of change divides by a negative number when the numerator is 0, it returns `0`

one of my test cases will cover it so if this removes it, one of the tests will fail

paologum: oh shoot but i need to switch the numbers, oops

paologum: ok nvm i misunderstood, `-0 = 0` but it displays incorrectly, so for display and test purposes i normalize it

asivath (approved):

PR #118, opened 2024-05-27
Tsc 303 add account customization and deleting account

asivath:
https://github.com/earthhistoryviz/tsconline/assets/94872897/42da7d7e-e049-4321-aa57-e39ff0495caa

Things that need to be added are forgot username and also admin controls

Forgot delete profile in the first video:

https://github.com/earthhistoryviz/tsconline/assets/94872897/ecb70a28-84cf-4c46-84f2-6e8680974a41


paologum:
your video shows only username and delete account but i assume the change email, change password are the same as the signup/login page flows? also does the change profile picture works? and anything else that you didn't demo. i haven't yet looked at the code btw

asivath:
the first video shows me changing the password. it works differently from the login page since it doesn't require you to go through your email, instead just needing your current password and the session. email change follows the normal flow. the profile picture does work I show me changing it to the tscreator logo. none of the settings stuff do anything just frontend.

app/src/Profile.tsx
```diff
@@ -0,0 +1,459 @@
...
+  const deleteProfile = async () => {
+    setLoading(true);
+    try {
+      const response = await fetcher("/auth/delete-profile", {
+        method: "POST",
+        credentials: "include"
+      });
+      if (response.ok) {
+        actions.removeAllErrors();
+        actions.pushSnackbar("Profile deleted successfully. Please log in again.", "success");
```
paologum, line 160: i think we can just keep it `Profile deleted successfully` and not the second part since they don't technically need/want to log in again

app/src/Profile.tsx
```diff
@@ -0,0 +1,459 @@
...
+                      }
+                      const ext = file.name.split(".").pop();
+                      if (!file.type.startsWith("image/")) {
+                        actions.pushError(ErrorCodes.UNRECOGNIZED_IMAGE_FILE);
+                        return;
+                      }
+                      if (!ext || !/^(jpg|jpeg|png|gif)$/.test(ext)) {
+                        actions.pushError(ErrorCodes.UNRECOGNIZED_IMAGE_FILE);
+                        return;
+                      }
```
paologum, line 385: styling comment so you don't have to do this but you can just combine these

app/src/Profile.tsx
```diff
@@ -0,0 +1,459 @@
...
+                      try {
+                        const response = await fetcher("/upload-profile-picture", {
+                          method: "POST",
+                          body: formData,
+                          credentials: "include"
+                        });
+                        if (response.ok) {
+                          const data = await response.json();
+                          actions.setPictureUrl(data.pictureUrl);
+                          window.location.reload();
```
paologum, line 398: does this not update properly since `pictureUrl` is observable? bc i think this causes the site to flash right

app/src/Profile.tsx
```diff
@@ -0,0 +1,459 @@
...
+import { Lottie, TSCButton, TSCPopupDialog, InputFileUpload } from "./components";
+import loader from "./assets/icons/loading.json";
+import { useNavigate } from "react-router";
+import Button from "@mui/material/Button";
+
+type Profile = {
+  pictureUrl: string | null;
+};
+
+export const Profile: React.FC<Profile> = observer(({ pictureUrl }) => {
```
paologum, line 30: since most of the values are kept in state that are used in this component, shouldn't we just access `pictureUrl` through the state

app/src/components/TSCAccountMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <>
+      <Box sx={{ display: "flex", alignItems: "center", textAlign: "center" }}>
+        <Tooltip title="Account settings">
+          <IconButton
+            onClick={handleClick}
+            size="small"
+            sx={{ ml: 2 }}
+            aria-controls={open ? "account-menu" : undefined}
+            aria-haspopup="true"
+            aria-expanded={open ? "true" : undefined}>
```
paologum, line 39: small q what does `aria` do

server/src/login-routes.ts
```diff
@@ -37,6 +44,155 @@ async function checkRecaptchaToken(token: string): Promise<number> {
...
+  if (!uuid) {
+    reply.status(401).send({ error: "Not logged in" });
+    return;
+  }
+  try {
+    await deleteUser({ uuid });
+    const userDirectory = path.join(assetconfigs.uploadDirectory, uuid);
+    await fs.promises.rm(userDirectory, { recursive: true });
+    request.session.delete();
+    reply.send({ message: "Profile deleted" });
```
paologum, line 58: also delete file metadata entries

server/src/login-routes.ts
```diff
@@ -37,6 +44,155 @@ async function checkRecaptchaToken(token: string): Promise<number> {
...
+  } catch (error) {
+    console.error("Error during profile deletion:", error);
+    reply.status(500).send({ error: "Unknown Error" });
```
paologum, line 61: shouldn't we send error object?

app/src/Profile.tsx
```diff
@@ -0,0 +1,459 @@
...
+          case 409:
+            errorCode = ErrorCodes.INCORRECT_PASSWORD;
+            break;
+          case 422:
+            errorCode = ErrorCodes.RECAPTCHA_FAILED;
+            break;
+          case 429:
+            actions.removeAllErrors();
+            errorCode = ErrorCodes.TOO_MANY_REQUESTS;
+            break;
```
paologum, line 134: no 403 status? they might not understand if they have a google account

server/src/login-routes.ts
```diff
@@ -37,6 +44,155 @@ async function checkRecaptchaToken(token: string): Promise<number> {
...
+      reply.status(409).send({ error: "Incorrect password" });
+      return;
+    }
+    const newHashedPassword = await hash(newPassword, 10);
+    await updateUser({ uuid }, { hashedPassword: newHashedPassword });
+    request.session.delete();
+    reply.send({ message: "Password changed" });
+  } catch (error) {
+    console.error("Error during password change:", error);
+    reply.status(500).send({ error: "Unknown Error" });
```
paologum, line 104: same with error obj

server/src/login-routes.ts
```diff
@@ -37,6 +44,155 @@ async function checkRecaptchaToken(token: string): Promise<number> {
...
+    }
+    const ext = file.filename.split(".").pop();
+    if (!file.mimetype.startsWith("image/") || !ext || !/^(jpg|jpeg|png|gif)$/.test(ext)) {
+      reply.status(400).send({ error: "Invalid file" });
+      return;
+    }
+    const pictureName = `profile-${uuid}.${ext}`;
+    const userDirectory = path.join(assetconfigs.uploadDirectory, uuid, "profile");
+    const filePath = path.join(userDirectory, pictureName);
+    await mkdirp(userDirectory);
```
paologum, line 165: try catch?

app/src/components/TSCAccountMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+        <MenuItem onClick={() => navigate("/profile")}>
+          <Avatar /> Profile
+        </MenuItem>
+        <Divider />
+        <MenuItem onClick={() => navigate("/signup")}>
+          <ListItemIcon>
+            <PersonAdd fontSize="small" />
+          </ListItemIcon>
+          Add another account
+        </MenuItem>
```
paologum, line 95: since i don't think there's support for adding another account, do they get signed out of their current account if they attempt to login to two different accounts?

server/src/index.ts
```diff
@@ -237,10 +252,14 @@ server.post("/auth/logout", moderateRateLimit, async (request, reply) => {
 });
 server.post("/auth/verify", strictRateLimit, loginRoutes.verifyEmail);
 server.post("/auth/resend", moderateRateLimit, loginRoutes.resendVerificationEmail);
-server.post("/auth/send-resetpassword-email", strictRateLimit, loginRoutes.sendResetPasswordEmail);
-server.post("/auth/reset-password", strictRateLimit, loginRoutes.resetPassword);
-server.post("/auth/reset-email", strictRateLimit, loginRoutes.resetEmail);
+server.post("/auth/send-forgotpassword-email", strictRateLimit, loginRoutes.sendForgotPasswordEmail);
```
paologum, line 255: lowk this should be `forgot-password` with a dash it looks kinda weird

paologum (changes requested):
also one thing, instead of a large file upload button, can you make the profile picture and icon button or however with like a pencil in the bottom right. like the picture below but in the bottom right but have both the pro pic and the pencil be clickable/do the same thing as the button.

image

app/src/Profile.tsx
asivath: this ties in with the next comment but for some reason the observer is not reacting to changes in pictureUrl after uploading. i tried just using state before but that doesn't work either. after uploading the only way to get it to change was to refresh the page or keep track of the url in local state.

app/src/Profile.tsx
asivath: above

app/src/components/TSCAccountMenu.tsx
asivath: it helps people with disabilities that use screen readers. it provides screen readers context about the website

app/src/Profile.tsx
asivath: a normal user will not come across 403 since I hide the change password section to google users. should we allow google users to change their password here because then they would have no current password to enter.

server/src/login-routes.ts
asivath: its surrounded by a try catch

app/src/components/TSCAccountMenu.tsx
asivath: that will make them go through the email link thing where they will get assigned a new session.

app/src/Profile.tsx
paologum: can you try making the onChange handler an action

app/src/components/TSCAccountMenu.tsx
paologum: what email link thing, in your demo it looked like they were still signed in on the sign in page?

app/src/Profile.tsx
paologum: hmm ok yeah that's fine

app/src/components/TSCAccountMenu.tsx
asivath: i didn't show this, if they were to try to add an account they would be going through user/pass verification which would send them an email which when they clicked on it would set them a new session. i'm still logged in because I did not actually go through that entire flow.

app/src/Profile.tsx
asivath: the issue was that the url was never changing so the browser was caching the request for the image. the solution was to add a random parameter to the url to force it to fetch the new content.

paologum:
that pencil is so big LMAO, can you make it smaller? maybe like 1/2 the size? send me a pic before you push and i'll lyk

app/src/Profile.tsx
paologum: that is crazy, good find

app/src/components/TSCAccountMenu.tsx
paologum: oh ok so they can't go back and forth between emails like google does right, this is just a way of skipping essentially signing out

asivath:
https://github.com/earthhistoryviz/tsconline/assets/94872897/1d42bd50-e326-4519-ab48-c1985747fd65

app/src/components/TSCAccountMenu.tsx
asivath: yes

app/src/Profile.tsx
```diff
@@ -0,0 +1,472 @@
...
+              <EditIcon className={`pencil${isHovered ? "-hover" : ""}`} fontSize="small" />
+            </Avatar>
+          }
+          onClick={handleBadgeClick}>
+          {state.user.pictureUrl ? (
+            <Avatar
+              src={state.user.pictureUrl}
+              className="editable-profile-picture"
+              onMouseEnter={() => setIsHovered(true)}
+              onMouseLeave={() => setIsHovered(false)}
```
paologum, line 262: instead of `pencil-hover`
```
.editable-profile-picture:hover + .MuiBadge-badge .badge .pencil {
  color: #188de6;
}
```
`+` is same level and space between css selecters mean direct descendents

paologum (changes requested):

paologum (approved):

PR #119, opened 2024-05-28
added function to apply row order

jay-jhk-lee:
loading settings file now applies the correct row order

https://github.com/earthhistoryviz/tsconline/assets/89719669/e80c5ca7-20bb-4ee1-9896-4493bbfa904d


app/src/state/actions/column-actions.ts
```diff
@@ -106,6 +106,49 @@ export const applyChartColumnSettings = action("applyChartColumnSettings", (sett
...
+      }
+      //switch the two columns
+      [column.children[columnIndex], column.children[switchIndex]] = [
+        column.children[switchIndex],
+        column.children[columnIndex]
+      ];
+    }
+    applyRowOrder(column.children[columnIndex], settingsChild);
+    columnIndex++;
+  }
```
paologum, line 149: should add a check that `columnIndex == column.children.length` for robustness

app/src/state/actions/column-actions.ts
```diff
@@ -106,6 +106,49 @@ export const applyChartColumnSettings = action("applyChartColumnSettings", (sett
...
+      "Chart Title in " + currName !== column.children[columnIndex].name
+    ) {
+      //find the index of the correct column
+      let switchIndex = columnIndex;
+      for (let j = columnIndex; j < column.children.length; j++) {
+        if (currName === column.children[j].name || "Chart Title in " + currName === column.children[j].name) {
+          switchIndex = j;
+          break;
+        }
+      }
```
paologum, line 140: for readability, i think `indexOf` would make this easier to read

also add a check for index value `-1` for robustness

app/src/state/actions/column-actions.ts
```diff
@@ -106,6 +106,49 @@ export const applyChartColumnSettings = action("applyChartColumnSettings", (sett
...
+      currName = extractName(settings._id) + " " + currName;
+    }
+    //column isn't in the loaded datapack(s) so skip
+    if (!state.settingsTabs.columnHashMap.get(currName)) {
+      continue;
+    }
+    //current index doesn't have the column specified by the settings file
+    if (
+      currName !== column.children[columnIndex].name ||
+      "Chart Title in " + currName !== column.children[columnIndex].name
```
paologum, line 131: since you check if the hash map has `currName`, the check for `"Chart Title in " + currName !== column.children[columnIndex].name` will never be hit

instead you should normalize `currName` earlier if the parent is `Chart Root` and the type of column is `RootColumn` and the unit is not `Ma` then it should append `Chart Title in` to the beginning.

also since we are using it a lot of places now, it might be better to localize `Chart Title in ` to a constant then apply it everywhere we use it for easier readability and change. Call it maybe `altUnitNamePrefix`

paologum (changes requested):

app/src/state/actions/column-actions.ts
```diff
@@ -106,6 +107,54 @@ export const applyChartColumnSettings = action("applyChartColumnSettings", (sett
...
+    //current index doesn't have the column specified by the settings file
+    if (childName !== column.children[columnIndex].name) {
+      //find the index of the column that matches
+      let indexOfMatch = -1;
+      for (let j = columnIndex; j < column.children.length; j++) {
+        if (childName === column.children[j].name) {
+          indexOfMatch = j;
+          break;
+        }
+      }
```
paologum, line 145: oops i meant this line can be `column.children.findIndex(child => child.name == childName)`

app/src/state/actions/column-actions.ts
paologum: do we want to increment the `columnIndex` if `indexOfMatch !== -1`?

jay-jhk-lee: I've made some changes so it should work now

paologum (approved):

PR #121, opened 2024-05-30
Tsc 373 fix download of datapack error

JacquiFreeElectron:
fix the problem of uuid being undefined when requesting a download. Also, no file will be downloaded to user's computer if not logged in

https://github.com/earthhistoryviz/tsconline/assets/104879321/e0491008-a2ac-4725-a496-f149a83bff8f

### DIFF
--- a/app/src/settings_tabs/Datapack.tsx
+++ b/app/src/settings_tabs/Datapack.tsx
@@ -62,8 +62,8 @@ export const Datapacks = observer(function Datapacks() {
 
   const handleDownload = async (needEncryption: boolean, fileName: string) => {
     const fileURL = await getFileURL(needEncryption, fileName);
-    const aTag = document.createElement("a");
     if (fileURL && typeof fileURL === "string") {
+      const aTag = document.createElement("a");
       aTag.href = fileURL;
 
       aTag.setAttribute("download", fileName);
--- a/app/src/settings_tabs/Datapack.tsx
+++ b/app/src/settings_tabs/Datapack.tsx
@@ -44,18 +44,21 @@ export const Datapacks = observer(function Datapacks() {
   };
   async function getFileURL(needEncryption: boolean, fileName: string) {
     const fileBlob = await actions.requestDownload(fileName, needEncryption);
-    try {
-      const reader = new FileReader();
-      reader.readAsDataURL(fileBlob);
-      await new Promise((resolve, reject) => {
-        reader.onloadend = resolve;
-        reader.onerror = reject;
-      });
-      return reader.result;
-    } catch (error) {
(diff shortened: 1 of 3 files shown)


app/src/state/actions/general-actions.ts
```diff
@@ -733,7 +734,12 @@ export const requestDownload = action(async (filename: string, needEncryption: b
         );
         break;
       }
+      case 401: {
+        displayServerError(response, ErrorCodes.NOT_LOGGED_IN, ErrorMessages[ErrorCodes.NOT_LOGGED_IN]);
+        break;
+      }
```
paologum, line 740: for less dupe code, and also for a default value to be in place, you should have the cases change the error code, and at the end of the `if`, `displayServerError` and set a default value initially, that way unknown status codes will still print a value

app/src/settings_tabs/Datapack.tsx
```diff
@@ -44,17 +44,19 @@ export const Datapacks = observer(function Datapacks() {
...
+        await new Promise((resolve, reject) => {
+          reader.onloadend = resolve;
+          reader.onerror = reject;
+        });
+        return reader.result;
+      } catch (error) {
+        actions.pushError(ErrorCodes.INVALID_PATH);
+        return "";
+      }
     }
```
paologum, line 60: if `getFileURL` returns null or empty, `handleDownload` creates an `aTag` that never gets removed, so you should either always remove it or only create it if `fileURL` exists

paologum (changes requested):

app/src/settings_tabs/Datapack.tsx
JacquiFreeElectron: I think `handleDownload `only create `aTag` when `fileURL` exists bc I checked `if (fileURL && typeof fileURL === "string") `

app/src/state/actions/general-actions.ts
```diff
@@ -709,31 +709,31 @@ export const requestDownload = action(async (filename: string, needEncryption: b
...
-          ErrorCodes.INCORRECT_ENCRYPTION_HEADER,
-          ErrorMessages[ErrorCodes.INCORRECT_ENCRYPTION_HEADER]
-        );
+      case 401: {
+        errorCode = ErrorCodes.NOT_LOGGED_IN;
+        errorMessage = ErrorMessages[ErrorCodes.NOT_LOGGED_IN];
         break;
       }
     }
+    displayServerError(response, errorCode, errorMessage);
```
paologum, line 735: errormessage is just `ErrorMessages[errorCode]` so you don't need to set it each time, you can just set `errorCode` then abstract it here

paologum (approved):

PR #122, opened 2024-05-31
TSC-353 Done with cookie notifications and preventing logins if rejected

Changk820:
image
Once accepted or rejected, will disappear from the bottom of Home.

image
Will always be there regardless of accept/reject on the sign in page.

image
Can't sign in if rejected or unset.

image
Can sign in when accepted, but this account does not exist. Maybe should change this error?


Changk820:
Need to make modifications as per 5/31 meeting

PR #123, opened 2024-06-02
added tests for parsing columns

jay-jhk-lee:
added tests for font change, basic, event, point columns, and a comprehensive settings file.


paologum:
no test file

app/__tests__/parse-settings.test.ts
```diff
@@ -0,0 +1,630 @@
...
+  });
+  it("should parse font change", async () => {
+    const settings = columnInfoTSCToXml(translateColumnInfoToColumnInfoTSC(inputs["font-change-test"]), "    ");
+    expect(settings).toEqual(readFileSync("app/__tests__/__data__/font-change-key.tsc").toString());
+  });
+  it("should parse all columns and settings", async () => {
+    const settings = jsonToXml(inputs["all-column-test"], inputs["default-settings"]);
+    expect(settings).toEqual(readFileSync("app/__tests__/__data__/all-column-key.tsc").toString());
+  });
+});
```
paologum, line 630: the way this should be testing is below, that way errors can be caught easily and found easily in the correct function. This tests the whole function and doesn't go with a unit testing approach.

make keys and test `columnInfoTSCToXml`, `extractColumnType` `generateFontsXml`, `translateColumnInfoToColumnInfoTSC`, `escapeHtmlChars` and `generateSettingsXml`.

you should also reduce a lot of the test size by mocking things simpler. for example `jsonToXml` is really simple. you don't need to use `jsonToXml` as an overarching test. you can just mock `generateSettingsXml` and `columnInfoTSCToXml` to return `""` so that you can just test that function. This means tests are a lot easier to read and are more unit testy. In the same area, you would use `generateSettingsXml` to see the settings only, and any other calls to this from other tests shouldn't need to worry about whether or not `generateSettingsXml` works

paologum (changes requested):

app/__tests__/parse-settings.test.ts
```diff
@@ -0,0 +1,710 @@
...
+  const mock1 = jest.spyOn(parseSettings, "generateSettingsXml");
+  const mock2 = jest.spyOn(parseSettings, "columnInfoTSCToXml");
+  it("should generate prolog and root of xml", async () => {
+    mock1.mockReturnValue("");
+    mock2.mockReturnValue("");
+    expect(parseSettings.jsonToXml(tests["generate-prolog-and-root-xml-test"], tests["default-settings-test"])).toEqual(
+      readFileSync("app/__tests__/__data__/generate-prolog-and-root-xml-key.tsc").toString()
+    );
+  });
+});
```
paologum, line 710: you probably shouldn't have the key in a seperate file like above stated. also can't you use a previous `ColumnInfo` test instead of making a new one that technically isn't tested? also what does prolog mean

shared/src/constants.ts
```diff
@@ -282,7 +282,7 @@ export const defaultPointColumnInfoTSC: PointColumnInfoTSC = {
     g: 233,
     b: 191
   },
-  doNotSetWindowAuto: false,
+  doNotSetWindowAuto: true,
```
paologum, line 285: i set this to always false because i do the logic myself and it wouldn't be good if the frontend changes from what is displayed without the user knowing why

paologum (changes requested):

app/__tests__/parse-settings.test.ts
```diff
@@ -0,0 +1,710 @@
...
+      },
+      "Range Box Label": {
+        on: false,
+        bold: false,
+        color: "rgb(0, 0, 0)",
+        fontFace: "Arial",
+        inheritable: false,
+        italic: true,
+        size: 14
+      }
```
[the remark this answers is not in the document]
jay-jhk-lee: I tried instantiating it outside as a const which resulted in "ReferenceError: Cannot access 'fontsInfo' before initialization" and putting it inside the mock statement which resulted in "Cannot find name 'fontsInfo'" but both didn't work, so I had to put it for each variable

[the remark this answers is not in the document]
jay-jhk-lee: and i can't import it from shared since it also throws the cannot access before initialization error

app/__tests__/parse-settings.test.ts
```diff
@@ -0,0 +1,710 @@
...
+    expect(parseSettings.extractColumnType("class datastore.RootColumn:Chart Root")).toEqual("RootColumn");
+  });
+  test.each([
+    ["", undefined],
+    ["RootColumn", undefined],
+    ["class datastoreRootColumn:Chart Root", undefined],
+    ["class datastore.RootColumn Chart Root", undefined]
+  ])("should return undefined for incorrectly formatted id", (input, expected) => {
+    expect(parseSettings.extractColumnType(input)).toEqual(expected);
+  });
```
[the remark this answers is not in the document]
jay-jhk-lee: > you can add the `it` to the `test.each` ?

I've tried it and it throws an error: `Tests cannot be nested. Test "" cannot run because it is nested within "should return undefined for incorrectly formatted id".`

app/__tests__/parse-settings.test.ts
jay-jhk-lee: prolog is the very top line of the xml with the xml version and encoding

paologum (approved):

PR #124, opened 2024-06-03
Tsc 353 ask for cookies

Changk820:
TSC-355
~~Demo: https://www.youtube.com/watch?v=6aEuoBVOVwo~~
New Demo:  https://youtu.be/vgWTW4cLy0g


paologum:
can you just disable the buttons instead of removing them? if you do so that way you shouldn't need to refresh the page if you use `useState`. I also haven't read the code yet but this is just from the demo

app/src/CookieConsent.tsx
```diff
@@ -0,0 +1,82 @@
...
+  // localStorage.removeItem("cookieConsent");
+
+  useEffect(() => {
+    const savedConsent = localStorage.getItem("cookieConsent");
+    if (!savedConsent) {
+      setIsVisible(true);
+    } else if (persistent && savedConsent !== "accepted") {
+      setIsVisible(true);
+    }
+  }, [persistent]);
```
paologum, line 37: i'm not sure if we need a `useEffect` hook that updates on `persistent`
what's wrong with just having these two statements and settings
to my knowledge, `isVisible` is the refresher, and `persistent` does not need to enable refreshes
```
  const savedConsent = localStorage.getItem("cookieConsent");
  const shouldShowConsent = !savedConsent || (persistent && savedConsent !== "accepted");
  const [isVisible, setIsVisible] = useState(shouldShowConsent);
```

app/src/CookieConsent.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const handleAccept = () => {
+    localStorage.setItem("cookieConsent", "accepted");
+    setIsVisible(false);
+    window.location.reload(); // Refresh the page after accepting cookies
```
paologum, line 42: i don't believe this is necessary, could you explain why you need to reload?

app/src/CookieConsent.tsx
```diff
@@ -0,0 +1,82 @@
...
+  bottom: 0,
+  left: 0,
+  right: 0,
+  background: theme.palette.background.paper, // Use the background color from the theme
+  padding: theme.spacing(2),
+  boxShadow: theme.shadows[3],
+  display: "flex",
+  justifyContent: "space-between",
+  alignItems: "center",
+  zIndex: 1000 // Ensure it's above everything else
```
paologum, line 16: verify this goes over map points

app/src/Login.tsx
```diff
@@ -155,9 +163,11 @@ export const Login: React.FC = observer(() => {
               id="password"
               autoComplete="current-password"
             />
-            <TSCButton type="submit" fullWidth variant="contained" sx={{ mt: 3, mb: 2 }} endIcon={<LoginIcon />}>
-              Sign In
-            </TSCButton>
+            {cookieConsent === "accepted" ? (
+              <TSCButton type="submit" fullWidth variant="contained" sx={{ mt: 3, mb: 2 }} endIcon={<LoginIcon />}>
+                Sign In
+              </TSCButton>
```
paologum, line 169: you should probs just disable instead of not showing it

app/src/Login.tsx
```diff
@@ -179,17 +189,20 @@ export const Login: React.FC = observer(() => {
...
-            />
+            {cookieConsent === "accepted" ? (
+              <GoogleLogin
+                onSuccess={async (credentialResponse) => {
+                  const credential = credentialResponse.credential;
+                  await handleLogin(true, { credential, recaptchaToken: null });
+                }}
+                onError={() => actions.pushError(ErrorCodes.UNABLE_TO_LOGIN_SERVER)}
+                width="400px"
+              />
```
paologum, line 200: same here if possible

paologum: also for future reference if you only want to load based on a condition you can do
```
cookieConsent === "accepted" && <Component/>
```

app/src/CookieConsent.tsx
```diff
@@ -0,0 +1,82 @@
...
+}));
+
+interface CookieConsentProps {
+  persistent?: boolean;
+}
+
+const CookieConsent: React.FC<CookieConsentProps> = ({ persistent = false }) => {
```
paologum, line 23: is persistent needed anymore?

paologum (changes requested):

app/src/CookieConsent.tsx
Changk820: Aaron wanted the sign in button to be removed if the cookies are not accepted, but after accepting, I had to refresh the page to see the buttons show up again.

app/src/CookieConsent.tsx
Changk820: I thought we were only displaying this while you were on the homepage/signin? So it doesn't show up at same time asa map point

app/src/CookieConsent.tsx
Changk820: You're right

app/src/Login.tsx
Changk820: made the changes

app/src/Login.tsx
Changk820: apparently GoogleLogin doesn't have a disable component, but did it with CSS

app/src/CookieConsent.tsx
Changk820: I think so.
isVisible makes it show or not show
Persistent is to differentiate between being on the home page and being on the sign in page(persistent)

Changk820:
New video
https://youtu.be/vgWTW4cLy0g
<sup><sub>~~Like, subscribe and hit that notification button~~<sub><sup>

app/src/CookieConsent.tsx
paologum: ok i guess that's fine then

app/src/CookieConsent.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [isVisible, setIsVisible] = useState(shouldShowConsent);
+  const theme = useTheme();
+
+  // uncomment the following line to see the cookie consent banner
+  // localStorage.removeItem("cookieConsent");
+
+  const handleAccept = () => {
+    localStorage.setItem("cookieConsent", "accepted");
+    setIsVisible(false);
+    if (persistent) window.location.reload(); // Refresh the page after accepting cookies so sign in shows
```
paologum, line 36: we don't want to do `window.location.reload()` because it refreshes components that don't need to be refreshed. instead we want to hold the cookie acceptance policy in the state. if cookie acceptance policy is accepted, enable the `sign-in` if not disable all `sign-in`. on `initialize` we should then set that state variable to the `localStorage` keys value

app/src/Login.tsx
```diff
@@ -123,7 +131,7 @@ export const Login: React.FC = observer(() => {
     };
     await handleLogin(false, formData);
   };
-
+  const cookieConsent = localStorage.getItem("cookieConsent");
```
paologum, line 134: from my previous comment, this would be a part of the large state object and would auto refresh the components that use it

app/src/state/state.ts
```diff
@@ -19,6 +19,7 @@ import { settings } from "../constants";
 
 export type State = {
   loadSaveFilename: string;
+  cookieConsent: boolean | null;
```
paologum, line 22: this should probably just be `boolean` and auto set to `false`

app/src/state/actions/general-actions.ts
```diff
@@ -757,13 +757,14 @@ export const logout = action("logout", async () => {
...
 export const sessionCheck = action("sessionCheck", async () => {
+  state.cookieConsent = localStorage.getItem("cookieConsent") === "true";
   try {
     const response = await fetcher("/auth/session-check", {
       method: "POST",
       credentials: "include"
     });
     const data = await response.json();
-    if (data.authenticated) {
+    if (data.authenticated && state.cookieConsent) {
```
paologum, line 767: you don't need to have consent to check the session so this should stay what it was

app/src/Login.css
```diff
@@ -28,3 +28,7 @@
 .account-form {
   width: 100%;
 }
+.disabled {
```
paologum, line 31: call this a more specific name like `disabled-google-login` just so no overlap in css occurs elsewhere

app/src/CookieConsent.tsx
```diff
@@ -0,0 +1,81 @@
...
+  zIndex: 1000 // Ensure it's above everything else
+}));
+
+interface CookieConsentProps {
+  persistent?: boolean;
+}
+
+const CookieConsent: React.FC<CookieConsentProps> = ({ persistent = false }) => {
+  const savedConsent = localStorage.getItem("cookieConsent");
+  const shouldShowConsent = !savedConsent || (persistent && savedConsent !== "true");
```
paologum, line 27: these should change to what you have now in state and `CookieConsent` should be wrapped in an `observer`

app/src/state/actions/general-actions.ts
```diff
@@ -962,3 +963,6 @@ export const setEnableHideBlockLabel = action("setEnableHideBlockLabel", (newval
 export const setLoadSaveFilename = action("setLoadSaveFilename", (newval: string) => {
   state.loadSaveFilename = newval;
 });
+export const setCookies = action("setCookies", (newval: boolean) => {
+  state.cookieConsent = newval;
+});
```
paologum, line 968: like aaron said, this should also set `localStorage` so that whenever you change it, it changes both vars to be the same value

paologum (changes requested):

app/src/state/actions/general-actions.ts
Changk820: But what if say use accepted the cookies, logged in, then signed in and cleared localstorage/cookies? they would be put in a  weird place, where they are signed in, but cookies are technically not accepted, so I feel like this should be here.

app/src/state/state.ts
Changk820: I have to have it null because if I initialize it as false, the home screen cookie notification would go away because if any action is taken, it disappears. Being set as false means action was taken.

app/src/Login.css
Changk820: okay

app/src/state/state.ts
paologum: i see i see nvm then

app/src/state/actions/general-actions.ts
paologum: if they cleared local storage/cookies then they won't have a valid session anymore and shouldn't be signed in on reload

app/src/CookieConsent.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { useState, useContext } from "react";
+import { observer } from "mobx-react-lite";
+import { useTheme, Button, Typography, Box } from "@mui/material";
+import styled from "@mui/material/styles/styled";
+import Slide from "@mui/material/Slide";
+import { context, state } from "./state";
```
paologum, line 6: don't import `state` from `./state`,
do `const { state, actions } = useContext(context)`

app/src/Login.tsx
```diff
@@ -154,8 +162,15 @@ export const Login: React.FC = observer(() => {
...
+              disabled={state.cookieConsent === false}
             />
-            <TSCButton type="submit" fullWidth variant="contained" sx={{ mt: 3, mb: 2 }} endIcon={<LoginIcon />}>
+            <TSCButton
+              type="submit"
+              fullWidth
+              variant="contained"
+              sx={{ mt: 3, mb: 2 }}
+              endIcon={<LoginIcon />}
+              disabled={state.cookieConsent === false}>
```
paologum, line 173: shouldn't this just be `!state.cookieConesnt` because if `state.cookieConsent = null` it won't be disabled.

applies to line 165 and 154 as well

app/src/Login.tsx
```diff
@@ -179,17 +194,21 @@ export const Login: React.FC = observer(() => {
...
-            <GoogleLogin
-              onSuccess={async (credentialResponse) => {
-                const credential = credentialResponse.credential;
-                await handleLogin(true, { credential, recaptchaToken: null });
-              }}
-              onError={() => actions.pushError(ErrorCodes.UNABLE_TO_LOGIN_SERVER)}
-              width="400px"
-            />
+            {/* GoogleLogin does not have a "disable" property */}
+            <div className={state.cookieConsent === false ? "disabled-google-login" : ""}>
```
paologum, line 198: same here

app/src/state/actions/general-actions.ts
```diff
@@ -757,6 +757,7 @@ export const logout = action("logout", async () => {
 });
 
 export const sessionCheck = action("sessionCheck", async () => {
+  state.cookieConsent = localStorage.getItem("cookieConsent") === "true";
```
paologum, line 760: this should check if it exists and set the boolean to the corresponding value, but if it doesn't exist then the `state.cookieConsent` should be `null`

paologum (changes requested):

paologum (approved):
great!

PR #126, opened 2024-06-12
Tsc 348 red guideline

jay-jhk-lee:
Added a red timeline to the chart.
Clicking on chart will lock the timeline vertically; the timeline labels can move horizontally.
if chart is generated with popups, then turning on the timeline will disable popups and turning off timeline will enable popups.

https://github.com/earthhistoryviz/tsconline/assets/89719669/447cb929-19fb-4ed2-8675-a81cdf066376


paologum (approved):
looks great!

PR #129, opened 2024-06-14
Tsc 393 fix min max problems with datamining auto scaling

JacquiFreeElectron:


app/src/state/actions/column-actions.ts
```diff
@@ -321,7 +321,12 @@ export const addDataMiningColumn = action((column: ColumnInfo, type: EventFreque
       console.log("WARNING: unknown column display type", column.columnDisplayType);
       return;
   }
-  const { min, max } = findRangeOfWindowStats(windowStats);
+  const units = Object.keys(state.settings.timeSettings)[0];
+  const { min, max } = findRangeOfWindowStats(
+    windowStats,
+    state.settings.timeSettings[units].topStageAge,
+    state.settings.timeSettings[units].baseStageAge
+  );
```
paologum, line 329: you're assuming here the column is in the unit at index 0 but what if it's at 1,2,etc

paologum (changes requested):
have you verified this auto scales the same as the java program? should be the same number

JacquiFreeElectron:
Now the datamining auto scale range is the same as the jar, expect for a very small different due to round up/ precision difference (and I verified the way jar calculate the range is most accurate for drawing the chart so I used their way tho it doesn't look very elegant)

Combined:
combined
FAD:
FAD
LAD:
LAD
Freq:
Freq
Max:
max
Min:
min
Avg:
Avg
Combined Event:
RoC:
RoC

app/src/state/actions/column-actions.ts
```diff
@@ -350,20 +364,27 @@ export const addDataMiningColumn = action(
...
         if (!isDataMiningPointDataType(type)) {
           console.log("WARNING: unknown data mining type associated with a point", type);
           return;
         }
+        const pointData = column.subInfo.map((subPoint) => {
+          return { age: subPoint.age, value: subPoint.xVal };
+        });
+        pointData.shift();
+        pointData.pop();
         windowStats = computeWindowStatisticsForDataPoints(
```
paologum, line 381: just wondering why you don't filter the age if it's lower or higher than the `baseStageAge` and `topStageAge`

also why do you `shift` and `pop`

app/src/state/actions/column-actions.ts
```diff
@@ -417,14 +440,23 @@ export const addDataMiningColumn = action(
         windowStats = computeWindowStatistics(
           column.subInfo.map((subChron) => subChron.age),
           column.columnSpecificSettings.windowSize,
+          column.columnSpecificSettings.stepSize,
```
paologum, line 443: why no filter here either

paologum (commented):
some comments but overall great

also add a picture of the `chron` frequency being the same as well

app/src/state/actions/column-actions.ts
```diff
@@ -309,24 +313,34 @@ export const addDataMiningColumn = action(
...
+            if (type === "Combined Events") return true;
+            else return subEvent.subEventType === type;
+          })
+          .map((subEvent) => subEvent.age)
+          .filter((age) => {
+            return (
+              age >= state.settings.timeSettings[column.units].topStageAge &&
+              age <= state.settings.timeSettings[column.units].baseStageAge
+            );
+          });
```
paologum, line 336: i'm also slightly confused as to why you filter here if we find the correct window later on

app/src/state/actions/column-actions.ts
JacquiFreeElectron: yeah it's just the way jar used. if I use filter in here, the result will be a lot different, and some case will even cause the jar to throw an error

app/src/state/actions/column-actions.ts
JacquiFreeElectron: same reason

JacquiFreeElectron: actually for this one I need to double check, because when I was rewriting the functions, I didn't rebase, so I haven't really checked the situation for `chron` yet

app/src/state/actions/column-actions.ts
JacquiFreeElectron: if I don't filter here it , the range will be slightly different than the jar  I think it's bc when we're using the mid point to find the correct window it's not that accurate. Somehow, for events, they want it to be more accurate, so they give the function filtered data, while for other data points, they just give the function unfiltered data.

app/src/state/actions/column-actions.ts
JacquiFreeElectron: and the reason for using `shift `and `pop `is the same as ^

app/src/state/actions/column-actions.ts
paologum: interesting but alright, can you add a comment in the code saying that for future reference if anyone else visits this

app/src/state/actions/column-actions.ts
paologum: ok i would double check that before i review again then

app/src/state/actions/column-actions.ts
JacquiFreeElectron: sure

JacquiFreeElectron:
chron is the same as the jar now: 
Screenshot 2024-07-02 at 5 55 25 PM

app/src/state/actions/column-actions.ts
```diff
@@ -437,8 +439,19 @@ export const addDataMiningColumn = action(
...
+          .map((subChron) => subChron.age)
+          .filter((age) => {
+            return (
+              age >= state.settings.timeSettings[column.units].topStageAge &&
+              age <= state.settings.timeSettings[column.units].baseStageAge
+            );
+          });
+        chronData.shift();
+        chronData.pop();
         windowStats = computeWindowStatistics(
```
paologum, line 453: your comment says we don't filter data, but you are filtering?

JacquiFreeElectron: yeah actually I shouldn't filter

paologum (approved):
i'm assuming with the non-filter you changed/checked recently that chron is still the same and if not check that real quick before you merge

PR #130, opened 2024-06-18
Tsc 381 add tests for google login login and signup

asivath:
Migrated to vitest and started login route tests.


app/package.json
```diff
@@ -41,7 +41,8 @@
     "@types/react-dom": "^18.0.10",
     "@vitejs/plugin-react": "^4.2.1",
     "typescript": "5.2.2",
-    "vite": "^5.2.2"
+    "vite": "^5.2.2",
+    "vitest": "^1.6.0"
```
paologum, line 45: do the root packages not get inherited?

package.json
```diff
@@ -33,9 +33,9 @@
     "prettier-format": "prettier --write \"**/*.{ts,tsx,css}\"",
     "setup": "yarn && yarn install && yarn build",
     "start": "yarn workspace @tsconline/server start",
-    "test": "jest -t",
-    "test-all": "jest",
-    "test-clear-cache": "jest --clearCache",
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "coverage": "vitest run --coverage",
     "clean": "yarn workspaces foreach -A --parallel run clean"
```
paologum, line 39: does `vitest run` have functionality to accept test names like jest did so you don't have to run all the tests?

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+    it("should return 422 if recaptcha fails", async () => {
+      const spy = vi.spyOn(verifyModule, "checkRecaptchaToken").mockResolvedValue(0.0);
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/signup",
+        payload: payload
+      });
+
+      spy.mockRestore();
```
paologum, line 147: does spy carry over ? isn't the scope only in this `it`?

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+      vi.mocked(databaseModule.findUser).mockResolvedValueOnce([testUser]);
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/signup",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(200);
+      expect(response.json().message).toBe("Email sent");
```
paologum, line 178: should add checks that `sendEmail`, `createVerification` `createUser` and `findUser` are called once

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+});
+
+describe("login-routes tests", () => {
+  const originalEnv = { EMAIL_USER: "testuser", EMAIL_PASS: "testpass" };
+
+  beforeEach(() => {
+    process.env = { ...originalEnv };
+  });
+
+  describe("/signup", () => {
```
paologum, line 106: should add a test for any missing field recaptchaToken and non-matching email regex email

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+        url: "/oauth",
+        payload: {
+          credential: "",
+          recaptchaToken: "test"
+        }
+      });
+
+      expect(response.statusCode).toBe(400);
+      expect(response.json().error).toBe("Invalid form");
+    });
```
paologum, line 199: can a recaptchaToken be 0 btw? if it can then a token of 0 might return invalid form, weird case that i don't think will happen though

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+        method: "POST",
+        url: "/oauth",
+        payload: payload
+      });
+
+      spy.mockRestore();
+
+      expect(response.statusCode).toBe(422);
+      expect(response.json().error).toBe("Recaptcha failed");
+    });
```
paologum, line 214: i think to be robust you should also check to make sure that `checkRecaptchaToken` gets called once

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+      const response = await app.inject({
+        method: "POST",
+        url: "/signup",
+        payload: payload
+      });
+
+      spy.mockRestore();
+
+      expect(response.statusCode).toBe(422);
+      expect(response.json().error).toBe("Recaptcha failed");
```
paologum, line 150: above

server/__tests__/login-routes.test.ts
paologum: also super super robust but you should mock a situation where the database doesn't insert the user so that we know this function handles it correctly and throws an error and returns the proper status code

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+        })
+      }));
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/oauth",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(400);
```
paologum, line 230: also check message

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+    it("should return 409 is user/pass user exists", async () => {
+      vi.mocked(databaseModule.findUser).mockResolvedValueOnce([testUser]);
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/oauth",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(409);
```
paologum, line 242: msg

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+      const response = await app.inject({
+        method: "POST",
+        url: "/oauth",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(400);
+    });
+
+    it("should return 409 is user/pass user exists", async () => {
```
paologum, line 233: if ** right

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+      const response = await app.inject({
+        method: "POST",
+        url: "/oauth",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(409);
+    });
+
+    it("should return 200 if successful is google users exists", async () => {
```
paologum, line 245: kinda confused with the test message, can you make it more clear

paologum, line 245: also should add a test for if `userRow` isn't found then we create the entry in the database. make sure to look for `createUser` being called once and checking session for that. technically you might have to add the same code from `signup` to where the database doesn't insert because that code doesn't exist yet. similar test to the above comments also applies here for mocking the situation where the database doesn't insert correctly

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+
+    it("should return 403 if email is not verified", async () => {
+      vi.mocked(databaseModule.findUser).mockResolvedValueOnce([{ ...testUser, emailVerified: 0 }]);
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/login",
+        payload: payload
+      });
+
```
paologum, line 343: i would add the check just to be explicit since we aren't intentionally mocking here for it to make sure we call `compare` method once

server/__tests__/login-routes.test.ts
```diff
@@ -0,0 +1,375 @@
...
+      vi.mocked(databaseModule.findUser).mockResolvedValueOnce([{ ...testUser, invalidateSession: 1 }]);
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/login",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(423);
+      expect(response.json().error).toBe("Account locked");
```
paologum, line 358: above

server/__tests__/login-routes.test.ts
paologum: also for this and the other failures, i think it might be good to make sure the session isn't set properly or that `checkSession` fails

server/src/login-routes.ts
```diff
@@ -742,7 +720,7 @@ export const login = async function login(
       return;
     }
     const { uuid, hashedPassword, emailVerified, invalidateSession } = userRow;
-    if (hashedPassword && (await compare(password, hashedPassword))) {
+    if (hashedPassword && (await compare(password, hashedPassword ?? ""))) {
```
paologum, line 723: do you need the `??`, shouldn't the types be happy if you checked for `hashedPassword` alr

vite.config.ts
```diff
@@ -0,0 +1,18 @@
+import { fileURLToPath, URL } from "url";
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    cache: false,
+    environment: "node",
+    include: ["server/__tests__/**.test.ts", "app/__tests__/**.test.ts"],
```
paologum, line 8: shared has tests

paologum (changes requested):

app/package.json
asivath: vitest works without this but ts can't get the types without this for some reason, there is potentially a fix with some of the workspace stuff they have but it didn't seem worth my time when this works

package.json
asivath: you just need to specify a name so `yarn test util` would run all the util tests

server/__tests__/login-routes.test.ts
asivath: yes but it should only mock 0.0 for this `it` not the other ones

server/__tests__/login-routes.test.ts
asivath: the token is actually some random string that the google api creates and gives to the frontend which then passes it to the backend where we use the token to retrieve the actual score

server/src/login-routes.ts
asivath: i forgot about this it was to make jest happy but vitest handles it better

server/__tests__/login-routes.test.ts
asivath: not sure 100% what you mean but lmk if i should add something else

package.json
paologum: can it also run like a specific test in `util`?

server/__tests__/login-routes.test.ts
paologum: wait but since we redeclare `const spy` everytime aren't they seperate `spy` instances?

server/__tests__/login-routes.test.ts
paologum:
```
    await createUser(newUser);
    const insertedUser = (await findUser({ email }))[0];
    if (!insertedUser) {
      throw new Error("User not inserted");
    }
```
^ signup route
should add code in google login that checks if the user was correctly made since this isn't in the googleLogin route
also should add a test for `googleLogin` and `signup` where the database doesn't create the correct entry (`findUser` returns `[]`)

then to just add an expects `findUser` and `createUser` are called once

paologum: ah ok i saw you added the fix, but for google user you check if `createUser` throws an Error but we should also keep it consistent with the sign up code above right and `findUser` ourselves then test that way?

server/__tests__/login-routes.test.ts
asivath: the spy variable itself is different but the module we mocked stays there. getting rid of the mockRestore causes other tests to start giving out 422. but honestly I should have done this the way I did the others and just use mockResolvedValueOnce.

package.json
asivath: yea you can I'll add a command for that

server/__tests__/login-routes.test.ts
asivath: for consistency sake yeah

asivath:
on line 93 there is a commented out mock that suppresses the error messages thrown by the routes. idk if we want that because they can be kind of annoying to read through and unless u know that its supposed to throw errors it can also be unclear. I could also just use the silent mode for vitest but that would stop everything. wdyt?

paologum:
i think that you should just suppress the `console.errors` for these tests (uncomment that line) since it won't really be useful imo

paologum (approved):

PR #135, opened 2024-06-22
added tests for verifyEmail, resendVerifyEmail, sendForgotPasswordEmail

asivath:


server/__tests__/login-routes.test.ts
```diff
@@ -490,4 +517,513 @@ describe("login-routes tests", () => {
...
+      vi.mocked(databaseModule.findUser).mockResolvedValueOnce([testUser]);
+      const response = await app.inject({
+        method: "POST",
+        url: "/verify",
+        payload: payload
+      });
+
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(409);
+      expect(response.json().error).toBe("Email already verified");
```
paologum, line 563: if this happens, shouldn't it delete the verification row as well?

server/__tests__/login-routes.test.ts
```diff
@@ -490,4 +517,513 @@ describe("login-routes tests", () => {
...
+      process.env.EMAIL_PASS = "";
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/resend",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(500);
+      expect(response.json().error).toBe("Email service not configured");
```
paologum, line 687: session check

server/__tests__/login-routes.test.ts
```diff
@@ -490,4 +517,513 @@ describe("login-routes tests", () => {
...
+        expect.objectContaining({
+          to: testUser.email,
+          link: undefined,
+          buttonText: undefined,
+          message: expect.stringContaining("already been verified")
+        })
+      );
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(200);
+      expect(response.json().message).toBe("Email sent");
```
paologum, line 802: delete the verification row here like above?

server/__tests__/login-routes.test.ts
```diff
@@ -490,4 +517,513 @@ describe("login-routes tests", () => {
...
+      process.env.EMAIL_PASS = "";
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/send-forgot-password-email",
+        payload: payload
+      });
+
+      expect(response.statusCode).toBe(500);
+      expect(response.json().error).toBe("Email service not configured");
```
paologum, line 855: check session

server/__tests__/login-routes.test.ts
```diff
@@ -490,4 +517,513 @@ describe("login-routes tests", () => {
...
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/send-forgot-password-email",
+        payload: payload
+      });
+
+      expect(deleteVerificationSpy).not.toHaveBeenCalled();
+      expect(createVerificationSpy).not.toHaveBeenCalled();
+      expect(emailSpy).not.toHaveBeenCalled();
```
paologum, line 948: i feel like for these local spies watching local methods, you could create one for this `describe` and reset them in an `afterEach` method, that way you're not instantiating them every time and also can `expect.not.toHaveBeenCalled()` easily. because technically we should be consistent if we are starting `not.toHavBeenCalled` here for other tests

paologum (changes requested):

server/__tests__/login-routes.test.ts
```diff
@@ -214,7 +235,14 @@ describe("login-routes tests", () => {
         payload: payload
       });
 
-      expect(verificationSpy).toHaveBeenCalledOnce();
+      expect(verificationSpy).toHaveBeenCalledWith(
+        expect.objectContaining({
+          expiresAt: expect.any(String),
```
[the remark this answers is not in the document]
asivath: i didn't do that because I thought that verifying it's an actual date string would be verifying the date object did not fail. one thing I could to check the logic in the route is to see whether it actually added an hour or whatever to the current time.

server/__tests__/login-routes.test.ts
asivath: the reason I didn't delete in the route is so that I could tell the user that they have already verified their email. this token will eventually get cleaned up every week by the interval thing.

server/__tests__/login-routes.test.ts
asivath: same logic as above

server/__tests__/login-routes.test.ts
paologum: isn't `emailVerified` kept in the `users` table though?

asivath: yeah but this route identifies users by their tokens

server/__tests__/login-routes.test.ts
asivath: for the not.toHaveBeenCalled I wanted it specifically for the things that return 200 even though the operation was technically not successful (like no username/email existing) since it's important those functions are not called for security reasons.

server/__tests__/login-routes.test.ts
paologum: ok then i would add a check to make sure it doesn't get called here

paologum (approved):

PR #137, opened 2024-06-25
added tests for forgotPassword, changeEmail, accountRecovery

asivath:


server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+      const response = await app.inject({
+        method: "POST",
+        url: "/forgot-password",
+        payload: payload
+      });
+
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(404);
+      expect(response.json().error).toBe("Password reset token not found");
+    });
```
paologum, line 1096: i feel like you should `expect` a `findVerification` is called with the `token` passed in here

paologum: also add test for if `reason !== password)`

the code also confuses me a little
```
const verificationRow = (await findVerification({ token }))[0];
    if (!verificationRow || verificationRow.reason !== "password") {
      reply.status(404).send({ error: "Password reset token not found" });
      return;
    }
    const { userId, expiresAt, reason } = verificationRow;
    const expiresAtDate = new Date(expiresAt);
    if (expiresAtDate < new Date() || reason !== "password") {
      await deleteVerification({ token, reason: "password" });
      reply.status(401).send({ error: "Password reset token expired or invalid" });
      return;
    }
```
don't you check `reason` twice

also one of them calls `deleteVerification` while the other doesn't

server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+      const response = await app.inject({
+        method: "POST",
+        url: "/forgot-password",
+        payload: payload
+      });
+
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(500);
+      expect(response.json().error).toBe("Unknown Error");
+    });
```
paologum, line 1172: i think you `expect(findUser)` here just for clarity and `expect` `updateUser` and `deleteVerification`

server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+
+    it("should return 422 if recaptcha fails", async () => {
+      const spy = vi.spyOn(verifyModule, "checkRecaptchaToken").mockResolvedValueOnce(0.0);
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/change-email",
+        payload: payload,
+        headers: { cookie: `loginSession=${cookieHeader}` }
```
paologum, line 1253: might be better to make `header` a var

server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+      });
+
+      expect(spy).toHaveBeenCalledOnce();
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(422);
+      expect(response.json().error).toBe("Recaptcha failed");
+    });
+
+    it("should return 500 if user find fails", async () => {
+      vi.mocked(databaseModule.findUser).mockResolvedValueOnce([]);
```
paologum, line 1263: fails feels different than this test, probably `user doesn't exist`

server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+      vi.mocked(databaseModule.findUser).mockResolvedValueOnce([testUser]);
+      vi.mocked(databaseModule.deleteVerification).mockRejectedValueOnce(new Error("Database Error"));
+
+      const response = await app.inject({
+        method: "POST",
+        url: "/account-recovery",
+        payload: payload
+      });
+
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
```
paologum, line 1539: i would `expect` it to be called just to be super robust that it fails there since we don't really check an error message that was thrown

server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+        url: "/change-email",
+        payload: payload,
+        headers: { cookie: `loginSession=${cookieHeader}` }
+      });
+
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(500);
+      expect(response.json().error).toBe("Unknown Error");
+    });
+
```
paologum, line 1324: same with these

server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+        method: "POST",
+        url: "/account-recovery",
+        payload: payload
+      });
+
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(500);
+      expect(response.json().error).toBe("Unknown Error");
+    });
+
```
paologum, line 1575: ope and here

server/__tests__/login-routes.test.ts
```diff
@@ -1023,4 +1045,567 @@ describe("login-routes tests", () => {
...
+        method: "POST",
+        url: "/account-recovery",
+        payload: payload
+      });
+
+      expect(checkSession(response.headers["set-cookie"] as string)).toBe(false);
+      expect(response.statusCode).toBe(404);
+      expect(response.json().error).toBe("Token not found");
+    });
+
```
paologum, line 1492: also should have one for if the reason is wrong

paologum (changes requested):

server/__tests__/login-routes.test.ts
asivath: yeah the second one shouldn't be there

paologum (approved):

PR #143, opened 2024-06-28
Tsc 351 encryption jar tests(new branch)

JacquiFreeElectron:


.github/workflows/merge-pipeline.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           if git diff --name-only main...HEAD | grep 'file_metadata.json'; then
             echo "file_metadata.json detected, please remove" >> disallowed_files.txt
           fi   
-          git diff --name-only main...HEAD | grep -E '\.(dpk|mdpk|txt|map|java|jar|DS_Store|db)$' | grep -v '^server/__tests__/__data__/.*\.txt$' >> disallowed_files.txt || true
+          git diff --name-only main...HEAD | grep -E '\.(dpk|mdpk|txt|map|java|jar|DS_Store|db)$' | grep -v '^server/__tests__/__data__/.*\.dpk$' | grep -v '^server/__tests__/__data__/encryption-test-keys/.*\.dpk$' | grep -v '^server/__tests__/__data__/.*\.txt$' >> disallowed_files.txt || true
```
asivath, line 31: this can be simplified by using or statements in the regex

server/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,540 @@
...
+  it("should correctly encrypt an unencrypted TSCreator txt file", async () => {
+    if (!(await checkFileExists("server/__tests__/__data__/encryption-test-generated-file/encryption-test-1.txt"))) {
+      await unmockedRunJavaEncrypt(jarFilePath, "server/__tests__/__data__/encryption-test-1.txt", resultPath);
+    }
+    const resultFilePath = "server/__tests__/__data__/encryption-test-generated-file/encryption-test-1.txt";
+    const keyFilePath = "server/__tests__/__data__/encryption-test-keys/test-1-key.txt";
+    const [result, key] = await Promise.all([unmockedReadFile(resultFilePath), unmockedReadFile(keyFilePath)]);
+    expect(result).toEqual(key);
+    // Compare the byte size (length)
+    expect(result.length).toBe(key.length);
```
paologum, line 470: if you do these checks, do the length before the `expect(result).toEqual(key)` since if the length is different the first `expect` will throw and this will never be reached (for all the tests)

server/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,540 @@
...
+    if (error) {
+      console.error(`exec error: ${error}`);
+      return;
+    }
+    if (stderr) {
+      console.error(`stderr: ${stderr}`);
+    }
+    console.log(`stdout: ${stdout}`);
+  });
+});
```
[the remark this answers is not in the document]
paologum: actually, you shouldn't call `rm -rf` and instead use `rm` from a library to remove the files, because this assumes you are on a `bash` or `linux` system

paologum (changes requested):

paologum:
also try rerunning `yarn install` because you have a large `.pnp.cjs` file

if it doesn't make it smaller it's fine

PR #150, opened 2024-07-01
Security Pipeline (Truffle, code ql, and dependabot - in settings)

paologum:
Changed pipelines to help resolve security issues

will require dismissals for rate limits because code ql does not recognize fastify rate limiting
(ONLY IF FIXED)

everything else should be manually resolved through the comments it will comment

if truffle finds a secret, you will have to erase the secret from the whole history of git online and if it's a server related secret we will probably have to generate a new one.

There will not be comments for truffle, so you will have to look at the pipeline to resolve it since it already has the previously discussed overhead


JacquiFreeElectron (approved):
looks good!

jay-jhk-lee (approved):
Looks good!

PR #152, opened 2024-07-03
Tsc 424 event search

jay-jhk-lee:
Added event search functionality.
Added "in context" functionality.
notes:
- state holds list of added events so that if multiple events are added and one (or more) is removed, then the time interval goes back to the last highest/lowest ages added.
- this list only persists on the search page and is removed on tab change. This is to reduce possible confusion on changing time intervals when removing events on search page after leaving and returning.
- the "add to chart" checkbox for an event resets on search change or tab change.
- unchecking "add to chart" resets the "on" state of the parents to before toggling.

https://github.com/earthhistoryviz/tsconline/assets/89719669/5ef82396-087c-4447-8669-58057b7d0628

https://github.com/earthhistoryviz/tsconline/assets/89719669/e8e9a2ed-5d17-4a9b-9ccf-139334856e16


paologum:
this is my comments from just watching the video

so we are not going to let them select multiple different "critters" or events to generate?
you didn't really show it in the demo but in your description you said that it'll "reset the add to chart" when they search for another event. why should they not be able to search different ones?

app/src/settings_tabs/Results.tsx
```diff
@@ -0,0 +1,249 @@
...
+        <ErrorOutline />
+      </SvgIcon>
+    );
+  }
+  return (
+    <TSCCheckbox
+      //use global state because if state is in this component, the checkbox rerenders out of view so state is reset
+      //and if state is in parent component, changing it rerenders the entire component which sets scroll position to the top
+      checked={state.settingsTabs.addSearchResultToChart[info.id][0]}
+      onClick={() => {
```
paologum, line 30: to be more deliberate, i think you should make the handler function `onChange`

app/src/settings_tabs/Results.tsx
```diff
@@ -0,0 +1,249 @@
...
+            const parent = state.settingsTabs.columnHashMap.get(pathColumn.parent);
+            if (!parent) break;
+            pathColumn = parent;
+          }
+          actions.setAddSearchResultToChart([false, ...pathAdded.slice(1)], info.id);
+          actions.setColumnOn(false, column);
+        }
+        //in-context feature, adds 3myr to above and below the age
+        if (state.settingsTabs.eventInContext) {
+          if ("age" in info) {
```
paologum, line 60: probably also assert that `age` is what you want it to be with a regex

app/src/settings_tabs/Search.tsx
```diff
@@ -0,0 +1,165 @@
...
+      return [];
+    }
+    while (column.name !== "Chart Root") {
+      columnPath.push(column.editName);
+      column = state.settingsTabs.columnHashMap.get(column.parent!);
+      if (!column) break;
+    }
+    return columnPath;
+  }
+  const count = useRef(0);
```
paologum, line 25: why `useRef` instead of `useState`

app/src/settings_tabs/Search.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { useContext, useEffect, useRef } from "react";
+import "./Search.css";
+import { context } from "../state";
+import { Results } from "./Results";
+import { EventSearchInfo, GroupedEventSearchInfo } from "../types";
+import { TSCCheckbox } from "../components";
+
+export const Search = observer(function Search() {
+  const { state, actions } = useContext(context);
+  function columnPath(name: string): string[] {
```
paologum, line 12: i would call this `makeColumnPath` since `columnPath` is vague

app/src/settings_tabs/Search.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    //key: column name/event name
+    //info: info found in subinfo array
+    const results = new Map<string, EventSearchInfo[]>();
+
+    for (const columnInfo of state.settingsTabs.columnHashMap.values()) {
+      if (columnInfo.name === "Chart Root") {
+        continue;
+      }
+      if (columnInfo.editName.toLowerCase().includes(state.settingsTabs.eventSearchTerm)) {
```
paologum, line 38: `state.settingsTabs.eventSearchTerm.toLowercase()`?

app/src/settings_tabs/Search.tsx
```diff
@@ -0,0 +1,165 @@
...
+      }
+      if (columnInfo.subInfo) {
+        //skip since subInfo is not associated with this but the app does it for map points
+        if (columnInfo.columnDisplayType === "MetaColumn") {
+          continue;
+        }
+        for (let i = 0; i < columnInfo.subInfo.length; i++) {
+          const subInfo = columnInfo.subInfo[i];
+          if ("label" in subInfo) {
+            if (subInfo.label!.toLowerCase().includes(state.settingsTabs.eventSearchTerm)) {
```
paologum, line 60: `state.settingsTabs.eventSearchTerm.toLowercase()`?

app/src/settings_tabs/Search.tsx
```diff
@@ -0,0 +1,165 @@
...
+      <div>
+        <TSCCheckbox
+          checked={state.settingsTabs.eventInContext}
+          onClick={() => {
+            //in-context feature, adds 3myr to above and below the age
+            actions.setEventInContext(!state.settingsTabs.eventInContext);
+            if (state.settingsTabs.eventInContext) {
+              actions.createAgeBeforeContext();
+            }
+          }}
```
paologum, line 129: same here with `onChange`

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
 
+export const setEventSearchTerm = action((term: string) => {
+  state.settingsTabs.eventSearchTerm = term;
+});
+
+export const setEventInContext = action((inContext: boolean) => {
+  state.settingsTabs.eventInContext = inContext;
+});
+
+export const InsertEventInContextTopList = action((term: { key: string; age: number }, unit: string) => {
```
paologum, line 561: can you camel case this

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
+      if (term.age === compareEvent.age && term.key === compareEvent.key) {
+        return;
+      } else if (term.age <= compareEvent.age) {
+        topList.splice(index, 0, term);
+        break;
+      }
+    }
+    if (index === prevLength) {
+      topList.push(term);
+    }
```
paologum, line 581: instead of an extra if statement `index === prevLength`, you can just `return` after you splice and always push at the end

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
+  }
+});
+
+export const InsertEventInContextBaseList = action((term: { key: string; age: number }, unit: string) => {
```
paologum, line 585: camelCase

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
+        return;
+      } else if (term.age >= compareEvent.age) {
+        baseList.splice(index, 0, term);
+        break;
+      }
+    }
+    if (index === prevLength) {
+      baseList.push(term);
+    }
+  }
```
paologum, line 606: same here

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
+        topList.splice(i, 1);
+      }
+    }
+    if (topList.length === 0) {
+      delete state.settingsTabs.eventInContextTopList[unit];
+      if (_.isEmpty(state.settingsTabs.eventInContextTopList)) {
+        state.settingsTabs.eventInContextTopList = null;
+      }
+    }
+  }
```
paologum, line 626: if you are already returning, you don't need to `else` just for less indentation to promote readability

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
+        baseList.splice(i, 1);
+      }
+    }
+    if (baseList.length === 0) {
+      delete state.settingsTabs.eventInContextBaseList[unit];
+      if (_.isEmpty(state.settingsTabs.eventInContextBaseList)) {
+        state.settingsTabs.eventInContextBaseList = null;
+      }
+    }
+  }
```
paologum, line 646: same

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
+export const removeEventInContextBaseList = action((term: { key: string; age: number }, unit: string) => {
+  if (!state.settingsTabs.eventInContextBaseList || !state.settingsTabs.eventInContextBaseList[unit]) {
+    return;
+  } else {
+    const baseList = state.settingsTabs.eventInContextBaseList[unit];
+    for (let i = 0; i < baseList.length; i++) {
+      const compareEvent = baseList[i];
+      if (term.key === compareEvent.key && term.age === compareEvent.age) {
+        baseList.splice(i, 1);
+      }
```
paologum, line 638: you should break to prevent more iterations that are unecessary

app/src/state/actions/column-actions.ts
```diff
@@ -550,6 +550,139 @@ export const setColumnSearchTerm = action((term: string) => {
...
+export const removeEventInContextTopList = action((term: { key: string; age: number }, unit: string) => {
+  if (!state.settingsTabs.eventInContextTopList || !state.settingsTabs.eventInContextTopList[unit]) {
+    return;
+  } else {
+    const topList = state.settingsTabs.eventInContextTopList[unit];
+    for (let i = 0; i < topList.length; i++) {
+      const compareEvent = topList[i];
+      if (term.key === compareEvent.key && term.age === compareEvent.age) {
+        topList.splice(i, 1);
+      }
```
paologum, line 618: same

paologum (changes requested):
some small things but nice!

jay-jhk-lee:
The "reset add to chart" only refers to the checkbox, not the adding of the event itself. The user can add as many events as they want.

app/src/settings_tabs/Search.tsx
jay-jhk-lee: Using state causes the site to crash because it causes a re-render on every increment.

app/src/state/actions/column-actions.ts
```diff
@@ -588,6 +588,132 @@ export const setColumnSearchTerm = action((term: string) => {
...
+    if (_.isEmpty(state.settingsTabs.eventInContextBaseList)) {
+      state.settingsTabs.eventInContextBaseList = null;
+    }
+  }
+});
+
+export const resetEventInContextLists = action(() => {
+  state.settingsTabs.eventInContextBaseList = null;
+  state.settingsTabs.eventInContextTopList = null;
+});
```
paologum, line 683: because we reset the event context wouldn't we want to reset the `ageBeforeContext`?

i also think from a design perspective when we allow the user to center a column in the future in the second column, it seems odd that we allow them to revert back to the context. I don't personally think they should be able to revert because it creates unneeded remembering of state. I could see this being implemented in a larger undo function that remembers events using like preactions so they could revert, but the manual way we reset the `ageBeforeContext` seems unecessary. I look at it as the same way we allow manual top and base age changes, which we don't allow undos/resets on. It's just a further of that which makes it feel weird having memory for this but not that

paologum (changes requested):

app/src/state/actions/column-actions.ts
jay-jhk-lee: I agree with your comment as we've discussed in the meeting on Friday. I'm cleaning up all of the remembering stuff on a different branch, which is almost done and will be up for a pr soon. I personally think we can close this pr and the event search can be implemented through the other branch, let me know what you think.

PR #156, opened 2024-07-04
routes to admin delete, create, display user, delete user datapacks, add/delete server datapacks

paologum:
all backend, frontend will be developed after

realized the tests and the backend were getting quite large so i only made tests for the prehandlers, and deleting and creating users.

will create tests for displaying users, deleting user datapacks, and configuuring server datapacks after

will also implement ip permission configurations later as well

`admin-config.json` will be kept seperate from `config.json` because we want to minimize conflicts in pulls/deployments
so we will always prioritize the git version of `config.json` but we will append/remove as needed when the server is up


server/src/admin-routes.ts
```diff
@@ -0,0 +1,352 @@
...
+export const adminDeleteUser = async function adminDeleteUser(
+  request: FastifyRequest<{ Body: { uuid: string } }>,
+  reply: FastifyReply
+) {
+  const { uuid } = request.body;
+  if (!uuid) {
+    reply.status(400).send({ message: "Missing uuid" });
+    return;
+  }
+  const userDirectory = realpathSync(path.resolve(assetconfigs.uploadDirectory, uuid));
```
asivath, line 99: why realpathSync over realPath?

server/__tests__/admin-routes.test.ts
```diff
@@ -0,0 +1,563 @@
...
+    deleteUser: vi.fn().mockResolvedValue({})
+  };
+});
+
+vi.mock("../src/admin-routes", async (importOriginal) => {
+  const actual = await importOriginal<typeof adminRoutes>();
+  return {
+    ...actual
+  };
+});
```
asivath, line 83: what's the point of this?

server/src/admin-auth.ts
```diff
@@ -0,0 +1,146 @@
...
+          required: ["datapack"]
+        }
+      },
+      config: {
+        rateLimit: moderateRateLimit
+      }
+    },
+    adminDeleteServerDatapack
+  );
+};
```
asivath, line 146: this is hard to read, I would separate the schema definitions to variables

server/src/admin-routes.ts
paologum: the code ql suggeestions suggest `realpathSync` probably because this will also throw if the path doesn't exist to prevent  race conditions. it should probably be fine async though

server/src/admin-routes.ts
```diff
@@ -0,0 +1,352 @@
...
+ * @returns
+ */
+export const adminCreateUser = async function adminCreateUser(request: FastifyRequest, reply: FastifyReply) {
+  const { username, email, password, pictureUrl, isAdmin } = request.body as {
+    username: string;
+    email: string;
+    password: string;
+    pictureUrl: string;
+    isAdmin: number;
+  };
```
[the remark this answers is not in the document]
paologum: since the schema already defines it, i thought this looked cleaner

JacquiFreeElectron (commented):
looks good

asivath (approved):

PR #161, opened 2024-07-06
error message fix

JacquiFreeElectron:
Screenshot 2024-07-06 at 1 19 03 PM


paologum:
from the demo i see, it shouldn't pop up with the snackbar component as an error. it should be what we talked about when you go to the `/verify` and enter an incorrect email

image

JacquiFreeElectron:
Screenshot 2024-07-07 at 2 39 21 AM
The error will disappear when the user click anywhere or hover over it

JacquiFreeElectron:
btw I think the style in `/verify` depends on what browser we use (it looks different when I use Safari), which means it's rendered/offered by the browser(or some other API), I couldn't find how they do the styling, so I just use a tooltip to mimic a popup dialog

app/src/components/TSCError.tsx
```diff
@@ -29,6 +29,44 @@ export const TSCError: React.FC<TSCErrorProps> = observer(({ errorContext, messa
...
+                bgcolor: Color(theme.palette.error.light).lighten(0.1).string()
+              }
+            }
+          }}
+          arrow>
+          <div></div>
+        </Tooltip>
+      </ClickAwayListener>
+    );
+  }
```
paologum, line 69: instead of sort rewiring in a hard coded sort of way across components, i would create a wrapper of a text field and make this a `CustomErrorPopoverTextField` sort of deal, that way we can use this for other things as well

paologum (changes requested):

app/src/components/TSCError.tsx
paologum: which would make it easier to attach an `AnchorEl` since the component is available

paologum: also 1 more thing, we want this for top age too. if it's lower than the base age it should have one, which should make this refactor make more sense

paologum: also, i misspoke the error should be if top age is higher than base age*

JacquiFreeElectron:
how it looks now:
https://github.com/earthhistoryviz/tsconline/assets/104879321/abdb2a81-ea60-43fb-9571-99431fea62bb

JacquiFreeElectron:
new demo: 

https://github.com/earthhistoryviz/tsconline/assets/104879321/cc0058b6-5530-405e-a416-73c1f6a401ce

paologum:
from the demo, it probably should still show the error if `isNan` so that it doesn't go away when it is blank and they think it is fixed. also just in case, i could see this happening is that they are on a different tab and they click generate , they might not know what happened if they aren't on the time tab so you should add the error message that gets pushed on invalid base/top age
except this time it should just be a general one of the nature "Is bad range"

app/src/components/CustomErrorPopoverTextField.tsx
```diff
@@ -0,0 +1,60 @@
...
+  errorContext: ErrorCodes;
+  message: string;
+  anchorElementID: string;
+};
+
+export const CustomErrorPopoverTextField: React.FC<CustomErrorPopoverTextFieldProps> = observer(
+  ({ errorContext, message, anchorElementID }) => {
+    const { actions } = useContext(context);
+    const theme = useTheme();
+    const anchorElement = document.getElementById(anchorElementID);
```
paologum, line 20: we don't wanna query here when we are using javascript with the elements, instead pass in the actual ref of the component.

app/src/settings_tabs/Time.tsx
```diff
@@ -15,6 +16,13 @@ export const Time = observer(function Time() {
     throw new Error("There must be a unit used in the config");
   }
   const disabled = units !== "Ma";
+  function checkAgeRange(top: number, base: number, elementID: string) {
+    if (top > base) {
+      actions.pushError(ErrorCodes.INVALID_UNIT_RANGE, elementID);
+    } else {
+      state.errors.errorAlerts.delete(ErrorCodes.INVALID_UNIT_RANGE);
+    }
+  }
```
paologum, line 25: use `removeError`, you don't want to change state out of an action

paologum (changes requested):

JacquiFreeElectron:
new demo:

https://github.com/user-attachments/assets/2434c200-52ab-4a74-9e8d-fc8ed5d59d6b

app/src/components/CustomErrorPopoverTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const CustomErrorPopoverTextField: React.FC<CustomErrorPopoverTextFieldProps> = observer(
+  ({ errorContext, message, anchorElement }) => {
+    const { actions } = useContext(context);
+    const theme = useTheme();
+    function handleCloseError(_event: React.SyntheticEvent | Event, reason?: string) {
+      if (reason === "clickaway") return;
+      actions.removeError(errorContext);
+    }
```
paologum, line 23: i'm confused how this works, your demo has clickaway remove the error, but here in the code it seems to me like it would ignore clickaway?

just wondering how this works in the demo

JacquiFreeElectron: I honestly don't know, I didn't even notice it bc I was directly using it from the original code, didn't change anything. Probably it won't make a difference if I remove this?

JacquiFreeElectron: Just tried, I can totally remove this

paologum (approved):

PR #166, opened 2024-07-10
tests for admin upload configure datapacks and getting users

paologum:


server/src/admin-routes.ts
```diff
@@ -22,17 +22,17 @@ import { promisify } from "util";
...
-    };
-  });
-  reply.send(displayedUser);
+  try {
+    const users = await findUser({});
+    const displayedUsers = users.map((user) => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { hashedPassword, ...displayedUser } = user;
+      return displayedUser;
+    });
```
asivath, line 31: could be cleaner with underscore but whatever

asivath (approved):

server/src/admin-routes.ts
paologum: you can't destructure if you add an underscore unless you maybe use `as` but atp it would be weird

PR #174, opened 2024-07-12
Tsc 351 encryption jar tests (new branch)

JacquiFreeElectron:
new branch since there are too many test logs on the old one


server/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,608 @@
...
+  if (await checkFileExists(generatedFilePath)) {
+    throw new Error("test generated file directory shouldn't exist");
+  }
+});
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+describe("requestDownload", () => {
```
paologum, line 105: should have a test that checks for if the access fails for if the regular datapack file check throws an error that is not `ENOENT` on line 131 in `routes.ts`
```
if (error.code === "ENOENT") {
      const errormsg = "The file requested " + filename + " does not exist within user's upload directory";
      reply.status(404).send({ error: errormsg });
    } else {
      reply.status(500).send({ error: "An error occurred: " + e });
    }
```

paologum: also should have a test for errors that get thrown and caught on line 168-175 in `routes.ts`
the last access check which is basically the same code as above but you should still check that this gets correctly sent back as the correct message

server/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,608 @@
...
+let jarFilePath = "";
+let resultPath = "server/__tests__/__data__/encryption-test-generated-file";
+if (await checkFileExists("/home/runner/work/tsconline/tsconline/server/assets/jars/testUsageJar.jar")) {
+  jarFilePath = "/home/runner/work/tsconline/tsconline/server/assets/jars/testUsageJar.jar";
+  resultPath = "/home/runner/work/tsconline/tsconline/server/__tests__/__data__/encryption-test-generated-file";
+} else {
+  try {
+    const contents = JSON.parse((await unmockedReadFile("server/assets/config.json")).toString());
+    if (typeof contents !== "object") throw new Error("AssetConfig must be an object");
+    if (typeof contents.activeJar !== "string") throw new Error('AssetConfig must have an "activeJar" string');
```
paologum, line 507: you should just `assertAssetConfig` from `types`

server/src/encryption.ts
```diff
@@ -0,0 +1,31 @@
...
+    try {
+      const stdout = execFileSync("java", [
+        "-jar",
+        activeJar,
+        "-d",
+        filepath.replaceAll("\\", "/"),
+        "-enc",
+        encryptedFilepathDir,
+        "-node"
+      ]);
```
paologum, line 24: this should be not synchronous so use the async version

server/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,608 @@
...
+describe("checkHeader", () => {
+  test('checkHeader("encrypted.txt") returns true', async () => {
+    expect(await unmockedCheckHeader("encrypted.txt")).toEqual(true);
+  });
+  test('checkHeader("unencrypted.txt") returns false', async () => {
+    expect(await unmockedCheckHeader("unencrypted.txt")).toEqual(false);
+  });
+});
+
+const { rm: unmockedRm } = await import("fs/promises");
```
paologum, line 608: what is this last import for

paologum (changes requested):
sometimes the tests take longer, so for these tests specifically you should configure the timeouts to be longer ~20 seconds

also, mock ignore `console.logs`

server/__tests__/encryption.test.ts
JacquiFreeElectron: his is for using the real implementation of `rm `in clearing all generated files. `doUnmock()` requires importing again to make it effective.

server/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,608 @@
...
+      method: "GET",
+      url: "/hasuuid/download/user-datapacks/:filename?needEncryption=true"
+    });
+    expect(accessSpy).toHaveBeenCalledTimes(2);
+    expect(runJavaEncryptSpy).not.toHaveBeenCalled();
+    expect(checkHeaderSpy).toHaveNthReturnedWith(1, false);
+    expect(readFileSpy).toHaveNthReturnedWith(1, "default content");
+    expect(response.statusCode).toBe(500);
+    expect(response.json().error).toBe("Failed to create encrypted directory with error Error: Unknown Error");
+  });
```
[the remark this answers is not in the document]
JacquiFreeElectron: Do you mean I should also call `rm` on `maybeEncryptedFilepath` when `mkdir` fails and expect `rm` gets called in here? Or you mean I should check `rm `is not called?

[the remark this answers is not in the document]
paologum: expect that it doesn't get called

server/__tests__/encryption.test.ts
JacquiFreeElectron: As says in their documentation, `vi.rstoreAllMocks`"Will call [.mockRestore()](https://v0.vitest.dev/api/mock#mockrestore) on all spies. This will clear mock history and reset its implementation to the original one." and there's a note under `mockRestore()` says:  "Note that restoring mock from vi.fn() will set implementation to an empty function that returns undefined."    and bc I used vi.fn() for the fs/promises module, I need to use doUnmock().   I'll change the import go in the `afterAll`

PR #175, opened 2024-07-12
Tsc 413 download chart as png pdf make dialog focus on mount

jay-jhk-lee:
Download chart as pdf and png works
A few concerns are the time it takes to save a pdf and the resolution of the pdf
used webworker to asynchronously create the pdf since it is computationally intensive

https://github.com/user-attachments/assets/99b6f212-bfe2-4855-809d-b65f877ce34e

dark mode with loading button
Screen Shot 2024-07-15 at 7 56 02 PM

pdf

[test.pdf](https://github.com/user-attachments/files/16199248/test.pdf)

png

test


app/src/ChartOptionsBar.tsx
```diff
@@ -181,17 +181,85 @@ export const OptionsBar: React.FC<OptionsBarProps> = observer(({ transformRef, s
...
+          if (state.chartTab.downloadFiletype === "pdf") {
+            const downloadWorker: Worker = new Worker(new URL("./util/workers/download-pdf.ts", import.meta.url), {
+              type: "module"
+            });
+            const message: DownloadPdfMessage = { imgURI: imgURI, height: svgHeight, width: svgWidth };
+            downloadWorker.postMessage(message);
+            downloadWorker.onmessage = function (e: MessageEvent<DownloadPdfCompleteMessage>) {
+              const { status, value } = e.data;
+              if (status === "success" && value) {
+                FileSaver.saveAs(value, state.chartTab.downloadFilename + ".pdf");
```
paologum, line 237: technically all this code doesn't need to be in the `image.onload` function
the creation of the image is mainly just for the `url` so i think for better readability you can do something like
```

function svgToImageURI(svgString: (idk), width: number, height: number) {
  return new Promise((resolve, reject) => {
    const svgBlob = new Blob([svgString], { type: "image/svg+xml;charset=utf-8" });
    const url = URL.createObjectURL(svgBlob);

    const image = new Image();
    image.width = width;
    image.height = height;
    image.src = url;

    image.onload = function () {
      const canvas = document.createElement("canvas");
      const ctx = canvas.getContext("2d");
      if (!ctx) return;

      const pixelRatio = 3;
      canvas.width = width * pixelRatio;
      canvas.height = height * pixelRatio;
      ctx.setTransform(pixelRatio, 0, 0, pixelRatio, 0, 0);
      ctx.drawImage(image, 0, 0);

      URL.revokeObjectURL(url);
      const imgURI = canvas.toDataURL();
      canvas.remove();
      resolve(imgURI);
    };

    image.onerror = reject;
  });
}
```

and then use the resolved value for the webworker

app/src/ChartOptionsBar.tsx
```diff
@@ -181,17 +181,85 @@ export const OptionsBar: React.FC<OptionsBarProps> = observer(({ transformRef, s
...
+              downloadWorker.terminate();
+            };
+          } else if (state.chartTab.downloadFiletype === "png") {
+            const a = document.createElement("a");
+            a.download = state.chartTab.downloadFilename + ".png"; // filename
+            a.target = "_blank";
+            a.href = imgURI;
+            a.click();
+            actions.pushSnackbar("Saved Chart as PNG!", "success");
+            actions.setChartTabIsSavingChart(false);
```
paologum, line 252: should remove the `a` element

app/src/ChartOptionsBar.tsx
paologum: also let's say `ctx` returns, then we won't have an `imgUri` so you might/should want to `reject` and handle that properly with error pushing and the like

paologum (changes requested):
how does it look on dark mode (loading button)

also since we webworker this, exiting the dialog should still download the chart with the webworker when they leave/press cancel right

jay-jhk-lee:
Yes, I've realized that the text is misleading so I changed it

paologum:
the `.yarn` and `.pnp.cjs` were removed, so remove them from the commit

app/src/ChartOptionsBar.tsx
paologum, line 1: it might be good, if they select pdf to have a tooltip that says they can leave and do other stuff since generating a pdf takes a while. up to you though

paologum (approved):

PR #177, opened 2024-07-14
Tsc 431 pr puts up a server

asivath:
Check this fork for examples of the workflows: https://github.com/asivath/tsconline-test


.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,212 @@
...
+
+      - name: Comment if container is not running or is for the same PR
+        if: ${{ steps.fetch_deployment_status.outputs.container_running == 'false' }}
+        uses: thollander/actions-comment-pull-request@v2
+        with:
+          message: |
+            🚀 Preview deployed! You can access it [here](https://pr-preview.geolex.org).
+            To bring down the preview close the PR.
+
+  confirm-deploy:
```
paologum, line 111: should this "need" `deploy-preview` because i mean if it doesn't, why wouldn't everyone just use `/confirm-deploy` regardless

.github/workflows/merge-pipeline.yml
```diff
@@ -179,4 +178,80 @@ jobs:
 
       - name: Check for any test failures
         if: steps.run_tests.outputs.test_failed == 'true'
-        run: exit 1
\ No newline at end of file
+        run: exit 1
+
+  setup-preview:
+    needs: [check-files, code-quality, build-and-test]
```
paologum, line 184: no security pipeline need? this could technically make the server open to attacks if a preview is up with bad security code

paologum:
what happens if status check fails and the preview does not get setup on the server, and someone comments `/deploy`

.yarn/cache/fsevents-patch-6b67494872-10.zip
paologum, line 1: also remove this commit/rebase for the yarn cache commit stuff?

asivath:
> what happens if status check fails and the preview does not get setup on the server, and someone comments `/deploy`

nothing would really happen on the server since the preview directory won't exist and the action will fail

paologum (approved):

PR #178, opened 2024-07-15
change error popover to helper text

JacquiFreeElectron:
demo: 

https://github.com/user-attachments/assets/44b63bdf-d039-4088-9ee4-bc0db54b1ffe

in the video, the color of the outline may look like a combination of black & red when the error occurs due to screen recording issue, but it's actually just red as shown in the picture: 
Screenshot 2024-07-15 at 4 30 20 PM


paologum:
you should capitalize the b in base and also i think it would make more sense for the error to be `Base age must be greater than the top age`

paologum:
Also is there a reason in the video demo that the red outline doesn't go around the whole box?

app/src/components/TSCError.tsx
```diff
@@ -17,64 +16,53 @@ type TSCErrorProps = {
   count: number;
   anchorElementRef?: RefObject<HTMLElement>;
 };
-export const TSCError: React.FC<TSCErrorProps> = observer(
-  ({ errorContext, message, index, count, anchorElementRef }) => {
```
paologum, line 21: remove the `TSCErrorProps` `anchorElementRef`

.yarn/cache/@esbuild-darwin-arm64-npm-0.20.2-e287d70c91-10.zip
paologum, line 1: we removed `.yarn/cache` so can you remove these from your pr

paologum (approved):

PR #179, opened 2024-07-17
TSC-351-Encryption tests

JacquiFreeElectron:
try to solve the yarn install issue in pipeline so I open a new branch


server/__tests__/encryption.test.ts
```diff
@@ -0,0 +1,681 @@
...
+} else {
+  try {
+    const contents = JSON.parse((await unmockedReadFile("server/assets/config.json")).toString());
+    assertAssetConfig(contents);
+    jarFilePath = "server/" + contents.activeJar;
+  } catch (e) {
+    throw new Error("ERROR: Failed to load local jar file path from assets/config.json.  Error was: " + e);
+  }
+}
+if (!jarFilePath) throw new Error("jar file path shouldn't be empty");
```
paologum, line 572: i just realized that all the tests in these files come from different files, which is hk very confusing so you should abstract each set of tests to their own files. 

the route stuff here is fine, but put it/rename this to `encryption-routes.test.ts`
you might also want to move the encryption route to it's own file but that's up to you. but since the tests for that route are pretty intensive itself it should be by itself.

the `encryption` stuff that runs java should be in it's own probably called `encryption.test.ts` or what this file is currently named.

lastly, the `util` stuff should go in the already made `util.test.ts`

paologum: this will also help make this have less overhead on what is really mocked and isn't
so you won't need to `doUnmock()` and can just import once with better hindsight

JacquiFreeElectron: > i just realized that all the tests in these files come from different files, which is hk very confusing so you should abstract each set of tests to their own files.
> 
> the route stuff here is fine, but put it/rename this to `encryption-routes.test.ts` you might also want to move the encryption route to it's own file but that's up to you. but since the tests for that route are pretty intensive itself it should be by itself.
> 
> the `encryption` stuff that runs java should be in it's own probably called `encryption.test.ts` or what this file is currently named.
> 
> lastly, the `util` stuff should go in the already made `util.test.ts`

Ok I'll move the runJavaEncrypt test and the checkHeader test to other files, but I'm confused about the route you talked about, isn't there just one route? encrypted download & retrieve original is just one route (the `requestDownload` route)

paologum: sure yeah that sounds good, i just want the tests in separate files for ease of use

server/src/routes.ts
```diff
@@ -134,6 +135,7 @@ export const requestDownload = async function requestDownload(
       const errormsg = "The file requested " + filename + " does not exist within user's upload directory";
       reply.status(404).send({ error: errormsg });
     } else {
+      //add test
```
paologum, line 138: one removal and it should be good

paologum (approved):

PR #181, opened 2024-07-17
Tsc 446 event search new perspective

jay-jhk-lee:
After July 19th meeting
total column count: 7
Status:

- indicates the status of event
- 4 statuses: Column on & age in interval, Column on & age not in interval, Column off & age in interval, Column off & age not in interval
- events/columns with no ages will only have Column on and column off statuses

Column: Same as before
Center: previously part of age column, same functionality
Extend: previously part of age column, same functionality
Age: Only shows age of event now
Qualifier: Same as before
Notes: previously Additional Info, changed for conciseness

Help Icon: Explains what center and extend does

image

image

___________________________________________________________________
Before July 19th meeting
Reduced column count to 4
combined toggle column and column path

- column toggled on click of cell
- icon indicates if column is toggled (including its parents)

combined age and modify time interval

- icon indicates if age is within time interval

https://github.com/user-attachments/assets/c5f35ed2-f58b-4ad8-83d2-02fc0b84fead

image


app/src/settings_tabs/Results.tsx
```diff
@@ -14,28 +14,87 @@ import HorizontalRuleIcon from "@mui/icons-material/HorizontalRule";
...
+    console.error(info.unit + "not in time settings");
+    return false;
+  }
+  const chartTopAge = state.settings.timeSettings[info.unit].topStageAge;
+  const chartBaseAge = state.settings.timeSettings[info.unit].baseStageAge;
+  const ages = info.age;
+  if (!ages) return false;
+  if (ages.topAge + 3 >= chartTopAge && ages.baseAge - 3 <= chartBaseAge) {
+    return true;
+  }
```
paologum, line 32: hmm so it's considered out of the interval if it's only within 2 of the `chartBaseAge`?

app/src/settings_tabs/Results.tsx
```diff
@@ -14,28 +14,87 @@ import HorizontalRuleIcon from "@mui/icons-material/HorizontalRule";
...
+          placement="right"
+          title={columnPath.map((value, pathIndex) => (
+            <div key={pathIndex}>{value}</div>
+          ))}>
+          <Typography noWrap variant="subtitle2">
+            {columnPath[0]}
+          </Typography>
+        </CustomTooltip>
+      </div>
+    </div>
```
paologum, line 93: i would wrap in a fragment if this doesn't affect styling

app/src/settings_tabs/Results.tsx
```diff
@@ -94,25 +140,67 @@ const ModifyTimeInterval = observer(({ info }: { info: EventSearchInfo }) => {
...
+        <SvgIcon>
+          <ErrorOutline />
+        </SvgIcon>
+      );
+    if (info.age.topAge === info.age.baseAge) {
+      return <div>{info.age.topAge}</div>;
+    }
+    return (
+      <div>
+        {info.age.topAge} - {info.age!.baseAge}
```
paologum, line 156: why does only `baseAge` have a `!`

app/src/settings_tabs/Results.tsx
```diff
@@ -94,25 +140,67 @@ const ModifyTimeInterval = observer(({ info }: { info: EventSearchInfo }) => {
...
     }
   };
+
+  const AgeDisplay = () => {
+    if (!info.age)
+      return (
+        <SvgIcon>
+          <ErrorOutline />
+        </SvgIcon>
+      );
```
paologum, line 150: for events that don't have ages, currently it displays it sort of as two error icons then?

was this one originally just like a line if it didn't exist?

app/src/settings_tabs/Search.tsx
```diff
@@ -124,38 +124,74 @@ export const Search = observer(function Search() {
...
+        </ToggleButtonGroup>
+        <Box className="search-time-display-ages-container">
+          <Box
+            className="search-time-display-ages"
+            sx={(theme) => ({
+              display: "inline",
+              borderRadius: 2,
+              bgcolor: theme.palette.secondaryBackground.main
+            })}>
+            Top Age: {state.settings.timeSettings[units].topStageAge}
```
paologum, line 161: just a comment but you can just do `bgcolor="secondaryBackground.main"` in the `Box` props so you don't have to like deconstruct a larger theme object

app/src/settings_tabs/Results.tsx
```diff
@@ -167,48 +247,28 @@ export const Results = ({ groupedEvents }: { groupedEvents: GroupedEventSearchIn
       return (
         <>
           <TableCell align="left">
-            <div>
-              <ToggleColumn columnName={info.columnName} />
-            </div>
+            <Column info={info} />
```
paologum, line 250: is there any way to make the column cell as big as it can be, because rn it looks like it cuts column names off really early as opposed to the space it looks like it has to display the name

Screenshot 2024-07-18 at 12 35 30 PM

paologum (changes requested):

app/src/settings_tabs/Results.tsx
jay-jhk-lee: No, it should only display the line since it checks if age exists on the parent component. It was supposed to be a backup check, but now that I think about it the check should only be in that component. I'll change it

app/src/settings_tabs/Results.tsx
jay-jhk-lee: Oops, I'll change that

app/src/settings_tabs/Results.tsx
jay-jhk-lee: Mistake, will remove

jay-jhk-lee:
image

app/src/settings_tabs/Results.tsx
```diff
@@ -62,25 +62,47 @@ const Column = observer(({ info }: { info: EventSearchInfo }) => {
...
+          if (!columnCell) return;
+          columnCell.style.width = width + "px";
+        }
+      }
+    });
+    resize.observe(table!);
+    return () => {
+      resize.disconnect();
+    };
+  }, []);
```
paologum, line 84: instead of changing the width of the table cell, why can't you just make the `column` cell `flex-grow` and `flex-shrink`

jay-jhk-lee: resolved in latest push

paologum:
i would still have the tooltips that say center or extend within a 3mya (i forget the unit) boundary

jay-jhk-lee:
I added a help icon for that before i read your comment since its a little bit different for each function and I couldn't find a concise way to word it, let me know if I should still add it to the tooltip

jay-jhk-lee:
I added a help icon for that before I read your comment, it's a bit more specific in case someone is confused from the tooltip. let me know if you think I should remove it or change it.

paologum:
hmm i don't think we want four icons, (Column off & age not in interval and Column on & age not in interval) both mean the same thing in the context of this display

both mean that the event won't be included, so it should just have 3 similar to the column page because that's what the team agreed on today

app/src/settings_tabs/Results.tsx
```diff
@@ -2,117 +2,351 @@ import { TableCell, TableBody, TableContainer, Paper, SvgIcon, Typography, Box,
...
+const verifyAgesAndAddAgeMargin = (age: { topAge: number; baseAge: number } | undefined) => {
+  if (!age) return null;
+  let { topAge, baseAge } = age;
+  if ((topAge -= 3) < 0) {
+    topAge = 0;
+  }
+  if ((baseAge += 3) < 0) {
+    baseAge = 0;
+  }
+  return { topAge: topAge, baseAge: baseAge };
```
paologum, line 174: `baseAge += 3` can't be < 0 unless it's negative i'm assuming you mean the other way

also for readability you might wanna do like `min` and `max` so it's easier at a glance

app/src/settings_tabs/Results.tsx
```diff
@@ -2,117 +2,351 @@ import { TableCell, TableBody, TableContainer, Paper, SvgIcon, Typography, Box,
...
+};
+
+const truncToSecondDecimal = (num: number) => {
+  return Math.trunc(num * 100) / 100;
+};
```
paologum, line 179: we truncate now? if this is a general thing why only results and not the whole app in the action?

app/src/settings_tabs/Results.tsx
```diff
@@ -2,117 +2,351 @@ import { TableCell, TableBody, TableContainer, Paper, SvgIcon, Typography, Box,
...
+
+const Qualifier = observer(({ info }: { info: EventSearchInfo }) => {
+  if (!info.qualifier) {
+    return (
+      <SvgIcon>
+        <HorizontalRuleIcon />
+      </SvgIcon>
+    );
+  }
+  return <div>{info.qualifier}</div>;
```
paologum, line 315: i forgot to mention but these should be in `Typography` no?

paologum (changes requested):
and my comment above

app/src/settings_tabs/Results.tsx
paologum: there are also most likely in human events, Ma's that are very small and could go pretty far past like Lt. Pliest
so if events are like that, the top age and base age would get cut off?

jay-jhk-lee: I see, I did this for the case where the float bugs out and has a bunch of zeros in the end (8.000000000..). I forgot about that case, I'll modify it

app/src/settings_tabs/Results.tsx
jay-jhk-lee: yes, I was catching for the case where it could possibly be negative even though that is most likely not going to happen.

jay-jhk-lee:
Yes, both mean that the event won't be displayed on the chart. However, I think it's important to specify why it's not in the chart since we can both toggle the column and change the time interval. So I'll change the orange icon to be the same but change the tooltips to convey that information.

paologum:
yep i think that's a good idea with the tooltips
if you want a little distinction i think a different color might be a good idea too, probably not the same as the red x though but it's definitely something that might be helpful but i like this

app/src/settings_tabs/Results.tsx
```diff
@@ -164,18 +170,16 @@ const Column = observer(({ info }: { info: EventSearchInfo }) => {
...
 const verifyAgesAndAddAgeMargin = (age: { topAge: number; baseAge: number } | undefined) => {
   if (!age) return null;
-  let { topAge, baseAge } = age;
-  if ((topAge -= 3) < 0) {
-    topAge = 0;
+  let min = age.topAge;
+  let max = age.baseAge;
+  //for floating point inaccuracies
+  if (Number(bigDecimal.add(String(min), "-3")) < 0) {
+    min = 0;
```
paologum, line 177: why is it having large floating point decimals?

paologum (approved):
looks good, up to you if you wanna do color thing

app/src/settings_tabs/Results.tsx
jay-jhk-lee: I'm not sure, I think it's a problem with adding floats in general with javascript.

PR #184, opened 2024-07-21
Tsc 311 functionality to add blank age column

JacquiFreeElectron:
demo: 

https://github.com/user-attachments/assets/72372ccd-1d6d-45b6-8b80-ca935c3b0645

also I changed the default justification of Ma to "right" bc that's the default value in the jar. (and the default justification of any other age ruler is "left").

### DIFF
--- a/app/src/settings_tabs/column_menu/ColumnMenu.tsx
+++ b/app/src/settings_tabs/column_menu/ColumnMenu.tsx
@@ -97,11 +97,13 @@ type ColumnContentProps = {
 };
 const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) => {
   const { actions } = useContext(context);
+  const [ageSerialNum, setAgeSerialNum] = useState(1);
   function addBlankColumn() {
     actions.addBlankColumn(column);
   }
   function addAgeColumn() {
-    actions.addAgeColumn(column);
+    setAgeSerialNum(ageSerialNum + 1);
+    actions.addAgeColumn(column, ageSerialNum);
   }
   function changeAgeColumnJustification(event: React.ChangeEvent<HTMLInputElement>) {
     const newJustification = event.target.value === "right" ? "right" : "left";
--- a/app/src/settings_tabs/column_menu/ColumnMenu.tsx
+++ b/app/src/settings_tabs/column_menu/ColumnMenu.tsx
@@ -97,13 +97,11 @@ type ColumnContentProps = {
 };
 const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) => {
   const { actions } = useContext(context);
-  const [ageSerialNum, setAgeSerialNum] = useState(1);
   function addBlankColumn() {
     actions.addBlankColumn(column);
   }
   function addAgeColumn() {
-    setAgeSerialNum(ageSerialNum + 1);
-    actions.addAgeColumn(column, ageSerialNum);
+    actions.addAgeColumn(column);
   }
   function changeAgeColumnJustification(event: React.ChangeEvent<HTMLInputElement>) {
     const newJustification = event.target.value === "right" ? "right" : "left";
--- a/app/src/settings_tabs/column_menu/ColumnMenu.tsx
+++ b/app/src/settings_tabs/column_menu/ColumnMenu.tsx
@@ -5,7 +5,7 @@ import { Box, Typography } from "@mui/material";
 import "./ColumnMenu.css";
 import { FontMenu } from "../FontMenu";
 import { ChangeBackgroundColor } from "./BackgroundColor";
-import { ColumnInfo, assertRangeSettings } from "@tsconline/shared";
+import { ColumnInfo, assertRangeSettings, assertRulerSettings } from "@tsconline/shared";
 import {
   CustomDivider,
   CustomFormControlLabel,
@@ -21,6 +21,8 @@ import { DataMiningSettings } from "../advanced_settings/DataMiningSettings";
 import AccordionPositionControls from "./AccordionPositionControls";
 import { CustomTabs } from "../../components/TSCCustomTabs";
 import { RangeSpecificSettings } from "../advanced_settings/RangeSpecificSettings";
+import { Button } from "react-bootstrap";
+import { TSCRadioGroup } from "../../components/TSCRadioGroup";
 
 export const ColumnMenu = observer(() => {
   const { state } = useContext(context);
@@ -95,6 +97,26 @@ type ColumnContentProps = {
 };
 const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) => {
   const { actions } = useContext(context);
+  function addBlankColumn() {
+    actions.addBlankColumn(column);
+  }
+  function addAgeColumn() {
+    actions.addAgeColumn(column);
+  }
+  function changeAgeColumnJustification(event: React.ChangeEvent<HTMLInputElement>) {
+    const newJustification = event.target.value === "right" ? "right" : "left";
+    actions.changeAgeColumnJustification(column, newJustification);
+  }
+  const isRulerAgeColumn = () => {
+    if (!(column.columnDisplayType === "Ruler" && column.name.includes("Age"))) return "";
+    try {
+      assertRulerSettings(column.columnSpecificSettings);
+      return column.columnSpecificSettings.justification;
+    } catch (e) {
+      return "";
+    }
+  };
+
   switch (tab) {
     case "General":
       return (
@@ -126,6 +148,28 @@ const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) =
             <EventSpecificSettings column={column} />
             <RangeSpecificSettings column={column} />
             {!!column.popup && <InfoBox info={column.popup} />}
+            {column.children.length != 0 && (
+              <Box className="add-blank-or-age-button-container">
+                <Button className="add-blank-or-age-button" onClick={addBlankColumn}>
+                  <Typography>Add Blank Column</Typography>
+                </Button>
+                <Button className="add-blank-or-age-button" onClick={addAgeColumn}>
+                  <Typography>Add Age Column</Typography>
+                </Button>
+              </Box>
+            )}
+
+            {!!isRulerAgeColumn() && (
+              <TSCRadioGroup
+                onChange={changeAgeColumnJustification}
+                name={"Age Ruler Justification"}
+                value={isRulerAgeColumn()}
+                radioArray={[
+                  { value: "left", label: "Left" },
+                  { value: "right", label: "Right" }
+                ]}
+              />
+            )}
           </Box>
         </StyledScrollbar>
       );
--- a/app/src/settings_tabs/column_menu/ColumnMenu.tsx
+++ b/app/src/settings_tabs/column_menu/ColumnMenu.tsx
@@ -11,6 +11,7 @@ import {
   CustomFormControlLabel,
   GenericTextField,
   StyledScrollbar,
+  TSCButton,
   TSCCheckbox
 } from "../../components";
 import { InfoBox } from "./InfoBox";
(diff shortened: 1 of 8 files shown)


paologum:
for the buttons i would use `TSCButton` so they adhere to the theme

app/src/settings_tabs/column_menu/ColumnMenu.tsx
```diff
@@ -95,6 +97,26 @@ type ColumnContentProps = {
...
+  }
+  function addAgeColumn() {
+    actions.addAgeColumn(column);
+  }
+  function changeAgeColumnJustification(event: React.ChangeEvent<HTMLInputElement>) {
+    const newJustification = event.target.value === "right" ? "right" : "left";
+    actions.changeAgeColumnJustification(column, newJustification);
+  }
+  const isRulerAgeColumn = () => {
+    if (!(column.columnDisplayType === "Ruler" && column.name.includes("Age"))) return "";
```
paologum, line 111: for easier readability i would just have `||` instead of inverting using DeMorgan's law to inverse it when you read

app/src/state/actions/column-actions.ts
```diff
@@ -531,6 +532,86 @@ export const removeDataMiningColumn = action((column: ColumnInfo, type: string)
   state.settingsTabs.columnHashMap.delete(columnToRemove);
 });
 
+export const addBlankColumn = action((column: ColumnInfo) => {
+  if (column.children.length == 0) {
+    console.log("WARNING: tried to add a blank column to a column with no children");
+    return;
+  }
+  let serialNumber = 1;
+  const biggestExistedSerialNum = column.children.findLastIndex((child) => child.name.includes("Blank"));
```
paologum, line 541: grammar thing but it would make more sense for this to be called `largestExistingSerialNum`

app/src/state/actions/column-actions.ts
```diff
@@ -531,6 +532,86 @@ export const removeDataMiningColumn = action((column: ColumnInfo, type: string)
...
+  if (biggestExistedSerialNum > -1) {
+    serialNumber = Number(column.children[biggestExistedSerialNum].name.charAt(6)) + 1;
```
paologum, line 543: you should future proof this for multiple digits

app/src/state/actions/column-actions.ts
```diff
@@ -531,6 +532,86 @@ export const removeDataMiningColumn = action((column: ColumnInfo, type: string)
...
+    console.log("WARNING: tried to add a blank column to a column with no children");
+    return;
+  }
+  let serialNumber = 1;
+  const biggestExistedSerialNum = column.children.findLastIndex(
+    (child) => child.name.includes("Age") && child.columnDisplayType === "Ruler"
+  );
+  if (biggestExistedSerialNum > -1) {
+    serialNumber = Number(column.children[biggestExistedSerialNum].name.charAt(4)) + 1;
+  }
```
paologum, line 578: here too, possibly just make a helper function since this is basically the same code

app/src/state/actions/column-actions.ts
```diff
@@ -531,6 +532,86 @@ export const removeDataMiningColumn = action((column: ColumnInfo, type: string)
...
+  const ageColumnName = "Age " + `${serialNumber}` + " for " + column.name;
+  const maWidth = state.settingsTabs.columnHashMap.get("Ma")?.width;
```
paologum, line 580: probably should hard code or something, unless this is how they do it in the java file since other units also exist as ruler types, or alternatively make a constant somewhere that applies to both the ruler we apply and the one here

app/src/state/actions/column-actions.ts
```diff
@@ -531,6 +532,86 @@ export const removeDataMiningColumn = action((column: ColumnInfo, type: string)
...
+  const ageColumn: ColumnInfo = observable({
+    ...cloneDeep(column),
```
paologum, line 582: this age column will have the same `Subinfo`? it probably shouldn't
also applies to blank

app/src/state/parse-settings.ts
```diff
@@ -546,7 +567,14 @@ export function columnInfoTSCToXml(column: ColumnInfoTSC, indent: string): strin
       continue;
     }
     if (key === "title") {
-      xml += `${indent}<setting name="title">${column[key]}</setting>\n`;
+      let title = column[key];
+      if (/^Age \d+ for .+$/.test(column[key])) {
+        title = "Age";
+      } else if (/^Blank \d+ for .+$/.test(column[key])) {
+        title = "Blank " + column[key].charAt(6);
```
paologum, line 574: multi digits

shared/src/index.ts
```diff
@@ -195,6 +195,7 @@ export type ColumnInfoType = keyof ColumnInfoTypeMap;
 
 export type DisplayedColumnTypes =
   | ColumnInfoType
+  | "Data"
```
paologum, line 198: is this not just `Blank` in `ColumnInfoType`?

paologum (changes requested):

shared/src/index.ts
JacquiFreeElectron: Nope, it's data in the jar settings file, blank is another type of column

app/src/state/actions/column-actions.ts
JacquiFreeElectron: ok

app/src/state/actions/column-actions.ts
JacquiFreeElectron: In the jar, they use 'auto' to make it the same as the Ma. So far, the width of Ma is always 44.6953125, but I'm not sure if it could be changed under some circumstances... I can make a constant if you want to

JacquiFreeElectron: and I just realized it works even if the ruler column doesn't have a defined width, the jar will set a width for it automatically. so I can probably just set `width `to `undefined`. Do you think this is good? (bc there isn't an explicitly defined width for Ma too)

paologum: yup that sounds good

app/src/settings_tabs/column_menu/AddBlankColumnButton.tsx
paologum, line 1: empty file

app/src/util/util.ts
```diff
@@ -161,3 +161,10 @@ function getContrastText(color1: string) {
     return "#000000";
   }
 }
+
+export function findSerialNum(name: string) {
+  const firstBlankspaceIndex = name.indexOf(" ");
+  const secondBlankspaceIndex = name.indexOf(" for");
+  const serialNumber = Number(name.substring(firstBlankspaceIndex + 1, secondBlankspaceIndex));
+  return serialNumber;
+}
```
paologum, line 170: if either `first` or `second` are not found (i.e -1) then `return 0`

paologum: also have errors/returns for if `serialNumber` is `NaN`

paologum (changes requested):

paologum:
also the generated chart has negative `age`, is this normal?

JacquiFreeElectron:
> also the generated chart has negative `age`, is this normal?

yep, as least it's the same as the jar: 
Screenshot 2024-07-22 at 7 02 23 PM

app/src/util/util.ts
JacquiFreeElectron: ok

app/src/settings_tabs/column_menu/AddBlankColumnButton.tsx
JacquiFreeElectron: will remove

paologum (changes requested):

app/src/state/parse-settings.ts
```diff
@@ -466,6 +474,19 @@ export function translateColumnInfoToColumnInfoTSC(state: ColumnInfo): ColumnInf
...
+      if (/^Blank \d+ for .+$/.test(state.name)) {
+        column._id = "class datastore.DataColumn:Blank " + findSerialNum(state.name);
+        break;
+      }
+    }
+    case "Ruler": {
+      if (/^Age \d+ for .+$/.test(state.name)) {
+        column._id = "class datastore.RulerColumn:Age " + findSerialNum(state.name);
+        break;
+      }
```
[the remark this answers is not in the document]
JacquiFreeElectron: you mean just "Age 1" "Age 2" "Age 3"..?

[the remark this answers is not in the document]
paologum: yup, is there a reason we need to have the for if we just change it here?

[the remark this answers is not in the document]
JacquiFreeElectron: well at first, bc in the jar setting file the `_id` for age ruler is always "Age" (which means they are all the same), so I didn't think about using the id version and used 'for' bc the state name needs to be different. But then there was an issue that only one Age column was generated even if I added/selected two ages, so after encountering this issue I tried making the `_id` of each age column different and it worked. so I just left the implementation of 'for' in there. I can certainly change it if you want

app/src/settings_tabs/column_menu/ColumnMenu.tsx
```diff
@@ -97,11 +97,13 @@ type ColumnContentProps = {
 };
 const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) => {
   const { actions } = useContext(context);
+  const [ageSerialNum, setAgeSerialNum] = useState(1);
```
paologum, line 100: why did you revert back? this won't work because everytime this component reloads it resets at 1 ie
now both these ages are the same if i create, leave, and then come back and create another (they have the same id so they are both selected)

Screenshot 2024-07-23 at 1 14 26 PM

how you had it before was fine, it should just not take the `for ${name}` out when making id's if we don't need to. just so we can use the normal way so we don't have extra cases for no reason because the main intention for these serial numbers is for there to be unique ids right, so we shouldn't need to remove/append a for under certain cases

app/src/state/actions/column-actions.ts
```diff
@@ -569,19 +568,13 @@ export const addBlankColumn = action((column: ColumnInfo) => {
...
-  let serialNumber = 1;
-  const largestExistingSerialNum = column.children.findLastIndex(
-    (child) => /^Age \d+ for .+$/.test(child.name) && child.columnDisplayType === "Ruler"
-  );
-  if (largestExistingSerialNum > -1) {
-    serialNumber = findSerialNum(column.children[largestExistingSerialNum].name) + 1;
-  }
-  const ageColumnName = "Age " + `${serialNumber}` + " for " + column.name;
+  const ageColumnName = "Age " + `${serialNum}`;
+  console.log(ageColumnName);
```
paologum, line 577: log here

app/src/state/parse-settings.ts
```diff
@@ -466,6 +474,12 @@ export function translateColumnInfoToColumnInfoTSC(state: ColumnInfo): ColumnInf
     case "BlockSeriesMetaColumn":
       column._id = `class datastore.${state.columnDisplayType}:` + state.name;
       break;
+    case "Data": {
+      if (/^Blank \d+ for .+$/.test(state.name)) {
+        column._id = "class datastore.DataColumn:Blank " + findSerialNum(state.name);
+        break;
+      }
+    }
```
paologum, line 482: like the age, is there a reason we need to remove the for the column part? does it still generate correctly?
if it does then we should just leave it instead of casing it

app/src/settings_tabs/column_menu/ColumnMenu.tsx
JacquiFreeElectron: ok I get it now. sorry I misunderstood what you want me to do earlier. I’ll check if it’s going to work with the “for…” in the _id, if yes then I will keep it, if no then I will revert to what I have before

JacquiFreeElectron:
works fine with the "for..." in _id: 
Screenshot 2024-07-23 at 4 44 50 PM

paologum (approved):
looks good!

PR #188, opened 2024-07-24
Tsc 425 make chart generation errors print fatal

Changk820:
https://youtu.be/RSfqLPk34vs


server/src/graph-error-handler.ts
paologum, line 1: probably call it chart instead of graph

server/src/graph-error-handler.ts
```diff
@@ -0,0 +1,23 @@
...
+  "Premature end of file.",
+  "Content is not allowed in prolog."
+]);
+
+// this function checks if the line contains any of the known errors
+export function containsKnownError(line: string): boolean {
+  for (const error of knownErrors) {
+    if (line.includes(error)) {
+      return true;
+    }
```
paologum, line 20: Are these the full lines or only partial parts of the line

paologum (changes requested):

server/src/routes.ts
```diff
@@ -644,6 +645,9 @@ export const fetchChart = async function fetchChart(request: FastifyRequest, rep
     reply.send({ error: "ERROR: failed to save settings" });
     return;
   }
+  let errorFound = false; // error found during chart generation via java jar call
+  let errorMessage = ""; // error message found during chart generation via java jar call
```
paologum, line 649: if `errorMessage` existence is what matters we don't need two variables, you can just check `errorMessage` since empty strings are falsy

Changk820:
I optimized the code a bit, also I finally understand what you mean by partial/full messages, so I improved that part too.

Something to note: I added 1 line to java program to make it so if the java program can't make it to the finish line of generate chart, a print statement won't occur, thus we know for sure an error happened. But the issue is, I only added cases for if chart generation failed looking at the java program. See chart-error-handler.ts.

I also added some other errors cases that I know could potentially cause the program to crash/fail.

If we don't know the error, it will just be listed as unknown, but the chart generation will still stop as expected and we will tell the user.

One final note: We only know the actual issue if we have access to the server. The users will only see that chart generation failed, either because of bad database or settings. but they won't know which. I imagine this isn't very helpful to them?

If it's not, can we talk about how to do it Friday? I tried to get the actual message as a popup but had no success.

Otherwise, I'll just leave as it is.

app/src/state/actions/generate-chart-actions.ts
```diff
@@ -150,6 +150,13 @@ export const fetchChartFromServer = action("fetchChartFromServer", async (naviga
     });
     const answer = await response.json();
     // will check if svg is loaded
+    if (answer.error) {
+      // If the server sends an error, display it
+      displayServerError(answer, ErrorCodes.INVALID_CHART_FROM_JAR, answer.error);
+      generalActions.setChartLoading(false);
+      return;
+    }
```
paologum, line 158: you should use `isServerResponseError` and also passin `ErrorMessages[ErrorCodes.INVALID_CHART_FROM_JAR]` instead of `answer.error`
it will log the server error to console instead, so that errors will properly stack for now until we actually have the status codes setup in a future pr

paologum: for reference, this technically does work, but allocating all our error messages in one place, makes translation easier

paologum (approved):

PR #189, opened 2024-07-24
Tsc 433 wrap rotate text in columns

JacquiFreeElectron:
all Zone columns will have "orientation" radio group. the default value should be "normal", which is shown as "Horizontal" to users. the only exception is "Series Label", it should have "vertical" as its default orientation.

demo:

https://github.com/user-attachments/assets/164c0faf-5397-440d-9709-b4cd7002b6ec

block-Zone
chron label-Zone
facies label-Zone
members-Zone
series label-Zone
another sereis label-Zone


app/src/settings_tabs/column_menu/ColumnMenu.tsx
```diff
@@ -170,6 +183,17 @@ const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) =
...
+              <TSCRadioGroup
+                onChange={changeZoneColumnOrientation}
+                name={"Label Orientation"}
+                value={isZoneColumn()}
+                radioArray={[
+                  { value: "normal", label: "Horizontal" },
+                  { value: "vertical", label: "Vertical" }
+                ]}
+              />
+            )}
```
paologum, line 196: there are more block specific settings, so you should make a component in `advanced_settings` like `EventSpecificSettings` and `RangeSpecificSettings`,

additionally, this should be above the info box.
idk if i forgot to mention but the ruler age column above this should also be above the info box

app/src/settings_tabs/column_menu/ColumnMenu.tsx
```diff
@@ -116,6 +120,15 @@ const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) =
...
   };
+  const isZoneColumn = () => {
+    if (column.columnDisplayType !== "Zone") return "";
+    try {
+      assertZoneSettings(column.columnSpecificSettings);
+      return column.columnSpecificSettings.orientation;
+    } catch (e) {
+      return "";
+    }
+  };
```
paologum, line 131: when you implement it in the component have it just return null

app/src/state/parse-settings.ts
```diff
@@ -396,7 +397,16 @@ export function translateColumnInfoToColumnInfoTSC(state: ColumnInfo): ColumnInf
...
-      column = cloneDeep(defaultZoneColumnInfoTSC);
+      if (state.name != "Chart Root") {
+        assertZoneSettings(state.columnSpecificSettings);
+        column = {
+          ...cloneDeep(defaultZoneColumnInfoTSC),
+          orientation: state.columnSpecificSettings.orientation
+        };
+      } else {
+        column = cloneDeep(defaultZoneColumnInfoTSC);
+      }
```
paologum, line 408: i'm confused, `Chart Root` is a `RootColumn` not a `ZoneColumn`

paologum (changes requested):
just some small things but looks great so far

app/src/state/parse-settings.ts
JacquiFreeElectron: idk, one time when I ran `yarn test` it showed there's a Chart Root `ZoneColumn` and threw an error, and after I added the `if (state.name != "Chart Root") {...}` the error dissappeared. I can't remember what exactly the error was tho. I'll double check

JacquiFreeElectron: so there's a test has a chart root column with display type Zone, but the real chart root column doesn't have display type `Zone`. Do you want me to change that test?

paologum: the problem should then be that the test needs to have a valid `ZoneSettings` in `columnSpecificSettings`, not to ignore it so yes

paologum, line 408: you should revert this to not exclude `Chart Root`

shared/src/index.ts
```diff
@@ -633,6 +638,14 @@ export function assertRulerSettings(o: any): asserts o is RulerSettings {
...
     throwError("RulerSettings", "justification", "string and left | right", o.justification);
 }
+export function isZoneOrientation(o: any): o is ZoneOrientation {
+  return /^(normal|vertical)$/.test(o);
+}
+export function assertZoneSettings(o: any): asserts o is ZoneSettings {
+  if (!o || typeof o !== "object") throw new Error("ZoneSettings must be a non-null object");
+  if (typeof o.orientation !== "string" && isZoneOrientation(o.justification))
+    throwError("ZoneSettings", "orientation", "string and normal | vertical", o.orientation);
+}
```
paologum, line 648: `!isZoneOrientation`

paologum (changes requested):

app/src/state/parse-settings.ts
JacquiFreeElectron: oops I forgot

paologum:
the deployment has errors when i select a column

JacquiFreeElectron:
> the deployment has errors when i select a column

I mistakenly put `assertRulerSettings` before check if it is a Ruler column

paologum (approved):

PR #193, opened 2024-07-25
Tsc 408 read settings must support data mining columns

jay-jhk-lee:
Load settings now properly create datamining columns

https://github.com/user-attachments/assets/324e405f-6a42-4a45-8650-77cb081b34c5

known bug: addDataminingColumn doesn't go past 1 level of depth in some cases (ex. event then point)


app/src/state/actions/column-actions.ts
```diff
@@ -112,20 +113,114 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+  }
+  if (distanceFromInitial === 0) {
+    console.error("datamining column had wrong identifier", columnName);
+  } else {
+    const distanceBucket = state.settingsTabs.dataMiningColumnsCache.get(distanceFromInitial);
+    if (!distanceBucket) {
+      state.settingsTabs.dataMiningColumnsCache.set(distanceFromInitial, [settings]);
+    } else distanceBucket.push(settings);
+  }
+}
```
paologum, line 187: instead of buckets, why can't we just track all datamining columns in reading in some cache `foundCache` or something (where `isDataminingColumn=true`) that holds the settings

and also create a hash map (`shouldCache` or something) for columns that "should" exist ie `<setting name="drawExtraColumn">FAD</setting>`. which holds the ref column ( if it is a nested data mining it holds null, then when we attempt to iterate over it, we add the check (`IF NULL AND DOESN'T EXIST IN STATE { continue } `)

Then you loop over `shouldCache` after sorting it by string length, find the ref column and construct the string of the setting that should exist `FAD for ...` then see if it exists in `foundCache`, add the column, then apply the settings from the cache.

the approach above also seems more straightforward and doesn't rely on the name format as a signifier of existence and goes from a bottom up iteration

additionally any column that has `for` in it's name could disrupt the distanceBucket
whereas this solution would compensate by building up

app/src/state/actions/column-actions.ts
```diff
@@ -112,20 +113,114 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+      if (!column) {
+        console.log("WARNING: failed to get loaded datamining column");
+        continue;
+      }
+      setColumnProperties(column, settings);
+    }
+    distanceBucket = state.settingsTabs.dataMiningColumnsCache.get(++distanceFromInitial);
+  }
+  //reset cache
+  state.settingsTabs.dataMiningColumnsCache = new Map<number, PointColumnInfoTSC[]>();
```
paologum, line 166: is there a reason this cache needs to be in `state`? it technically doesn't refresh any components. it might be better to just make this a local variable that gets reset everytime we call `readSettings`
it might be better to have it as a `parse-settings.ts` artifact

app/src/state/actions/column-actions.ts
```diff
@@ -112,20 +113,114 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+      const parentCol = state.settingsTabs.columnHashMap.get(refCol.parent);
+      if (!parentCol) continue;
+      let addedColumnName: string | undefined = "";
+      switch (refCol.columnDisplayType) {
+        case "Event":
+          if (isEventFrequency(dataMiningType)) addedColumnName = addDataMiningColumn(refCol, dataMiningType);
+          break;
+        case "Chron":
+        case "Point":
+          if (isDataMiningPointDataType(dataMiningType)) addedColumnName = addDataMiningColumn(refCol, dataMiningType);
```
paologum, line 143: `Chron` although they share `Frequency` you should use the seperate functions for defining them to future proof this method

app/src/state/actions/column-actions.ts
```diff
@@ -112,20 +113,114 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
-      curcol = state.settingsTabs.columnHashMap.get(columnName + " " + extractName(settings.children[i]._id));
-      if (curcol !== undefined) setColumnProperties(curcol, settings.children[i]);
+      const child = settings.children[i];
+      curcol = state.settingsTabs.columnHashMap.get(columnName + " " + extractName(child._id));
+      if (curcol !== undefined) setColumnProperties(curcol, child);
+      //there is a datamining column
+      if (extractColumnType(child._id) === "PointColumn") {
+        assertPointColumnInfoTSC(child);
+        addColumnToDataMiningCache(child);
+      }
```
paologum, line 223: should only add if `isDataMiningColumn`

shared/src/settings-types.ts
```diff
@@ -260,7 +260,10 @@ export function assertEventColumnInfoTSC(o: any): asserts o is EventColumnInfoTS
     throwError("EventColumnInfoTSC", "rangeSort", "string", o.rangeSort);
   if (typeof o.windowSize !== "number") throwError("EventColumnInfoTSC", "windowSize", "number", o.windowSize);
   if (typeof o.stepSize !== "number") throwError("EventColumnInfoTSC", "stepSize", "number", o.stepSize);
-  if (o.drawExtraColumn != null && (typeof o.drawExtraColumn !== "string" || !isEventFrequency(o.drawExtraColumn)))
+  if (
+    o.drawExtraColumn != null &&
+    (typeof o.drawExtraColumn !== "string" || (!isEventFrequency(o.drawExtraColumn) && o.drawExtraColumn !== "OVERLAY"))
+  )
```
paologum, line 266: i'm thinking about this, and overlay should be it's own thing, in it's own tag because it functions different than a `dataMiningColumn`. This means that we'll have to change the java file but you can expect overlay to not be a part of the tag

you can keeep this here but add a todo to change it or you can remove it and just expect no overlay to be sent (although it technically can for a certain period, but we'll have tasks to remove it for the future)

app/src/state/parse-settings.ts
```diff
@@ -233,7 +233,13 @@ function processColumn(node: Element, id: string): ColumnInfoTSC {
         const child = <Element>maybeChild;
         const childName = child.getAttribute("id");
         if (child.nodeName === "column") {
-          column.children.push(processColumn(child, childName!));
+          let childColumn = processColumn(child, childName!);
+          //since isDataMiningColumn is an attribute, have to set it here
+          if (child.getAttribute("isDataMiningColumn") === "true") {
+            assertPointColumnInfoTSC(childColumn);
+            childColumn.isDataMiningColumn = true;
+          }
```
paologum, line 241: i feel like you should put this in the `processColumn`

paologum (changes requested):

app/src/state/actions/column-actions.ts
```diff
@@ -113,20 +114,168 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+
+const dataminingFoundCache = new Map<string, PointColumnInfoTSC>();
+
+//key: column name
+//currIdentifier: indicates the datamine column (if it exists) that references the key.
+//loadedIdentifier: the type of datamine column to draw according to the loaded settings
+const dataMiningRefCache = new Map<
+  string,
+  {
+    currIdentifier: EventFrequency | DataMiningChronDataType | DataMiningPointDataType | null;
```
paologum, line 126: this variable name kinda confused me, i would name it `existingDataMiningType` and `loadedDataMiningType`

app/src/state/actions/column-actions.ts
```diff
@@ -113,20 +114,168 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+            );
+            dataMiningRefCache.set(columnName, { currIdentifier: null, loadedIdentifier: settings.drawExtraColumn });
+            return;
+          }
+          assertPointSettings(column.columnSpecificSettings);
+          dataMiningRefCache.set(columnName, {
+            currIdentifier: column.columnSpecificSettings.dataMiningPointDataType,
+            loadedIdentifier: settings.drawExtraColumn
+          });
+        } else dataMiningRefCache.set(columnName, { currIdentifier: null, loadedIdentifier: settings.drawExtraColumn });
```
paologum, line 243: i feel like we don't need to log this difference and we can just do `if (column && column.columnDisplayType === "Point")`
that way we only set the cache once instead of twice with a return statement

paologum: additionally if they don't match i don't see why we should cache the column, at this point in the code or later when we add the column, the column still isn't gonna match so we are going to attempt to add a datamining column to a column that doesn't match. it should probably just skip

app/src/state/actions/column-actions.ts
```diff
@@ -113,20 +114,168 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+      //also could not be in selected datapacks, so commented since many of these columns could exist
+      //console.error("Datamining reference column is not in state");
+      continue;
+    }
+    if (!refInfo) {
+      console.error("While handling datamining columns, failed to find refInfo for reference column");
+      continue;
+    }
+    const { currIdentifier, loadedIdentifier } = refInfo;
+    const dmName = addDataMiningColumn(refCol, loadedIdentifier);
```
paologum, line 150: i would put this as a try catch because `loadedIdentifier` can be a bunch of different things and `refCol` can be a bunch of different things so there could be many errors if things get mismatched during a load
(just due to the combined assertion statements)

paologum (changes requested):

jay-jhk-lee:
Your suggestion for processColumn wouldn't work. The `column` variable is a placeholder variable (ColumnInfoTSC) that will be filled with the information from its child nodes. Once we enter the `processColumn` function for a particular column, we don't have access to its `isDataminingColumn` attribute since it is part of the column node itself. We either have to pass it through as a parameter, or set it like how I have it right now.

app/src/state/actions/column-actions.ts
```diff
@@ -113,20 +114,157 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+          assertEventSettings(column.columnSpecificSettings);
+          dataMiningRefCache.set(columnName, {
+            existingDataMiningType: column.columnSpecificSettings.frequency,
+            loadedDataMiningType: settings.drawExtraColumn
+          });
+        } else
+          dataMiningRefCache.set(columnName, {
+            existingDataMiningType: null,
+            loadedDataMiningType: settings.drawExtraColumn
+          });
```
paologum, line 200: can you give an example as to why we would want to keep it if the types are mismatched i'm still confused

app/src/state/actions/column-actions.ts
```diff
@@ -113,20 +114,184 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
```
paologum, line 209: hmm i'm still confused here do you have an example?
where the same column name exists and we would still want to store the null datamining column

app/src/state/actions/column-actions.ts
```diff
@@ -113,20 +114,184 @@ function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
...
+        //if loaded settings came from a different datapack
+        if (column && column.columnDisplayType !== "Event") {
+          console.log(
+            "WARNING: column in state and column in loaded settings have the same name ",
+            columnName,
+            "but have different types Event and ",
+            column.columnDisplayType
+          );
+          return;
+        }
```
paologum, line 198: i feel like we don't need to have a seperate log for if they have different types and the same name

it should really be up to the user to figure that out tbh because they loaded/created it wrong

this'll help clear complexity imo

paologum (approved):

PR #195, opened 2024-07-25
Tsc 388 public datapack upload

rupp3:
image
First screenshot shows what "username1" sees. The Malaysia datapack was uploaded publicly by "username" and nannos datapack was uploaded by "username1" 
image
Second screenshot shows "username" after uploading New Zealand datapack publicly


rupp3:
just realized public datapacks dont load until a user uploads their own datapack. also no clue why there are 35 commits with this, there should only be the last two, but i will also fix that

app/src/state/actions/general-actions.ts
```diff
@@ -206,6 +206,7 @@ export const fetchUserDatapacks = action("fetchUserDatapacks", async () => {
       credentials: "include"
     });
     const data = await response.json();
+    console.log("data in fetchUserdatapack general actions", data);
```
paologum, line 209: console

server/src/routes.ts
```diff
@@ -448,6 +476,49 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+    let publicMappacks: MapPackIndex = {};
+    let publicFiles: string[] = [];
+
+    try {
+      const publicData = await readFile(publicDatapackPath, "utf-8");
+      publicDatapacks = JSON.parse(publicData);
+      const publicMapData = await readFile(publicMappackPath, "utf-8");
+      publicMappacks = JSON.parse(publicMapData);
+      const publicFilesData = await readFile(publicFilesPath, "utf-8");
+      publicFiles = JSON.parse(publicFilesData);
```
paologum, line 497: need to `assert` here because you're reading from the files and these `JSON` could be outdated and would then technically be migrated but that's a different problem, but they need to be validated as the correct type

server/src/routes.ts
```diff
@@ -448,6 +476,49 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+      if ((e as NodeJS.ErrnoException).code != "ENOENT") {
+        await errorHandler("Failed to read public JSON files", 500, e);
+        return;
+      }
+    }
+    if (datapackIndex[filename]) {
+      publicDatapacks[filename] = datapackIndex[filename] as DatapackParsingPack;
+    }
+    if (mapPackIndex[filename]) {
+      publicMappacks[filename] = mapPackIndex[filename] as MapPack;
```
paologum, line 508: the above comment would remove the need for the casting

server/src/routes.ts
```diff
@@ -256,6 +259,27 @@ export const fetchUserDatapacks = async function fetchUserDatapacks(request: Fas
...
+      .then((data) => JSON.parse(data))
+      .catch(() => ({}));
+
+    const publicMapPacks: MapPackIndex = await readFile(path.join(publicDir, "MapPackIndex.json"), "utf8")
+      .then((data) => JSON.parse(data))
+      .catch(() => ({}));
+
+    const publicFiles = await readFile(path.join(publicDir, "publicFiles.json"), "utf8")
+      .then((data) => JSON.parse(data))
+      .catch(() => []);
```
paologum, line 273: need to `assert` here too

server/src/routes.ts
```diff
@@ -240,6 +242,7 @@ export const fetchUserDatapacks = async function fetchUserDatapacks(request: Fas
     return;
   }
   const userDir = path.join(assetconfigs.uploadDirectory, uuid);
+  const publicDir = path.join(assetconfigs.uploadDirectory, "public");
```
paologum, line 245: you should just make this in `assetconfigs` as the public directory

server/src/routes.ts
```diff
@@ -448,6 +476,49 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+    }
+    if (datapackIndex[filename]) {
+      publicDatapacks[filename] = datapackIndex[filename] as DatapackParsingPack;
+    }
+    if (mapPackIndex[filename]) {
+      publicMappacks[filename] = mapPackIndex[filename] as MapPack;
+    }
+    if (!publicFiles.includes(filename)) {
+      publicFiles.push(filename);
+    }
```
paologum, line 512: this check needs to happen earlier, and it needs to cleanup and throw error if they attempted to make it public and it already exists

paologum (changes requested):

paologum:
additionally, these public datapacks should be added into the logic of when we fetch server datapacks, because right now noone can view these datapacks/ the logic for providing them to all users (logged in and not) is not there

server/src/routes.ts
```diff
@@ -323,6 +348,9 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
   }
   const name = fields.name;
   const description = fields.description;
+  if (fields.isPublic === "true") {
+    isPublic = true;
+  }
```
paologum, line 353: one more thing instead of making `isPublic` be a `let` instead of `const`
to keep it in tune with the name and description you should
`const isPublic = fields.isPublic === "true"`

PR #197, opened 2024-07-27
Tsc 456 move dev branch to it s own env

paologum:


.github/workflows/deploy-dev.yml
```diff
@@ -0,0 +1,58 @@
...
+
+concurrency:
+  group: deploy-dev
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
asivath, line 11: make sure to remove when done testing

.github/workflows/deploy-dev.yml
```diff
@@ -0,0 +1,58 @@
...
+            echo 'DOMAIN="dev.timescalecreator.org"' >> ./server/.env
+
+            echo 'VITE_APP_URL=${{ vars.WEBSITE_URL }}' > ./app/.env
+            echo 'VITE_SERVER_URL=${{ vars.WEBSITE_URL }}' >> ./app/.env
+
+            if [ ! -f "docker-compose.yml" ]; then
+              echo "Error: docker-compose.yml not found"
+              exit 1
+            fi
+            docker-compose -f docker-compose.yml restart
```
asivath, line 58: since yarn install and build are not in entrypoint how do the files get rebuilt/installed if you don't manually tell the container to do that?

docker-compose.yml
```diff
@@ -19,5 +19,9 @@ services:
...
     volumes:
       - .:/code
+      - yarn_cache:/root/.yarn
     networks:
-      - shared_network
\ No newline at end of file
+      - shared_network
+    
+volumes:
+  yarn_cache:
```
asivath, line 27: yarn build currently cleans the entire cache, if the point of this is to speed up build then that should probably be removed

asivath (commented):
if not using the deploy script then you should probably remove it, though the point of that was the pseudo terminal was behaving weird with docker compose commands like always succeeding or exiting early for the commands so just make sure that doesn't happen

docker-compose.yml
asivath: wait nvm thinking of something else

.github/workflows/deploy-dev.yml
paologum: i added it in the `docker-compose.yml`

paologum: git says it might be faster, but it doesn't really look like it is. i think i'll leave it, just for the case that someone is setting up the build without an action, which means it won't be built or installed which is slightly confusing for someone unless we had proper documentation

paologum: also one more thing, your code for the preview in checking if the server is listening would probably be good for both dev and production so i'm making a task for that

Dockerfile
```diff
@@ -19,6 +19,8 @@ Run apt-get install -y xterm
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh | bash 
 # Note: you need the --login option and the PS1 variable to be set in order for /root/.bashrc to run properly
 RUN bash --login -c "export PS1=':' && source /root/.bashrc && nvm install node && nvm use node && npm install -g yarn"
+RUN bash --login -c "export PS1=':' && source /root/.bashrc && yarn install"
+RUN bash --login -c "export PS1=':' && source /root/.bashrc && yarn build"
```
asivath, line 23: i think volume mounting happens at run time which is why the deploy failed, need to copy files or do something else

.github/workflows/deploy-dev.yml
asivath: what did you add to the compose file that does that stuff?

Dockerfile
paologum: ok i'll just do it in entrypoint then

.github/workflows/deploy-dev.yml
paologum: whoops i mean `DockerFile`

COPY package.json yarn.lock /code/
```
# Install yarn/npm/node via nvm
RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh | bash 
# Note: you need the --login option and the PS1 variable to be set in order for /root/.bashrc to run properly
RUN bash --login -c "export PS1=':' && source /root/.bashrc && nvm install node && nvm use node && npm install -g yarn"
RUN bash --login -c "export PS1=':' && source /root/.bashrc && yarn install"
COPY . /code/
RUN bash --login -c "export PS1=':' && source /root/.bashrc && yarn build"
```

Dockerfile
paologum: nvm i kept it and manually mounted which other people online seem to do

docker-compose.yml
```diff
@@ -19,5 +20,9 @@ services:
       - "5800:5800"
     volumes:
       - .:/code
+      - yarn_cache:/root/.yarn
```
asivath, line 23: shouldn't this be /code/.yarn/cache

asivath (commented):
something you could consider is also making yarn install run in production mode so it doesn't download dev dependencies. idk if you want to do it through compose and an environment variable or a different command that the entrypoint script could use.

asivath (approved):

PR #200, opened 2024-07-29
Tsc 448 sort maps so that parent maps are at the top

JacquiFreeElectron:
Screenshot 2024-07-29 at 1 00 14 AM

world map always be the first, then all parents map shown in the alphabetic order, then their children

### DIFF
--- a/app/src/settings_tabs/map_points/MapPoints.tsx
+++ b/app/src/settings_tabs/map_points/MapPoints.tsx
@@ -80,7 +80,7 @@ const MapList: React.FC<MapRowComponentProps> = observer(({ mapInfo }) => {
       <Box>
         <List>
           {Object.entries(sortedMapInfoObj).map(([name, map]) => {
-            console.log("name: " + name, " ,parent: " + map.parent?.name)
+            console.log("name: " + name, " ,parent: " + map.parent?.name);
             return (
               <MapListItemButton
                 disableRipple
--- a/app/src/settings_tabs/map_points/MapPoints.tsx
+++ b/app/src/settings_tabs/map_points/MapPoints.tsx
@@ -2,7 +2,7 @@ import { useContext } from "react";
 import { context } from "../../state";
 import { useTheme } from "@mui/material/styles";
 import { Typography, Dialog, List, Box, ListItemButton, ListItemAvatar, ListItemText, Avatar } from "@mui/material";
-import type { MapInfo } from "@tsconline/shared";
+import { assertMapInfo, type MapInfo } from "@tsconline/shared";
 import { styled } from "@mui/material/styles";
 import { devSafeUrl } from "../../util";
 import { observer } from "mobx-react-lite";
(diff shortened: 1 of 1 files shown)


app/src/settings_tabs/map_points/MapPoints.tsx
```diff
@@ -52,12 +52,34 @@ const MapList: React.FC<MapRowComponentProps> = observer(({ mapInfo }) => {
...
+  // then their children
+  for (const parent of sortedMapHierarchyKeys) {
+    for (const child of state.mapState.mapHierarchy[parent]) {
+      if (!sortedMapEntries.includes([child, mapInfo[child]]) && mapInfo[child]) {
+        sortedMapEntries.push([child, mapInfo[child]]);
+      }
+    }
+  }
+  const sortedMapInfoObj = Object.fromEntries(sortedMapEntries);
+  assertMapInfo(sortedMapInfoObj);
```
paologum, line 77: just so we make sure things don't crash when someone edits this, can you define the types as
`const sortedMapEntries: [string, MapInfo[string][] = []`
and
`const sortedMapInfoObj: MapInfo`
if that doesn't work the asserts is fine though

JacquiFreeElectron: `const sortedMapInfoObj: MapInfo` works, the other one doesn't

paologum: hmm it might be Array<[string, MapInfo[string]]>
try that?

app/src/settings_tabs/map_points/MapPoints.tsx
```diff
@@ -52,12 +52,34 @@ const MapList: React.FC<MapRowComponentProps> = observer(({ mapInfo }) => {
...
+  // then their children
+  for (const parent of sortedMapHierarchyKeys) {
+    for (const child of state.mapState.mapHierarchy[parent]) {
+      if (!sortedMapEntries.includes([child, mapInfo[child]]) && mapInfo[child]) {
+        sortedMapEntries.push([child, mapInfo[child]]);
+      }
+    }
+  }
+  const sortedMapInfoObj: MapInfo = Object.fromEntries(sortedMapEntries);
+  assertMapInfo(sortedMapInfoObj);
```
paologum, line 77: if you declare `: MapInfo`
you already know that it has to be `MapInfo` so you don't need to `assert`

JacquiFreeElectron: ok

app/src/settings_tabs/map_points/MapPoints.tsx
JacquiFreeElectron: ok

JacquiFreeElectron: `Array<[string, MapInfo[string]]>` works

paologum (approved):

PR #203, opened 2024-07-29
Tsc 461 event search change status to checkbox

jay-jhk-lee:
Changed logic of search tab
status column now has a checkbox or error icon
checkbox only shows up if age is within time interval and can be used to toggle columns
error icon shows up if age is not within time interval

https://github.com/user-attachments/assets/99563c7a-b7e6-4518-ae02-f67af5a0dd3b

image


app/src/settings_tabs/Results.tsx
```diff
@@ -202,6 +149,7 @@ const Center = observer(({ info }: { info: EventSearchInfo }) => {
   }
   const centerTimeOnEvent = () => {
     const ages = verifyAgesAndAddAgeMargin(info.age);
+    console.log(ages);
```
paologum, line 152: log

paologum (changes requested):
one thing

paologum (approved):

PR #205, opened 2024-07-29
Tsc 467 make deploy actions listen to docker logs to return errors properly

paologum:
listen to docker logs for if the server restarts/starts properly

add some proper error checking for if `/confirm-deploy` gets called by itself and more specific docker querying for current running containers

not sure why but it doesn't use my workflow file when commenting so i can't for sure test it until it merges to main
although i tested all the commands in the console (and checked that it deploys/listens when deploying to development)

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -3,77 +3,8 @@ name: Deploy Preview
 on:
   issue_comment:
     types: [created]
-  workflow_dispatch:
-    inputs:
-      pr_number:
-        description: 'PR number'
-        required: true
-        default: ''
 
 jobs:
-  test-deploy:
-    if : ${{ !github.event.issue.pull_request }}
-    runs-on: ubuntu-latest
-    steps:
-      - name: Set up SSH key
-        uses: webfactory/ssh-agent@v0.5.3
-        with:
-          ssh-private-key: ${{ secrets.SSH_KEY }}
-
-      - name: Add known hosts
-        run: |
-          mkdir -p ~/.ssh
-          ssh-keyscan -H dev.geolex.org >> ~/.ssh/known_hosts
-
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy-preview.yml
```diff
@@ -170,19 +185,42 @@ jobs:
...
+              echo "Stopping and removing container $RUNNING_CONTAINER"
+              docker stop $RUNNING_CONTAINER
+              docker rm $RUNNING_CONTAINER
+            fi 
+            docker-compose -f docker-compose-preview.yml up -d --remove-orphans pr-preview-build
+            BUILD_CONTAINER_ID=$(docker-compose -f docker-compose-preview.yml ps -q pr-preview-build)
+            if [ -z "$BUILD_CONTAINER_ID" ]; then
+              echo "Error: Failed to get container ID for pr-preview-build."
+              exit 1
+            fi
```
asivath, line 205: it should be built at this point so why build again?

asivath (commented):
the reason your file is not used is because actions that react to comments only work if they're in the main branch. you should probably create a fork and merge your changes to test.

.github/workflows/deploy-preview.yml
paologum: because the running container includes all other pr-preview-builds including it's own if it is still up so i rebuilt it if it took it down if you `/confirm-deploy` and the container id wouldn't show up
i can just change it so that it only looks for `pr-preview-run` so that it doesn't take down the build

paologum: also, i'd rather just debug on main so i don't have to fork because the setup is killing me and it's the same code as the deploy stuff

paologum:
confirmed to work on fork

asivath (approved):

PR #209, opened 2024-08-01
Tsc 403 coverage report enforcement

asivath:
You can see what the coverage comment looks like: https://github.com/earthhistoryviz/tsconline/pull/208


.github/workflows/merge-pipeline.yml
```diff
@@ -172,45 +174,63 @@ jobs:
...
       - name: Run Tests
         id: run_tests
         run: |
           Xvfb :1 -screen 0 1024x768x16 &
           export DISPLAY=:1
-          set -o pipefail
-          if ! yarn test > test_output.log 2>&1; then
-              echo "test_failed=true" >> $GITHUB_OUTPUT
-              fi
+          yarn coverage:ci > test_output.log 2>&1 || (grep -q 'ERROR: Coverage' test_output.log && echo "coverage_failed=true" >> $GITHUB_OUTPUT) || echo "test_failed=true" >> $GITHUB_OUTPUT
```
paologum, line 202: does `yarn coverage` not have an exit code we can use to verify it failed instead of greping

vite.config.ts
```diff
@@ -1,14 +1,34 @@
 import { fileURLToPath, URL } from "url";
 import { defineConfig } from "vitest/config";
 
+const config = {
+  statements: 80,
+  branches: 80,
+  functions: 80,
+  lines: 80
+};
+
```
paologum, line 10: i was thinking about this, and since we have such a high test coverage on the files below, i think we should have it 90 so that anyone updating these files will make tests immediately, rather than maybe squeaking by since some are 90~100 and small functions or whatever can get by.

except `encryption.ts` because it has like three lines it doesn't account for but you can do your exception like you do for it already or add a test for it

paologum (changes requested):

.github/workflows/merge-pipeline.yml
asivath: vitest doesn't provide different error codes or a way to actually edit that so I don't know of another way to do this

paologum (approved):

PR #212, opened 2024-08-05
Tsc 452 webworker for setting columns datapack config (refactoring for datapack selection)

JacquiFreeElectron:
I'll upload the demo tmr, for now please use `/deploy ` 😬


paologum:
//deploy

paologum:
there is a delay i put on clicking the plus icon on the `Datapacks` page so the loading ui starts up a little bit before the overhead of the big function call takes place. you should remove that so that it just clicks on now

JacquiFreeElectron:
demo: (haven't removed the loading delay yet)

https://github.com/user-attachments/assets/d61934e0-419a-466e-9e4f-b92b903ca871

app/src/state/initialize.ts
```diff
@@ -1,11 +1,44 @@
...
+          value.mapHierarchy,
+          value.mapInfo,
+          value.datapacks,
+          value.chartSettings
+        );
+      } else {
+        actions.pushSnackbar("Setting Datapack Config Timed Out", "info");
+      }
+      setDatapackConfigWorker.terminate();
+    };
```
paologum, line 40: you should create a function that wraps this into a function that returns a promise
resolve/reject on a message/error
accepting params of `datapacks:[]` and `chartSettings`

need an `onerror` which should `terminate` as well

app/src/state/actions/general-actions.ts
```diff
@@ -421,117 +424,170 @@ const setPreviousDatapackConfig = action("setPreviousDatapackConfig", (datapacks
...
+);
+
 export const setDatapackConfig = action(
   "setDatapackConfig",
-  async (datapacks: string[], settingsPath?: string): Promise<boolean> => {
-    if (state.datapackCachedConfiguration.has(datapacks.join(","))) {
-      setPreviousDatapackConfig(datapacks);
-      return true;
-    }
+  (datapacks: string[], chartSettings: ChartInfoTSC | null) => {
```
paologum, line 490: we need this function? we have a webworker now for it

app/src/state/actions/general-actions.ts
```diff
@@ -421,117 +424,170 @@ const setPreviousDatapackConfig = action("setPreviousDatapackConfig", (datapacks
...
+        removeError(ErrorCodes.INVALID_SETTINGS_RESPONSE);
+        chartSettings = JSON.parse(JSON.stringify(settings));
+      } else {
+        pushError(ErrorCodes.INVALID_SETTINGS_RESPONSE);
+        return false;
+      }
+    }
+  }
+);
+
```
paologum, line 487: this should either just be a setter that sets the `chartSettings` in `state` or actually fetches the `settingsPath` with a route and sets it. otherwise this function doesn't make sense

app/src/state/actions/general-actions.ts
```diff
@@ -421,117 +424,170 @@ const setPreviousDatapackConfig = action("setPreviousDatapackConfig", (datapacks
...
+export const afterSetDatapackConfig = action(
+  "afterSetDatapackConfig",
+  async (
+    columnRoot: ColumnInfo,
+    foundDefaultAge: boolean,
+    mapHierarchy: MapHierarchy,
+    mapInfo: MapInfo,
+    datapacks: string[],
+    chartSettings: ChartInfoTSC | null
```
paologum, line 588: you should just call this `setDatapackConfig` and the actual overhead function as `processDatapackConfig`

paologum: this function also does not need to assert because we declare types

app/src/settings_tabs/Datapack.tsx
```diff
@@ -15,27 +15,50 @@ import { TSCDatapackRow } from "../components/datapack_display/TSCDatapackRow";
...
         <CustomTooltip title="Deselect All" placement="top">
-          <IconButton className={styles.ib} onClick={() => actions.setDatapackConfig([], "")}>
+          <IconButton
+            className={styles.ib}
+            onClick={async () => {
+              actions.setSelectedDatapacks([]);
+              const finished = await actions.processSelectedDatapackList([], null);
+              if (finished) {
+                setProcessing(false);
+              }
```
paologum, line 60: need to await if you're using actions

app/src/settings_tabs/Datapack.tsx
```diff
@@ -15,27 +15,50 @@ import { TSCDatapackRow } from "../components/datapack_display/TSCDatapackRow";
 import DeselectIcon from "@mui/icons-material/Deselect";
 import ViewCompactIcon from "@mui/icons-material/ViewCompact";
 import { TSCCompactDatapackRow } from "../components/datapack_display/TSCCompactDatapackRow";
+import { TSCLoadingDatapacks } from "../components/TSCLoadingDatapacks";
 
 export const Datapacks = observer(function Datapacks() {
   const { actions } = useContext(context);
   const [formOpen, setFormOpen] = useState(false);
+  const [processing, setProcessing] = useState(false);
```
paologum, line 23: as said before, should be a `state` variable

app/src/settings_tabs/Datapack.tsx
```diff
@@ -72,37 +95,70 @@ export const Datapacks = observer(function Datapacks() {
...
             <TSCDatapackCard
               key={datapack}
               name={datapack}
               datapack={state.datapackIndex[datapack]}
-              value={state.config.datapacks.includes(datapack)}
+              value={
+                (state.datapackSelection.selectedDatapacks.includes(datapack) ||
+                  state.config.datapacks.includes(datapack)) &&
+                !state.datapackSelection.unselectedDatapacks.includes(datapack)
+              }
```
paologum, line 126: explain?

paologum (changes requested):
pretty confused overall
mainly the implementation should be like this

there should be a dialog in `App.tsx` that opens when `state.isProcessingDatapackConfig` is `true`

We should wrap the webworker work into a promise function that way we can just call it like `await setDatapackConfig` or ig like `await processDatapackConfig` and then set it using the setter function. This will reduce the amount of duplicated code a lot, because it essentially shouldn't be a "new" function. All the `onClick` or component handlers that we pass should really call one function and maybe have one or two other lines besides that

Another big point like i commented is that it shouldn't have cache because that was a temp fix and is not memory efficient

app/src/components/TSCConfirmSelectionChange.tsx
paologum, line 1: empty file

app/src/settings_tabs/Datapack.tsx
```diff
@@ -72,37 +95,70 @@ export const Datapacks = observer(function Datapacks() {
...
+              state.datapackSelection.unselectedDatapacks.length > 0
+                ? state.datapackSelection.selectedDatapacks.concat(
+                    state.config.datapacks.filter(
+                      (datapack) => !state.datapackSelection.unselectedDatapacks.includes(datapack)
+                    )
+                  )
+                : state.datapackSelection.selectedDatapacks.concat(state.config.datapacks);
+            const finished = await actions.processSelectedDatapackList(datapacks, null);
+            if (finished) {
+              setProcessing(false);
```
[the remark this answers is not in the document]
paologum: this should also popup when someone clicks a preset configuration, because there is a case where a large datapack configuration is selected and if that isn't done configuring, the user can add more datapacks/queue more in the settings if we don't popup the loader

JacquiFreeElectron:
> pretty confused overall mainly the implementation should be like this
> 
> there should be a dialog in `App.tsx` that opens when `state.isProcessingDatapackConfig` is `true`
> 
> We should wrap the webworker work into a promise function that way we can just call it like `await setDatapackConfig` or ig like `await processDatapackConfig` and then set it using the setter function. This will reduce the amount of duplicated code a lot, because it essentially shouldn't be a "new" function. All the `onClick` or component handlers that we pass should really call one function and maybe have one or two other lines besides that
> 
> Another big point like i commented is that it shouldn't have cache because that was a temp fix and is not memory efficient

ok, so three major changes: shouldn't use `state.datapackCachedConfiguration` anymore, wrap the webworker work into a promise function, change the place of the processing datapacks dialog to `App.tsx`.

paologum:
> > pretty confused overall mainly the implementation should be like this
> > there should be a dialog in `App.tsx` that opens when `state.isProcessingDatapackConfig` is `true`
> > We should wrap the webworker work into a promise function that way we can just call it like `await setDatapackConfig` or ig like `await processDatapackConfig` and then set it using the setter function. This will reduce the amount of duplicated code a lot, because it essentially shouldn't be a "new" function. All the `onClick` or component handlers that we pass should really call one function and maybe have one or two other lines besides that
> > Another big point like i commented is that it shouldn't have cache because that was a temp fix and is not memory efficient
> 
> ok, so three major changes: shouldn't use `state.datapackCachedConfiguration` anymore, wrap the webworker work into a promise function, change the place of the processing datapacks dialog to `App.tsx`.

yes, also add logic for preset clicking and `await`ing the `processDatapackConfig`
and abstracting the configuration to just a `string[]` unless i'm misunderstanding the approach

JacquiFreeElectron:
> setDatapackConfig

by "abstracting the configuration to just a string[]" you mean which configuration? The `state.config`?

JacquiFreeElectron:
> > > pretty confused overall mainly the implementation should be like this
> > > there should be a dialog in `App.tsx` that opens when `state.isProcessingDatapackConfig` is `true`
> > > We should wrap the webworker work into a promise function that way we can just call it like `await setDatapackConfig` or ig like `await processDatapackConfig` and then set it using the setter function. This will reduce the amount of duplicated code a lot, because it essentially shouldn't be a "new" function. All the `onClick` or component handlers that we pass should really call one function and maybe have one or two other lines besides that
> > > Another big point like i commented is that it shouldn't have cache because that was a temp fix and is not memory efficient
> > 
> > 
> > ok, so three major changes: shouldn't use `state.datapackCachedConfiguration` anymore, wrap the webworker work into a promise function, change the place of the processing datapacks dialog to `App.tsx`.
> 
> yes, also add logic for preset clicking and `await`ing the `processDatapackConfig` and abstracting the configuration to just a `string[]` unless i'm misunderstanding the approach

by "abstracting the configuration to just a string[]" you mean which configuration? The state.config?

app/src/state/state.ts
```diff
@@ -203,5 +209,11 @@ export const state = observable<State>({
...
   presetColors: JSON.parse(localStorage.getItem("savedColors") || JSON.stringify(defaultColors)),
   snackbars: [],
-  datapackCachedConfiguration: new Map<string, CachedConfig>()
+  datapackCachedConfiguration: new Map<string, CachedConfig>(),
+  datapackSelection: {
+    regenerateChart: false,
+    isDirty: false,
+    selectedDatapacks: [],
+    unselectedDatapacks: []
+  }
```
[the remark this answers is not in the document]
JacquiFreeElectron: I probably overcomplicate this, I will check if only a `selectedDatapacks` list is needed

app/src/state/actions/general-actions.ts
JacquiFreeElectron: this is for the datapack selection I think. I can try if after refactoring the webworker will be more efficient/stable when selecting multiple datapacks, if yes then I will remove this.

app/src/settings_tabs/Datapack.tsx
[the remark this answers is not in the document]
JacquiFreeElectron: Screenshot 2024-08-05 at 6 22 19 PM
Is this the thing you're talking about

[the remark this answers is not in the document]
paologum: i mean the generate button, when it is clicked it sets the datapack config first and then generates the chart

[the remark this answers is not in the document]
JacquiFreeElectron: yeah actually I just checked the logic is needed for this add button as well, otherwise the datapack config won't update. I will add logic for this whole unit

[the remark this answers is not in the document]
paologum: it technically was never there to begin with but if you want to add it in feel free to 👍

JacquiFreeElectron:
new demo: 

https://github.com/user-attachments/assets/aaf930b3-4798-4333-8447-3c84987040e5

app/src/NavBar.tsx
```diff
@@ -34,6 +35,13 @@ export const NavBar = observer(function Navbar() {
   const { anchorProps, hoverProps } = useHover(settingsMenuState.state, settingsMenuToggle);
 
   const location = useLocation();
+  const checkOpen = () => {
+    // need to use state.datapackSelection.regenerateChart to check if the user is trying to generate a chart without confirming selections
+    return (
+      state.datapackSelection.regenerateChart &&
+      JSON.stringify(state.datapackSelection.selectedDatapacks) !== JSON.stringify(state.config.datapacks)
+    );
+  };
```
paologum, line 44: i don't think you should go about opening the popup here, there are too many spread out parts of this functionality compared to the way we do the age popup
use a `state` variable more aptly named like `showConfirmDatapackSelectionPopup` and set this when `initiateChartGeneration` is called when `state.datapackSelection.selectedDatapacks !== state.config.datapacks`

it would have it's own condition before the age popup part of that function `initiateChartGeneration`

then you would just have that variable for opening in the `TSCDialog` here where the `onYes` `onNo` `onClose` all set it properly

paologum: although from a user perspective, if they selected it, then pressed the generate button they probably want to keep that selection. and they're probably ready for a loading process as well knowing that. so you instead of having a popup for this, you should just set it for them (which would put up the loading screen) and then generate normally. (disregard previous comment then)

app/src/NavBar.tsx
```diff
@@ -118,9 +126,32 @@ export const NavBar = observer(function Navbar() {
...
+          buttonType="gradient"
+          onClick={() => {
+            actions.initiateChartGeneration(navigate, location.pathname);
+          }}>
           Generate Chart
         </TSCButton>
+        <TSCPopupDialog
+          open={checkOpen()}
+          title="Confirm Datapack Selection Change"
+          message="You have unsaved change on datapack selection. Do you want to save the change?"
```
paologum, line 139: try
`You have unsaved changes! If you leave now, your changes will not be saved.`

app/src/Settings.tsx
```diff
@@ -32,6 +34,12 @@ export const Settings = observer(function Settings() {
   const tabs = Object.values(SettingsMenuOptionLabels).map((val) => ({ id: val, tab: val }));
   const tabKeys = Object.keys(SettingsMenuOptionLabels);
   const tabIndex = tabKeys.indexOf(state.settingsTabs.selected);
+  const checkOpen = () => {
+    return (
+      state.settingsTabs.selected != "datapacks" &&
+      JSON.stringify(state.config.datapacks) !== JSON.stringify(state.datapackSelection.selectedDatapacks)
+    );
+  };
```
paologum, line 42: this is a boolean rather than a function

app/src/settings_tabs/Datapack.tsx
```diff
@@ -15,27 +15,34 @@ import { TSCDatapackRow } from "../components/datapack_display/TSCDatapackRow";
...
   return (
     <div className={styles.dc}>
       <div className={styles.hdc}>
         <CustomTooltip title="Deselect All" placement="top">
-          <IconButton className={styles.ib} onClick={() => actions.setDatapackConfig([], "")}>
+          <IconButton
+            className={styles.ib}
+            onClick={async () => {
+              actions.setSelectedDatapacks([]);
+              await processDatapackConfig([]);
```
paologum, line 44: same thing here with combining the two so it's easier

app/src/state/state.ts
```diff
@@ -203,5 +206,9 @@ export const state = observable<State>({
   },
   presetColors: JSON.parse(localStorage.getItem("savedColors") || JSON.stringify(defaultColors)),
   snackbars: [],
-  datapackCachedConfiguration: new Map<string, CachedConfig>()
+  isProcessingDatapacks: false,
+  datapackSelection: {
+    regenerateChart: false,
+    selectedDatapacks: []
```
paologum, line 212: from previous comment you won't need `regenerateChart` but i would change `selectedDatapacks` to `unsavedDatapackConfig` to remove any confusion between the two

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,130 @@
...
+import { cloneDeep } from "lodash";
+
+/**
+ * sets chart to newval and requests info on the datapacks from the server
+ */
+self.onmessage = async (e: MessageEvent<SetDatapackConfigMessage>) => {
+  const { datapacks, chartSettings, stateCopy } = e.data;
+  let stateCopyObj: State;
+  try {
+    stateCopyObj = JSON.parse(stateCopy);
```
paologum, line 13: i would create an asserts specifically that `assertsState` for at the bare minimum the fields you need
beecause otherwise, we are "hoping" that the `stateCopy` is of the right type 

this will catch errors for `JSON` parsing but will not catch whether `stateCopy` is of the right type

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,130 @@
...
+      }
+      const mapPack = stateCopyObj.mapPackIndex[datapack]!;
+      if (!mapInfo) mapInfo = mapPack.mapInfo;
+      else Object.assign(mapInfo, mapPack.mapInfo);
+      if (!mapHierarchy) mapHierarchy = mapPack.mapHierarchy;
+      else Object.assign(mapHierarchy, mapPack.mapHierarchy);
+    }
+    // makes sure things are named correctly for users and for the hash map to not have collisions
+    for (const [unit, column] of unitMap) {
+      //await new Promise((resolve) => setTimeout(resolve, 0));
```
paologum, line 86: remove

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,130 @@
...
+      expanded: true
+    };
+    // all chart root font options have inheritable on
+    for (const opt in columnRoot.fontsInfo) {
+      columnRoot.fontsInfo[opt as keyof FontsInfo].inheritable = true;
+    }
+    // add everything together
+    // uses preparsed data on server start and appends items together
+    for (const datapack of datapacks) {
+      //await new Promise((resolve) => setTimeout(resolve, 0));
```
paologum, line 54: remove

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,130 @@
...
+  });
+  const message: SetDatapackConfigCompleteMessage = { status: "success", value: undefined };
+  async function runWithTimeout() {
+    try {
+      await Promise.race([SetDatapackConfig(), timeoutPromise]);
+      if (SetDatapackConfig()) {
+        message.value = SetDatapackConfig();
+      } else {
+        message.status = "failure";
+      }
```
paologum, line 122: you call `SetDatapackConfig()` three times also change it to camel case (`setDatapackConfig`)
then you can do `const result = await Promise...` and use the result to set the value

app/src/util/workers/set-datapack-config.ts
paologum: but i mean if you sent the actual state object and inferred it with `SetDatapackConfigMessage`, then it wouldn't need an `asserts` so could you explain why it needs to be a string

app/src/process-datapack-config.ts
```diff
@@ -0,0 +1,75 @@
...
+};
+
+export async function processDatapackConfig(
+  datapacks: string[],
+  settingsPath?: string,
+  presetChartGeneration?: InitialChartGeneration
+) {
+  if (state.isProcessingDatapacks) {
+    return;
+  } else {
```
paologum, line 17: if we return, we don't need an else

app/src/process-datapack-config.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (presetChartGeneration) {
+        actions.initiateChartGeneration(presetChartGeneration.navigate, presetChartGeneration.path);
+        actions.setRegenerateChart(false);
+      }
+    };
+    setDatapackConfigWorker.onerror = function (error) {
+      setDatapackConfigWorker.terminate();
+      console.error("Webworker failed with error: " + error);
+    };
+  }
```
paologum, line 74: this needs to be wrapped in a `Promise` where a successful message `resolve()` and an error `reject()`

app/src/process-datapack-config.ts
```diff
@@ -0,0 +1,75 @@
...
+    setDatapackConfigWorker.onmessage = async function (e: MessageEvent<SetDatapackConfigCompleteMessage>) {
+      const { status, value } = e.data;
+      if (status === "success" && value) {
+        await actions.setDatapackConfig(
+          value.columnRoot,
+          value.foundDefaultAge,
+          value.mapHierarchy,
+          value.mapInfo,
+          value.datapacks,
+          value.chartSettings
```
paologum, line 57: we set the `datapacks` here, so you shouldn't need to set it before you call this function like the other times. if your thought process was that they should see the change, they probably shouldn't bc if it fails it would be misleading

paologum (changes requested):

app/src/process-datapack-config.ts
paologum: this is needed because otherwise this function returns before the webworker is done, making the function operate oddly since we don't necessarily know from the function finishing whether there was an error or not which is misleading. after the webworker promise we create, you should call `actions.setIsProcessDatapacks(false)` after it (since it needs to be called after an error/success

paologum: just to make sure it makes sense, rn the execution is 

call function -> start webworker -> finish function -> finish webworker -> see changes

whereas we want

call function -> start webworker -> finish webworker -> finish function -> see changes

this might look like it disregards the new thread that the webworker makes but if we wrap it in a promise it acts as just another async function with the added benefit of no extra thread blockage (we just return to it every cycle to check if it is done)

app/src/Home.tsx
```diff
@@ -97,15 +100,12 @@ const TSCPresetHighlights = observer(function TSCPresetHighlights({
...
                     preset={preset}
                     generateChart={async () => {
-                      const success = await actions.setDatapackConfig(
-                        preset.datapacks.map((datapack) => datapack.file),
-                        preset.settings
+                      actions.setSelectedDatapacks(preset.datapacks.map((datapack) => datapack.file)); //make sure seletedDatapacks gets updated
+                      await processDatapackConfig(
+                        JSON.parse(JSON.stringify(state)).datapackSelection.selectedDatapacks,
+                        preset.settings,
+                        { navigate: navigate, path: "/home" }
```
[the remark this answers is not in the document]
JacquiFreeElectron: Use string copy because webworker can’t access to state object copy

app/src/state/state.ts
JacquiFreeElectron: I think we should do selectedDatapacks because when user click generate the selectedDatapacks will get updated and it’s kinda untrue to call it unsaved

app/src/process-datapack-config.ts
JacquiFreeElectron: you mean I should set/update the `selectedDatapacks[]` in here? sorry I kinda lost

app/src/state/state.ts
JacquiFreeElectron: Or maybe call it something like "datapackSelectionBuffer"? to indicate that this is different from `datapack.config.datapacks[]` and save&generate&select can all change it

app/src/process-datapack-config.ts
JacquiFreeElectron: got it. thanks

app/src/Home.tsx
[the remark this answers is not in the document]
paologum: ok if that's the case you'll want to assert the necessary fields

app/src/NavBar.tsx
paologum: the two options should probably be `Leave Without Changes` and `Cancel` where `Leave` is the primary button

It should also probably be in `App.tsx` as well since we want to have this condition to open the popup
```
location !== /settings && state.settingsTabs.selected != "datapacks" && datapacks != unSavedConfig
```

JacquiFreeElectron: okok ty

app/src/state/state.ts
paologum: well in naming this variable we have to distinctly seperate anyone from thinking that this might hold a meaningful state that contributes towards the actual datapack configuration that the state is based on. So in the var name there should be some dissonance that reflects that this variable is not something that should be the main focus, so in this case i think `unsavedDatapackConfig` would reflect that. If we were to call it buffer it would still be ambiguous to the point where they would have to read code more to get a better understanding.

technically all of the buttons can change this, but that doesn't change the fact that if it differs from the `config.datapacks` it is not reflected in state. once they are the same, then it's easily understandable this version is `saved` 
if you don't like `saved` i can see `draftDatapackConfig` as well



> I think we should do selectedDatapacks because when user click generate the selectedDatapacks will get updated and it’s kinda untrue to call it unsaved

this technically will always be the case with any "temporary" variable but if you don't call it a "temp" variable because it aligns with the main config, you lose some readability

JacquiFreeElectron: ok then I'll do unsavedDatapackConfig

app/src/process-datapack-config.ts
paologum: yeah sorry, you should set it in this function, because whenever we call this, they should be the same anyways

i didn't realize they were diff strings when i made the comment earlier but either way they should be set in this function when it finishes

JacquiFreeElectron: gotcha

JacquiFreeElectron:
https://github.com/user-attachments/assets/277583ae-8058-41b3-8be7-c6e9c6be83df

app/src/App.tsx
```diff
@@ -73,6 +79,24 @@ export default observer(function App() {
...
+            onNo={() => {
+              navigate("/settings");
+              actions.setSettingsTabsSelected("datapacks");
+            }}
+            onClose={() => {
+              navigate("/settings");
+              actions.setSettingsTabsSelected("datapacks");
+            }}
+            customNo="Cancel"
+            customYes="Leave Without Changes "></TSCPopupDialog>
```
paologum, line 99: you can just make this a single tag that closes itself (no need for `</TSCPopupDialog>` if there is an empty body

app/src/Home.tsx
```diff
@@ -49,8 +50,11 @@ export const Home = observer(function Home() {
           style={{
             fontSize: theme.typography.pxToRem(12)
           }}
-          onClick={() => {
+          onClick={async () => {
             actions.removeCache();
+
+            await processDatapackConfig([]);
+            actions.setIsProcessingDatapacks(false);
```
paologum, line 57: `setIsProcessingDatapacks` should be included in `processDatapackConfig`

app/src/Home.tsx
```diff
@@ -97,15 +101,12 @@ const TSCPresetHighlights = observer(function TSCPresetHighlights({
...
-                        preset.settings
+                        preset.settings,
+                        { navigate: navigate, path: "/home" }
                       );
-                      // wait to see if we can grab necessary data
-                      if (success) {
-                        actions.initiateChartGeneration(navigate, "/home");
-                      }
-                      //TODO add an error message saying the data is irregular and can't be loaded
+                      actions.setIsProcessingDatapacks(false);
```
paologum, line 109: same here

app/src/NavBar.tsx
```diff
@@ -34,6 +35,13 @@ export const NavBar = observer(function Navbar() {
   const { anchorProps, hoverProps } = useHover(settingsMenuState.state, settingsMenuToggle);
 
   const location = useLocation();
+  /* const checkOpen = () => {
+    // need to use state.datapackSelection.regenerateChart to check if the user is trying to generate a chart without confirming selections
+    return (
+      state.datapackSelection.regenerateChart &&
+      JSON.stringify(state.datapackSelection.selectedDatapacks) !== JSON.stringify(state.config.datapacks)
+    );
+  }; */
```
paologum, line 44: remove

app/src/process-datapack-config.ts
```diff
@@ -0,0 +1,92 @@
+import { NavigateFunction } from "react-router-dom";
+import { actions, state } from "./state";
```
paologum, line 2: if you are using actions and state, `processDatapackConfig` should be an action.
remember we should never import state or actions unless this is an action or we use a context

app/src/components/TSCPopupDialog.tsx
```diff
@@ -25,10 +38,10 @@ export const TSCPopupDialog: React.FC<TSCPopupDialogProps> = ({ open, title, mes
       </DialogContent>
       <DialogActions className="dialog-actions">
         <Button variant="outlined" className="tsc-button" onClick={onNo}>
-          No
+          {no}
         </Button>
         <TSCButton buttonType="gradient" className="tsc-button" onClick={onYes}>
-          Yes
+          {yes}
```
paologum, line 44: for easier readability i would have here instead of up above
`customYes ?? "Yes"`

app/src/NavBar.tsx
```diff
@@ -118,9 +126,22 @@ export const NavBar = observer(function Navbar() {
...
+            if (JSON.stringify(state.unsavedDatapackConfig) !== JSON.stringify(state.config.datapacks)) {
+              await processDatapackConfig(JSON.parse(JSON.stringify(state)).unsavedDatapackConfig, "", {
+                navigate: navigate,
+                path: location.pathname
+              });
+              actions.setIsProcessingDatapacks(false);
+            } else {
+              actions.initiateChartGeneration(navigate, location.pathname);
+            }
+          }}>
```
paologum, line 141: i think this should just be 
```
await processDatapackConfig(...)
actions.initiateChartGeneration
```

like before `processDatapackConfig` should be a contexted actions method
also `processDatapackConfig` should have the functionality that
``` 
if (JSON.stringify(state.unsavedDatapackConfig) == JSON.stringify(state.config.datapacks)) return
```

app/src/process-datapack-config.ts
```diff
@@ -0,0 +1,92 @@
...
+          console.error(e);
+          actions.pushError(ErrorCodes.INVALID_SETTINGS_RESPONSE);
+          actions.setIsProcessingDatapacks(false);
+          reject(new Error("Failed to fetch and process settings."));
+        }
+      }
+      return null;
+    };
+
+    fetchSettings().then((chartSettings) => {
```
paologum, line 39: you should just `const chartSettings = await fetchSettings()`
that way we are not unecessarily indenting a large function

app/src/process-datapack-config.ts
```diff
@@ -0,0 +1,92 @@
...
+  navigate: NavigateFunction;
+  path: string;
+};
+
+export function processDatapackConfig(
+  datapacks: string[],
+  settingsPath?: string,
+  needChartGeneration?: InitialChartGeneration
+): Promise<string> {
+  return new Promise((resolve, reject) => {
```
paologum, line 15: the promise only needs to wrap the worker.
the rest can throw errors or however
if you throw errors here, just make sure everywhere this gets used catches and pushes errors properly

app/src/process-datapack-config.ts
```diff
@@ -0,0 +1,92 @@
...
+        } else {
+          actions.pushSnackbar("Setting Datapack Config Timed Out", "info");
+          reject(new Error("Setting Datapack Config Timed Out."));
+        }
+        setDatapackConfigWorker.terminate();
+
+        if (needChartGeneration) {
+          actions.initiateChartGeneration(needChartGeneration.navigate, needChartGeneration.path);
+        }
+      };
```
paologum, line 83: if you do the other comment's fix then you won't need this

app/src/state/actions/general-actions.ts
```diff
@@ -572,7 +441,10 @@ export const setDatapackConfig = action(
     // this is for app start up or when all datapacks are removed
     if (datapacks.length === 0) {
       state.settings.timeSettings["Ma"] = JSON.parse(JSON.stringify(defaultTimeSettings));
+    } else {
+      delete state.settings.timeSettings["Ma"];
     }
+    console.log(datapacks.length);
```
paologum, line 447: log

app/src/state/actions/general-actions.ts
```diff
@@ -572,7 +441,10 @@ export const setDatapackConfig = action(
     // this is for app start up or when all datapacks are removed
     if (datapacks.length === 0) {
       state.settings.timeSettings["Ma"] = JSON.parse(JSON.stringify(defaultTimeSettings));
+    } else {
+      delete state.settings.timeSettings["Ma"];
```
paologum, line 445: this will delete `Ma` for every datapack

server/assets/file_metadata.json
paologum, line 1: don't commit this

paologum (changes requested):

app/src/state/actions/general-actions.ts
JacquiFreeElectron: I think this is fine because the later code add back `Ma` if that datapack has the unit Ma: 
```typescript
if (chartSettings !== null) {
      assertChartInfoTSC(chartSettings);
      await applySettings(chartSettings);
    } else {
      // set any new units in the time
      for (const chart of columnRoot.children) {
        if (!state.settings.timeSettings[chart.units]) {
          state.settings.timeSettings[chart.units] = JSON.parse(JSON.stringify(defaultTimeSettings));
        }
      }
    }
```
I tried with TSC internal datapack & human evolution & human culture and it worked fine  (First I selected them all, the unit changed to `MA` and `KA`; then I unselected TSC internal datapack & human evolution, the unit changed to just `KA`. Then I unselected human culture as well, the unit changed back to just `Ma`; At last, I selected TSC internal datapack again, the unit remained `MA`)

paologum: ok then it should be fine, thanks for letting me know. i'll double check with `/deploy` when you finish as well

app/src/process-datapack-config.ts
JacquiFreeElectron: can I pushError in here or should I pushError where it's used

JacquiFreeElectron: I now only wrap the worker in the promise, use a try-catch block to wrap that promise, and pushError when catching any error thrown by the promise, it this ok?

paologum: yeah that works

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,135 @@
...
+  SetDatapackConfigReturnValue,
+  assertState
+} from "../../types";
+import { cloneDeep } from "lodash";
+
+/**
+ * sets chart to newval and requests info on the datapacks from the server
+ */
+self.onmessage = async (e: MessageEvent<SetDatapackConfigMessage>) => {
+  const { datapacks, chartSettings, stateCopy } = e.data;
```
paologum, line 15: you don't actually use the `chartSettings` in this function, so you should probably not send it

app/src/state/initialize.ts
```diff
@@ -7,7 +7,8 @@ export async function initialize() {
   actions.fetchDatapackIndex();
   actions.fetchMapPackIndex();
   actions.fetchPresets();
-  actions.setDatapackConfig([], "");
+  await actions.processDatapackConfig([]);
+  actions.setIsProcessingDatapacks(false);
```
paologum, line 11: should remove `actions.setIsProcessingDatapacks(false);`

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,135 @@
...
+  let stateCopyObj: State;
+  try {
+    stateCopyObj = JSON.parse(stateCopy);
+    assertState(stateCopyObj);
+  } catch (e) {
+    console.log(e);
+    console.error("failed to parse state in worker");
+    return;
+  }
```
paologum, line 24: this needs to throw an error or set a status or value or else it will never terminate right?

app/src/state/actions/general-actions.ts
```diff
@@ -416,149 +417,107 @@ const applyChartSettings = action("applyChartSettings", (settings: ChartSettings
...
-          for (const child of column.children) {
-            child.parent = column.name;
+      return null;
+    };
+    const chartSettings = await fetchSettings();
+    try {
+      await new Promise((resolve, reject) => {
+        const setDatapackConfigWorker: Worker = new Worker(
+          new URL("../../util/workers/set-datapack-config.ts", import.meta.url),
+          {
```
paologum, line 460: you should set a timer for this webworker because otherwise this could stall forever if someone miswrites something

paologum (changes requested):

app/src/state/actions/general-actions.ts
```diff
@@ -646,7 +600,6 @@ export const removeCache = action("removeCache", async () => {
 export const resetState = action("resetState", () => {
   setChartMade(true);
   setChartLoading(true);
-  setDatapackConfig([], "");
   setChartHash("");
   setChartContent("");
```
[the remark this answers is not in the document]
JacquiFreeElectron: that part is done by calling `actions.processDatapackConfig([])` when click remove cache button. bc I saw the comments said that `Resets state Only implementation is used when we remove cache` I removed the `setDatapackConfig` in here(since `setDatapackConfig()` is already called once inside `processDatapackConfig()`

app/src/state/initialize.ts
JacquiFreeElectron: my bad, forgot again

app/src/state/actions/general-actions.ts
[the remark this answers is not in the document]
paologum: yeah but if we ever wanted to call it somewhere else, you should have `processDatapackConfig` in here is what i meant instead of seperate
encapsulating as much as we can into one method is what we want for one functionality

app/src/util/workers/set-datapack-config.ts
JacquiFreeElectron: after I change to use `toJS` probably I don't need this part anymore. I'll check

JacquiFreeElectron: yeah by using` toJS` everything works fine and I can directly declare the type of `stateCopy `of `SetDatapackConfigMessage` to be `State`. So I assume we don't need assertion here anymore?

app/src/state/actions/general-actions.ts
JacquiFreeElectron: I think I have a timer inside the worker (in the file `set-datapack-config.ts`, a 2min timeout promise)  Isn't that already a timer?

app/src/Home.tsx
```diff
@@ -49,8 +49,9 @@ export const Home = observer(function Home() {
           style={{
             fontSize: theme.typography.pxToRem(12)
           }}
-          onClick={() => {
+          onClick={async () => {
             actions.removeCache();
+            await actions.processDatapackConfig([]);
             actions.resetState();
```
paologum, line 55: this should just be in `resetState`

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+    const returnValue: SetDatapackConfigReturnValue = {
+      columnRoot: columnRoot,
+      foundDefaultAge: foundDefaultAge,
+      mapHierarchy: mapHierarchy,
+      datapacks: datapacks,
+      mapInfo: mapInfo
+    };
+    return returnValue;
+  };
```
paologum, line 95: for easier readability can you take this out of the `onmessage` event and put it at the bottom of the file

app/src/state/actions/general-actions.ts
paologum: i forgot that your fetch settings are now outside so the timer should be ok now

app/src/state/actions/general-actions.ts
```diff
@@ -408,120 +409,105 @@ const applyChartSettings = action("applyChartSettings", (settings: ChartSettings
...
+
+export const processDatapackConfig = action(
+  "processDatapackConfig",
+  async (datapacks: string[], settingsPath?: string) => {
+    if (state.isProcessingDatapacks) return;
+    //when first open the website they are both empty. We still need to process the config under that situation
+    if (
+      JSON.stringify(state.unsavedDatapackConfig) == JSON.stringify(state.config.datapacks) &&
+      state.config.datapacks.length !== 0
+    )
```
paologum, line 428: there's a use case where if we call `processDatapackConfig([])` we don't need a loading screen because this will take no time to set variables to `[]`

so if the `datapacks.length === 0` then set the fields to what they should be on an empty config (make a different function and call it, then return)

also you should align `datapacks === state.config.datapacks` not `unsavedDatapackConfig === state.config.datapacks` here (as long as you're epassing in `unsavedDatapackConfig`

paologum (changes requested):
some minor things

app/src/util/workers/set-datapack-config.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+  await runWithTimeout();
+  self.postMessage(message);
+};
+const setDatapackConfig = (datapacks: string[], stateCopy: State) => {
+  const unitMap: Map<string, ColumnInfo> = new Map();
+  let mapInfo: MapInfo = {};
+  let mapHierarchy: MapHierarchy = {};
+  let foundDefaultAge = false;
+  const columnRoot: ColumnInfo = defaultColumnRoot;
```
paologum, line 40: `cloneDeep` otherwise the obj changes

app/src/state/actions/general-actions.ts
```diff
@@ -408,120 +411,131 @@ const applyChartSettings = action("applyChartSettings", (settings: ChartSettings
...
+  state.isProcessingDatapacks = isProcessingDatapacks;
+});
+
+export const setUnsavedDatapackConfig = action("setUnsavedDatapackConfig", (newDatapacks: string[]) => {
+  state.unsavedDatapackConfig = newDatapacks;
+});
+
+const setEmptyDatapackConfig = action("setEmptyDatapackConfig", () => {
+  resetSettings();
+  const columnRoot: ColumnInfo = defaultColumnRoot;
```
paologum, line 424: same here

app/src/state/actions/general-actions.ts
```diff
@@ -408,120 +411,131 @@ const applyChartSettings = action("applyChartSettings", (settings: ChartSettings
...
+
+export const processDatapackConfig = action(
+  "processDatapackConfig",
+  async (datapacks: string[], settingsPath?: string) => {
+    if (state.isProcessingDatapacks) return;
+    if (JSON.stringify(datapacks) == JSON.stringify(state.config.datapacks)) return;
+    if (datapacks.length === 0) {
+      setEmptyDatapackConfig();
+      return;
+    }
```
paologum, line 456: we want to have `setEmptyDatapackConfig` before the equality check, just for the case that even if `datapacks === state.config.datapacks` doesn't have a correct configuration of everything that `setEmptyDatapackConfig` sets we confirm reset it regardless. hopefully that makes sense
also so we can have less lines `451` and `452` should prolly be merged

app/src/types.ts
```diff
@@ -192,3 +210,13 @@ export function assertSettingsTabs(value: string): asserts value is SettingsTabs
...
+
+/* const unitMap: Map<string, ColumnInfo> = new Map();
+let mapInfo: MapInfo = {};
+let mapHierarchy: MapHierarchy = {};
+let foundDefaultAge = false;*/
+
+// all chart root font options have inheritable on
+/*for (const opt in columnRoot.fontsInfo) {
+  columnRoot.fontsInfo[opt as keyof FontsInfo].inheritable = true;
+} */
```
paologum, line 222: comment

paologum (changes requested):
just some really minor things but the overall code looks really good :D

app/src/util/workers/set-datapack-config.ts
paologum: nope we should be good

paologum (approved):
nice, looks great!

PR #213, opened 2024-08-05
fixed loaded datamining columns not generating properly

jay-jhk-lee:
Now properly sets the reference column's respective `drawExtraColumn` variable when loading datamining columns

### DIFF
--- a/app/src/state/actions/column-actions.ts
+++ b/app/src/state/actions/column-actions.ts
@@ -169,7 +169,25 @@ export function handleDataMiningColumns() {
       );
       continue;
     }
+    
     setColumnProperties(createdDmColumn, foundDmColumn);
+    
+    switch (refCol.columnDisplayType) {
+      case "Event":
+        assertEventSettings(refCol.columnSpecificSettings);
+        if (isEventFrequency(loadedDataMiningType)) refCol.columnSpecificSettings.frequency = loadedDataMiningType;
+        break;
+      case "Chron":
+        assertChronSettings(refCol.columnSpecificSettings);
+        if (isDataMiningChronDataType(loadedDataMiningType))
+          refCol.columnSpecificSettings.dataMiningChronDataType = loadedDataMiningType;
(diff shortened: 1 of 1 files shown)


app/src/state/actions/column-actions.ts
```diff
@@ -169,7 +169,25 @@ export function handleDataMiningColumns() {
...
+      case "Chron":
+        assertChronSettings(refCol.columnSpecificSettings);
+        if (isDataMiningChronDataType(loadedDataMiningType))
+          refCol.columnSpecificSettings.dataMiningChronDataType = loadedDataMiningType;
+        break;
+      case "Point":
+        assertPointSettings(refCol.columnSpecificSettings);
+        if (isDataMiningPointDataType(loadedDataMiningType))
+          refCol.columnSpecificSettings.dataMiningPointDataType = loadedDataMiningType;
+    }
```
paologum, line 189: just to be safe so the app doesn't crash, wrap in `try` and `console.error` and `continue` on error

paologum: also if not any of these three cases, should also handle that with contiue/logging just to be robust

paologum (changes requested):

app/src/state/actions/column-actions.ts
```diff
@@ -171,25 +171,31 @@ export function handleDataMiningColumns() {
...
+        case "Point":
+          assertPointSettings(refCol.columnSpecificSettings);
+          if (isDataMiningPointDataType(loadedDataMiningType))
+            refCol.columnSpecificSettings.dataMiningPointDataType = loadedDataMiningType;
+          break;
+        default:
+          console.log("WARNING: datamining reference column's type is not event, chron, or point");
+      }
+    } catch (e) {
+      console.log(e);
```
paologum, line 197: i would `console.error` so we can stack trace this

paologum (approved):

PR #218, opened 2024-08-07
Tsc 436 load settings doesn t work properly for a certain settings file

Changk820:
From what I can tell, I was able to fix one of the loading issues for the settings, the issue was that leaf settings weren't being processed correctly, I'm honestly not 100% sure what the original code was doing, just that it seemed to be trying to change settings of a nonexistent child settings for leaf settings. By leaf settings, I mean:
image

I don't really understand what datamining even is, so I'm hoping I didn't break anything regards to that
  addColumnToDataMiningCache(settings);

Now for the other issue Gabi had, which was 
image

The unitsperMY not changing, I think the issue is that it isn't parsing the unitsPerMY properly prior to being passed into the function to apply the settings, because my console.log showed that setting as being empty.

I have some questions though, why in the code is there a for loop to do this? It looks like unitsPerMY was supposed to be an array? But isn't it just a single value we change in the image above?

  for (const unit of unitsPerMY) {
    if (!state.settings.timeSettings[unit.unit]) {
      state.settings.timeSettings[unit.unit] = JSON.parse(JSON.stringify(defaultTimeSettings));
    }
    setUnitsPerMY(unit.text / 30, unit.unit);
  }

The code snippet is in /app/src/state/actions/general-actions.ts

Also, what's up with the division by 30? When saving, we also multiplied by 30, what's up with that? Why even have that 30 if we are just going to remove it.

The above is just some weird code I saw, but the actual issue might be in the parser, which I'm looking at still.

### DIFF
--- a/app/src/state/actions/column-actions.ts
+++ b/app/src/state/actions/column-actions.ts
@@ -59,7 +59,7 @@ function extractColumnType(text: string): string {
 function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
   console.log("Settings(setColProp):", settings);
   console.log("Column(setColProp):", column);
-  
+
   setEditName(settings.title, column);
   setEnableTitle(settings.drawTitle, column);
   if ("showUncertaintyLabels" in column) setShowUncertaintyLabels(settings.drawUncertaintyLabel, column);
@@ -257,15 +257,15 @@ export function addColumnToDataMiningCache(settings: ColumnInfoTSC) {
 
 export const applyChartColumnSettings = action("applyChartColumnSettings", (settings: ColumnInfoTSC) => {
   const columnName = extractName(settings._id);
-  let curcol: ColumnInfo | undefined =
+  const curcol: ColumnInfo | undefined =
     state.settingsTabs.columnHashMap.get(columnName) ||
     state.settingsTabs.columnHashMap.get("Chart Title in " + columnName);
   console.log(`Processing column: ${columnName}`);
-  console.log('Current column:', curcol);
-  console.log('Settings:', settings);
+  console.log("Current column:", curcol);
+  console.log("Settings:", settings);
   // Apply settings to the current column
   if (curcol) {
-    console.log('Applying settings to column:', columnName);
+    console.log("Applying settings to column:", columnName);
     setColumnProperties(curcol, settings);
   } else {
     console.warn(`Column not found in hash map: ${columnName}`);
--- a/app/src/state/actions/column-actions.ts
+++ b/app/src/state/actions/column-actions.ts
@@ -57,6 +57,9 @@ function extractColumnType(text: string): string {
   return text.substring(text.indexOf(".") + 1, text.indexOf(":"));
 }
 function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
+  console.log("Settings(setColProp):", settings);
+  console.log("Column(setColProp):", column);
+  
   setEditName(settings.title, column);
   setEnableTitle(settings.drawTitle, column);
   if ("showUncertaintyLabels" in column) setShowUncertaintyLabels(settings.drawUncertaintyLabel, column);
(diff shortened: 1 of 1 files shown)


app/src/state/actions/column-actions.ts
```diff
@@ -254,28 +257,31 @@ export function addColumnToDataMiningCache(settings: ColumnInfoTSC) {
...
+  console.log(`Processing column: ${columnName}`);
+  console.log("Current column:", curcol);
+  console.log("Settings:", settings);
+  // Apply settings to the current column
   if (curcol) {
+    console.log("Applying settings to column:", columnName);
     setColumnProperties(curcol, settings);
+  } else {
+    console.warn(`Column not found in hash map: ${columnName}`);
   }
```
paologum, line 272: logs

app/src/state/actions/column-actions.ts
```diff
@@ -254,28 +257,31 @@ export function addColumnToDataMiningCache(settings: ColumnInfoTSC) {
...
+      const childSettings = settings.children[i];
+      console.log(`Recursively processing child: ${extractName(childSettings._id)}`);
+      applyChartColumnSettings(childSettings);
     }
   } else {
-    for (let i = 0; i < settings.children.length; i++) {
-      applyChartColumnSettings(settings.children[i]);
-    }
+    console.warn(`No children found for column: ${columnName}`);
   }
```
paologum, line 285: you need to keep this code because columns of `BlockSeriesMetaColumn` are always
`Facies`, `Chron`, `Series Label`, etc which are NOT unique names so they have to be changed into a unique name where their parent and their "label" are in the name
this way you ignore that and thus don't apply the settings to the block series meta column

app/src/state/actions/column-actions.ts
```diff
@@ -57,6 +57,9 @@ function extractColumnType(text: string): string {
   return text.substring(text.indexOf(".") + 1, text.indexOf(":"));
 }
 function setColumnProperties(column: ColumnInfo, settings: ColumnInfoTSC) {
+  console.log("Settings(setColProp):", settings);
+  console.log("Column(setColProp):", column);
+
   setEditName(settings.title, column);
```
paologum, line 63: log

paologum (changes requested):
i'm confused on what the fix is

paologum:
you have to multiply/divide by 30 i don't remember but you can take a look at the java file for reference.
everything we do in terms of backend logic is supported by the java file so if you're unsure you need to check the java file out
additionally, there can be multiple units so multiple `unitsPerMy`

app/src/state/actions/column-actions.ts
```diff
@@ -267,7 +267,7 @@ export const applyChartColumnSettings = action("applyChartColumnSettings", (sett
     for (let i = 0; i < settings.children.length; i++) {
       const child = settings.children[i];
       const childName = extractName(child._id);
-      curcol = state.settingsTabs.columnHashMap.get(columnName + " " + childName);
+      curcol = state.settingsTabs.columnHashMap.get(childName);
```
paologum, line 270: i'm confused how these two lines fix the error
the problem is we aren't writing the settings xml correctly when we write the state to the xml
if you revert this change, it should never find the column because the name in state should be `parent child`
while the xml version is just `child` when the column is `BlockSeriesMetaColumn`

app/src/state/parse-settings.ts
```diff
@@ -88,7 +88,7 @@ function processSettings(settingsNode: Element): ChartSettingsInfoTSC {
     if (!settingName) continue;
 
     const nestedSettingsNode = settingNode.getElementsByTagName("setting")[0];
-    if (!nestedSettingsNode) continue;
+    if (!nestedSettingsNode && settingName !== "unitsPerMY" && settingName !== "skipEmptyColumns") continue;
```
paologum, line 91: why do you continue these?
there's code below here on line 129 that sets these. if there is a fix at least it should have one or the other, but i don't see why we should skip them

paologum (changes requested):

app/src/state/parse-settings.ts
Changk820: If those aren't excluded, then those settings don't get set for some reason

app/src/state/actions/column-actions.ts
Changk820: When I printed, I just saw that it was showing repeats in the name, for some reason childname already had columnName, so I just removed it and it worked.

paologum: ok well what i think is happening is that you need too apply both the parent and the children settings. before this pr it only did the children, and now you changed it to only applying to the parent.
additionally it's still writing the settings file incorrectly so you will need to change that

paologum (changes requested):

app/src/state/parse-settings.ts
paologum: ok well i believe what's happening is `nestedSettingsNode` is checking for any nested children and if there aren't any nested children it skips them, so you should just fix the logic so that it does something if it is nested vs not nested

Changk820: I think I'm just going to remove it completely, since this nested stuff is only used 
if (settingName === "topAge" || settingName === "baseAge") 
which already has a check via 
nestedSettingsNode.getAttribute("name") === "stage"
since get returns null if nothing exists and is treated as falsy for boolean operations

Changk820:
Okay, the fix you suggested worked
We might need to make a note to Gabi that the TSC setting file she included originally won't work, since we wrote it out wrong originally like <column id="class datastore.ZoneColumn:Right column (region N2) Facies Label" >
instead of just Facies label

paologum:
it's because it's attempting to import `Mobx observable state` into `vitest` which isn't allowed.
you should probably find a workaround so that you don't actually import the state but you change the object

PR #227, opened 2024-08-10
Tsc 485 event search minor changes

jay-jhk-lee:
Changed label of "status" column to "column on?"
changed label of "qualifier" to "type"
changed time display to a table for line up

image

### DIFF
--- a/app/src/settings_tabs/Search.css
+++ b/app/src/settings_tabs/Search.css
@@ -17,6 +17,13 @@
   min-height: 70px;
   resize: both;
 }
+.search-time-display-first-header {
+  padding: 0;
+}
+.search-time-display-second-header {
+  top: 26px;
+  padding: 0 16px;
+}
 
 .search-header-container {
   display: flex;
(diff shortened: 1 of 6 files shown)


paologum:
it probably would make more sense to align text to the left
## `text-align: left`

Screenshot 2024-08-14 at 4 12 37 PM

## `text-align: right`
Screenshot 2024-08-14 at 4 12 51 PM

paologum:
probably atp we might just want to have two cells for `Top` and `Base` and potentially put it on the left or right but we can do that/reconvene on that later

jay-jhk-lee:
changed alignment of times
image

app/src/settings_tabs/Search.css
```diff
@@ -13,35 +13,25 @@
 }
 
 .search-time-display-container {
-  min-width: 10vw;
-  padding-left: 20px;
+  min-width: 200px;
+  min-height: 70px;
+  resize: both;
```
paologum, line 18: you should add a `max-height` and `max-width` so that they can't expand it to the whole page

paologum (changes requested):

paologum (approved):

PR #228, opened 2024-08-12
Update server.md to include decrypter info

Changk820:


server.md
```diff
@@ -98,6 +98,87 @@ After decrypting the datapacks, the server will simply listen for any requests f
 
 ---
 
+### Decrypter
+
+The TSC Online platform uses a specific JAR file, named `datapack-decrypter_DDMMMYYYY.jar`, to decrypt encrypted datapacks during server startup. For instance, a typical JAR file might be named `datapack-decrypter_21Jun2024.jar`.
+
+To ensure the server starts correctly, this JAR file must be placed in the `tsconline/server/assets/jars` directory alongside the active server JAR file. Additionally, the JAR file and its input arguments should be specified in the `tsconline/server/assets/config.json` file.
```
paologum, line 105: this is not true it's labeled in a .gitignored `admin-config.json` file

server.md
```diff
@@ -98,6 +98,87 @@ After decrypting the datapacks, the server will simply listen for any requests f
...
+Change it to:
+```
+private static String version_number = "1.01";
+```
+
+This ensures that the new decryption JAR will redecrypt the datapacks using the updated logic.
+
+
+#### Building the JAR
+To create the decryption JAR file, use Maven.
```
paologum, line 154: add a note or whatever here that this is within the bitbucket repo since they could think this is seperate from above or something. just to clarify

paologum (changes requested):

server.md
Changk820: I think other places might still say config.json instead of admin-config.json. I changed mine, but I'm not sure if other areas should also be changed since we have two sites.

Is this readme supposed to be for future VIP students or the deepearth people?

paologum: anyone really, it's just meant as a documentation for anyone who wants to run/dev on this code

paologum (approved):

PR #230, opened 2024-08-13
Tsc 349 search regex expressions for searching columns

paologum:
## Main

- added search for `*` wildcards
- distinct from the jar in the fact that we always look for substrings while the jar only looks for exact equality
- therefore `*` is only ever useful in the middle of two terms like `goc*tii` if you want to match the front and back of a title
- known weird occurence: there is a moment where sometimes i think the `groupedEvents` get pushed twice to `stretchedEvents` on an observer refresh but it only ever happens when you `add/remove` the same datapack and it cleans up on the second refresh. just noting it here (ie. have `Internal` and `Australia` with `goch*ii` as search term then `remove Internal` and immediately switch to `Search`

## demo

https://github.com/user-attachments/assets/163a855f-047b-401d-ac0e-fe80d2f5f81a


app/src/state/actions/column-actions.ts
```diff
@@ -779,6 +785,127 @@ export const addAgeColumn = action((column: ColumnInfo) => {
...
+  }
+  while (column.name !== "Chart Root") {
+    columnPath.push(column.editName);
+    column = state.settingsTabs.columnHashMap.get(column.parent!);
+    if (!column) break;
+  }
+  return columnPath;
+});
+export const searchEvents = action(async (searchTerm: string, counter = { count: 0 }) => {
+  setEventSearchTerm(searchTerm);
```
asivath, line 802: while typing this keeps getting called so wouldn't there be previous runs happening at the same time? there should be something to cancel previous runs, like a generated token or something

asivath (approved):

PR #232, opened 2024-08-13
Tsc 489 user datapack deletion front end

paologum:
## Main

- delete datapack frontend
- always present in datapack menu which is conditionally loaded if the datapack has a `uuid`
- `Compact` and `Row` views of datapack have delete icons
- Future: could add an are you sure dialog

## Other

- took off a `snackbar` push when `userDatapacks` are fetched due to multiple uploads triggering duplicate keys when it's just fetching all the datapacks again.
- could potentially need a way to fetch that could verify staleness/integrity of the index currently on the app

## Demo

https://github.com/user-attachments/assets/9d170b02-e725-42b7-83f3-f7338f04a7a1

### dark mode

image


JacquiFreeElectron (approved):
looks good

PR #237, opened 2024-08-18
added error codes for chart generation failure

asivath:


app/src/state/actions/generate-chart-actions.ts
```diff
@@ -151,22 +151,32 @@ export const fetchChartFromServer = action("fetchChartFromServer", async (naviga
...
-          );
+          errorCode = ErrorCodes.SERVER_FILE_METADATA_ERROR;
           break;
         case 400:
-          displayServerError(
-            answer,
-            ErrorCodes.INVALID_CHART_FROM_JAR,
-            ErrorMessages[ErrorCodes.INVALID_CHART_FROM_JAR]
-          );
+          errorCode = ErrorCodes.INVALID_CHART_FROM_JAR;
```
paologum, line 160: you probably should join 400 with `INTERNAL_ERROR` because that means the jar didn't execute/failed to run
if `INVALID_CHART_FROM_JAR` isn't used anywhere else you can also get rid of it

paologum (changes requested):

PR #239, opened 2024-08-22
Tsc 512 refactor mapindex to be a part of base datapack

paologum:
# Main

- this is purely a refactor pr, unfortunately changing things to be more understanding requires changing the implementation everywhere so this `git diff` is pretty big
- mappacks exist within datapacks and not in their own index
- datapacks are now indexed by `title` in the `uploads` directory on the server and on the app side `DatapackIndex` is indexed by `title`
- routes for viewing datapack metadata on the `app` side are now `/datapack/:title` (still `:id` in the code but this is how it operates)
- refactor popups to point to the correct directory (added more fields)
- refactor chart request to require more than just datapack names. theoretically any amount of datapacks can be used with the same names as long as the proper directories exist. (ie public user, private user, workshop, and server datapacks can have the same name and it'll still generate the chart/point to the correct areas properly)
- fixed `parse-map-packs.ts` to not use `any` and have correct `typing`
- fix a map points issue where correct map point maps were not being displayed (the error was that only parents and their children were considered/displayed which disregarded the lone map points ie Africa Bight)
- upload directory now looks like below (uuids store datapacks individually and fetch requests parse each `Datapack.json`
- also `public` below refers to `uploads/public` not `server/public`
# Directory Structure

- **uploads/**
  - `{uuid}/`
    - `{datapack.title}`...
    - `public/`
  
- **public/**
  - `datapacks/`
  - `DatapackIndex.json`
  
- **{datapack.title}**
    - `datapack.file`
    - `decrypted/`
    - `encrypted-datapacks/`
    - `Datapack.json`


JacquiFreeElectron:
I got this two errors when clicked `generate `on the home page datapack card
Screenshot 2024-08-22 at 3 05 18 PM
Screenshot 2024-08-22 at 3 05 28 PM

asivath (approved):

JacquiFreeElectron (approved):
looks good

PR #246, opened 2024-09-01
Added additional info to database profile

lbostre:
08ef30e6458e1d121e5c346b748ff8a8
85fd888a48dbb2c4850489e474531e82

a7349068e397aa3e2344850fc66986ac
54ea4b5d9409c80901e49533d888aecd

1ef06928a4c844487a797ba4e6097a09

Added additional data pack information including contact, notes, and column type count.


paologum:
for the piechart can you change the title to be `Number of Columns` and also make it so the slices have little slivers/borders between them. also can you add either a table or some sort of display so that someone can see the exact numbers (not percents but you could do both thought idk if that would look good). you could additionally add it as a subtitle of the slice label
 (like below) example is mainly for the gap and the extra label not really emphasizing the pie vs donut both look good

Screenshot 2024-09-02 at 8 25 33 PM

paologum:
on my local, the only display is one point as well

Screenshot 2024-09-02 at 9 03 48 PM

app/src/DatapackProfile.tsx
```diff
@@ -194,3 +199,143 @@ export const DatapackWarningAlert: React.FC<DatapackWarningProps> = ({ warning }
...
+          .attr("d", function (d) {
+            return arcOver(d as d3.PieArcDatum<{ label: string; value: number }>);
+          });
+
+        tooltip.transition().duration(200).style("opacity", 0.9);
+        tooltip
+          .html(`${d.data.label}: ${d.data.value}`)
+          .style("left", event.pageX + 10 + "px")
+          .style("top", event.pageY - 28 + "px");
+      })
```
paologum, line 289: is this a hard coded coordinate to where the graph sits on the whole page? if that's the case it would be a little annoying to change if we decided to move the graph somewhere else or use this chart somewhere else with different data. unless this is representative of the svg's width and height then it should be fine

app/src/DatapackProfile.tsx
```diff
@@ -194,3 +199,143 @@ export const DatapackWarningAlert: React.FC<DatapackWarningProps> = ({ warning }
...
+      .style("fill", "#fff")
+      .transition()
+      .ease(d3.easeLinear)
+      .delay(1000)
+      .duration(500)
+      .style("opacity", 1);
+
+    // Clean up the tooltip on unmount
+    return () => {
+      tooltip.remove();
```
paologum, line 336: you probably want to clean up (remove all of the svg) on unmount too

app/src/DatapackProfile.tsx
paologum: you should also probably add some pictures at different window/screen sizes

app/src/DatapackProfile.tsx
```diff
@@ -194,3 +199,143 @@ export const DatapackWarningAlert: React.FC<DatapackWarningProps> = ({ warning }
...
+    const chartData = Object.entries(data)
+      .filter(([_, value]) => value > 0)
+      .map(([label, value]) => ({ label, value }));
+
+    const g = svg.append("g").attr("transform", `translate(${width / 2},${height / 2})`);
+
+    const tooltip = d3
+      .select("body")
+      .append("div")
+      .attr("class", "tooltip")
```
paologum, line 259: this is a class that is pretty generic and could get accidentally affected by other class names in the repo so you might want to make this more specific. unless i'm understanding incorrectly and this is an inherent class in d3

paologum (changes requested):
some comments

app/src/DatapackProfile.tsx
lbostre: those values are hard-coding the offset of the tooltip from the cursor whenever the user hovers over the chart

paologum: ok as long as the tooltip works on different screen sizes this is fine

lbostre:
> on my local, the only display is one point as well (checked for each datapack and it's the same) it does work on `pr-preview.geolex.org` but on local it's weird for me i'm not sure if you have input on the below
> Screenshot 2024-09-02 at 9 03 48 PM
> Screen.Recording.2024-09-02.at.9.07.42.PM.mov

Ya I'm not sure why this could be happening. What I have found on my local is that every chart has these same values, which is also odd.
cc780be793eb7f88f575118e5e0cddb3

paologum (approved):
looks good!

PR #247, opened 2024-09-02
Tsc 79 multi language support

JacquiFreeElectron:
### Multi Language Support Part 1 ###
**What's Finished?**

almost everything that's visible to users before they logging in, including:
- Nav bar
- Login page
- Settings of server datapacks
- Dialogs & Errors

are all now available in Chinese.
_Notice: Snackbars are not translated at this point._

**Usage**
click/hover over the little world icon on the nav bar to switch languages. By default, when opening it, the website uses the default language of your browser, or English if your default browser language is not available.
Screenshot 2024-09-01 at 8 45 19 PM


paologum:
# running locally

- i would run through preview but for some reason, requests to `pr-preview` on my wifi are soo bad so i pulled locally and built
- the image below shows up when i don't load, the option for en is removed when i pick a language and it works fine after

Screenshot 2024-09-02 at 10 18 20 PM

- some missing translations (i would only expect these to be translated for chinese so you don't have to worry about others)

Screenshot 2024-09-02 at 10 21 41 PM

Screenshot 2024-09-02 at 10 22 09 PM

Screenshot 2024-09-02 at 10 21 31 PM

Screenshot 2024-09-02 at 10 20 37 PM

paologum:
also i do get some errors in the `app` console if you can fix those, i'm not sure where they come from though

app/src/settings_tabs/advanced_settings/EventSpecificSettings.tsx
```diff
@@ -10,52 +10,59 @@ import { Box } from "@mui/material";
...
+  alphabeticalText: string;
 };
-export const EventSpecificSettings: React.FC<EventSpecificSettingsProps> = observer(({ column }) => {
-  const { actions } = useContext(context);
-  if (column.columnDisplayType !== "Event" || !column.columnSpecificSettings) return null;
-  assertEventSettings(column.columnSpecificSettings);
-  const handleTypeChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+export const EventSpecificSettings: React.FC<EventSpecificSettingsProps> = observer(
+  ({ column, eventsText, rangesText, firstOccurrenceText, lastOccurrenceText, alphabeticalText }) => {
+    const { actions } = useContext(context);
```
paologum, line 21: are you not able to use the `hook` here?

if you probably be better to have less parameters than we need for the settings so if you can, i would make another `useTranslate` hook here

app/src/settings_tabs/advanced_settings/RangeSpecificSettings.tsx
```diff
@@ -6,30 +6,35 @@ import { TSCRadioGroup } from "../../components/TSCRadioGroup";
...
+  alphabeticalText: string;
 };
-export const RangeSpecificSettings: React.FC<RangeSpecificSettingsProps> = observer(({ column }) => {
-  const { actions } = useContext(context);
-  if (column.columnDisplayType !== "Range" || !column.columnSpecificSettings) return null;
-  assertRangeSettings(column.columnSpecificSettings);
-  const handleRangeSortChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+export const RangeSpecificSettings: React.FC<RangeSpecificSettingsProps> = observer(
+  ({ column, firstOccurrenceText, lastOccurrenceText, alphabeticalText }) => {
+    const { actions } = useContext(context);
```
paologum, line 15: same here and all the other ones

paologum (changes requested):
two things
- what happens when the key doesn't exist, does it display nothing? it should probably have a default text so we know for sure (if you can set it)
- secondly is there a way that `react-i18next` can tell us that every `t(...)` maps to a key that is present in a translated text? or something to make that easier i guess so we don't have to manually debug whether every language has every translation correctly? if not we will just have to manually keep track everytime we make new features and have a default text from the first point above to remind ourselves

app/src/settings_tabs/advanced_settings/EventSpecificSettings.tsx
JacquiFreeElectron: I can't use useTranslate here. If I use a hook here the console will say I break the rule of the usage of hook and after a few clicks the whole page goes blank

JacquiFreeElectron:
> * is there a way that `react-i18next` can tell us that every `t(...)` maps to a key that is present in a translated text? or something to make that easier i guess so we don't have to manually debug whether every language has every translation correctly

Yeah I think there's a build-in functionality called `missing keys`, not sure if that's 100% what we want but after a quick look I think that's at least a solid solution

> also i do get some errors in the `app` console if you can fix those, i'm not sure where they come from though

Do you mind telling me what exactly are those errors? bc I don't see any in my `app` console

app/src/settings_tabs/advanced_settings/EventSpecificSettings.tsx
paologum: that's weird, what is the error because we use the hook `useContext` here

JacquiFreeElectron: hmm yeah you're right. I'll add the hook back and double check the error / let you know tmr

JacquiFreeElectron: So it turns out a bug of data mining causes the blank page. Not sure how to fix that for now. But I've added the hooks back. The bug is that sometimes when clicking datamining settings menu without clicking any of the options and leaving to another column, the page whites out, giving the error that Error: Object 'DataMiningSettings' must have a 'window-size' number property, and some other times, when doing datamining for a datamining column (such as `LAD for ..` , even if I chose an option, the same issue still happened

JacquiFreeElectron:
The `missing key` call back function works perfectly fine, we can just pass a default text to the translator and if the key doesn't exist the default text will automatically show (ex. `t("a.key.that.doesnt.exist", "default text"`).  It's just it would be a huge amount of work for me to add a default text to every string I've translated so far. So I suggest that it be another separate pr.

JacquiFreeElectron:
The error in the console was caused by using the hover menu for language. Since I can't rename the `anchorProp` (and the other prop) to use the hook on different hover menus (which is the main reason for the errors we encountered), I switched to using MUI's Menu component. It doesn't look as polished as before, but since we're likely to implement a kebab menu and move its position anyway, I'll leave it as is for now
Screenshot 2024-09-12 at 11 11 49 PM
also, the MissingKeyHandler function is added. It's pretty simple, about 3 lines of code
also added some more translations

paologum:
the order of the languages should not change

also remove all cookies locally from local host and then try accessing the languages.
i don't have the picture but english appears twice (i'd assume for you it would appear as chinese twice)

https://github.com/user-attachments/assets/28767760-85be-4e18-b181-f0e5da6bad45

paologum (approved):

app/src/settings_tabs/column_menu/ColumnMenu.tsx
```diff
@@ -88,26 +90,30 @@ type ColumnContentProps = {
...
     case "General":
       return (
         <StyledScrollbar>
           <Box display="flex" flexDirection="column" gap="10px">
-            <EditNameField column={column} />
-            {column.children.length === 0 && <ChangeBackgroundColor column={column} />}
+            <EditNameField column={column} text={t("settings.column.menu.edit-name")} />
+            {column.children.length === 0 && (
+              <ChangeBackgroundColor column={column} text={t("settings.column.menu.background-color")} />
+            )}
```
paologum, line 109: this can be a hook?

app/src/settings_tabs/column_menu/ColumnMenu.tsx
```diff
@@ -119,7 +125,7 @@ const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) =
               />
             )}
             <div className="column-advanced-controls">
-              <AccordionPositionControls column={column} />
+              <AccordionPositionControls column={column} text={t("settings.column.menu.shift-row")} />
```
paologum, line 128: this too right

app/src/settings_tabs/column_menu/ColumnMenu.tsx
```diff
@@ -129,14 +135,14 @@ const ColumnContent: React.FC<ColumnContentProps> = observer(({ tab, column }) =
...
+                  <Typography>{t("settings.column.menu.add-blank-column")}</Typography>
                 </TSCButton>
                 <TSCButton className="add-blank-or-age-button" onClick={addAgeColumn}>
-                  <Typography>Add Age Column</Typography>
+                  <Typography>{t("settings.column.menu.add-age-column")}</Typography>
                 </TSCButton>
               </Box>
             )}
-            {!!column.popup && <InfoBox info={column.popup} />}
+            {!!column.popup && <InfoBox info={column.popup} titleText={t("settings.column.menu.info-box")} />}
```
paologum, line 145: this

paologum (approved):

PR #252, opened 2024-09-04
Tsc 522 add assets config class function implementation that makes it more modular easier to test

paologum:
# Main 
- made loading/removing a datapack modular for ease of use in testing and in making atomic (for when we do edits)
- fix after uploading a server datapack, it doesn't add the correct datapack to the index
- fix for spaces in title name saving to the correct directory/name in the filesystem when decrypting

### DIFF
--- a/server/src/admin/admin-config.ts
+++ b/server/src/admin/admin-config.ts
@@ -29,6 +29,7 @@ export async function loadAdminConfig(newFilepath?: string) {
       setAdminConfig(ADMIN_DEFAULT_CONFIG);
     } else {
       handleError(e);
+      throw e;
     }
   } finally {
     filepath = tempPath;
--- a/server/src/admin/admin-config.ts
+++ b/server/src/admin/admin-config.ts
@@ -1,0 +1,110 @@
+import { readFile, writeFile } from "fs/promises";
+import { assertAdminConfig } from "../types";
+import { AdminConfigType } from "../types";
+import { Mutex } from "async-mutex";
+import { DatapackMetadata } from "@tsconline/shared";
+
+const ADMIN_DEFAULT_CONFIG: AdminConfigType = {
+  datapacks: []
+};
+
+export class AdminConfig {
+  private mutex = new Mutex();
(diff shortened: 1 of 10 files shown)


server/src/admin/admin-config.ts
```diff
@@ -0,0 +1,111 @@
...
+    const raw = await readFile(tempPath, "utf8");
+    const adminconfig = JSON.parse(raw);
+    assertAdminConfig(adminconfig);
+    setAdminConfig(adminconfig);
+  } catch (e) {
+    if (e instanceof Error && (e as NodeJS.ErrnoException).code === "ENOENT") {
+      console.error("Admin config file not found. Loading default admin config.");
+      setAdminConfig(ADMIN_DEFAULT_CONFIG);
+    } else {
+      handleError(e);
```
asivath, line 31: shouldn't this throw an error as well

server/src/admin/admin-routes.ts
```diff
@@ -188,6 +189,8 @@ export const adminUploadServerDatapack = async function adminUploadServerDatapac
   let filepath: string | undefined;
   let decryptedFilepath: string | undefined;
   const fields: { [fieldname: string]: string } = {};
+  const datapacks = getAdminConfigDatapacks();
+  console.log(datapacks);
```
asivath, line 193: log

server/src/util.ts
```diff
@@ -76,7 +76,8 @@ export async function grabFilepaths(files: string[], topDirectory: string, botDi
       .map((name) => {
         const lastIndex = name.lastIndexOf(".");
         const filename = lastIndex !== -1 ? name.substring(0, lastIndex) : name;
-        return `${topDirectory}/${filename}/${botDirectory}/.*`;
+        const escapedFilename = filename.replace(/[-/\\^$*+?.()|[\]{}]/g, "\\$&");
+        return `${topDirectory}/${escapedFilename}/${botDirectory}/.*`;
```
asivath, line 80: will this work on windows?

asivath (approved):

PR #258, opened 2024-09-13
Tsc 502 workshop priority during workshop

asivath:
Besides priority I did a little refactoring and also added the workshopTItle on session check which I forgot to add in last PR


server/src/routes/login-routes.ts
```diff
@@ -205,13 +206,21 @@ export const sessionCheck = async function sessionCheck(request: FastifyRequest,
...
     }
-    const { email, username, pictureUrl, hashedPassword, isAdmin } = user;
+    const { email, username, pictureUrl, hashedPassword, isAdmin, workshopId } = user;
+    let workshopTitle = "";
+    if (workshopId) {
+      const workshop = await getAndHandleWorkshopEnd(workshopId);
+      if (workshop && new Date(workshop.start) <= new Date()) {
+        workshopTitle = workshop.title;
+      }
+    }
```
paologum, line 216: what if instead of actually deleting the workshop, we keep all workshops in the same table and just have a field marked `status`  with flags `expired` and `active`
whenever you add, you just enforce that `active` workshops have unique titles and that way whenever you get from the `workshop` table for users you don't have the additional overhead of removing it from the `workshop` table and adding it to the `workshop history` table

asivath: i forgot about this but aaron said that workshop titles shouldn't be unique, so I got rid of that now. I added the status column with an extra flag `inactive`

asivath: also what is our process for updating things that have been translated, or adding new things that need translation

server/db/migrations/2024.05.08-17.59.45-adjust-columns-in-verification-table.ts
```diff
@@ -12,6 +11,7 @@ export async function up(db: Kysely<any>): Promise<void> {
     .addColumn("verifyOrReset", "text", (col) => col.notNull())
     .execute();
   const oldData = await db.selectFrom("verification").selectAll().execute();
+  // This is actually wrong, you can't bulk insert data with Kysely
   if (oldData.length != 0) await db.insertInto("verification_new").values(oldData).execute();
```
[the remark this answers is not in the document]
paologum: ok now i see what you should do, so shouldn't you change this to that?

server/db/migrations/2024.09.17-10.58.43-drop-unique-constraint-for-workshop-title.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+  if (nonUniqueWorkshops.length > 0) {
+    await db.schema
+      .createTable("workshopBackup")
+      .addColumn("workshopId", "integer", (col) => col.autoIncrement().primaryKey())
+      .addColumn("title", "text", (col) => col.notNull())
+      .addColumn("start", "datetime", (col) => col.notNull())
+      .addColumn("end", "datetime", (col) => col.notNull())
+      .execute();
+    for (const workshop of nonUniqueWorkshops) {
```
paologum, line 53: wouldn't you want the up to use this if it has a workshop backup?

shared/src/index.ts
```diff
@@ -7,11 +7,12 @@ export * from "./constants.js";
 export * from "./util.js";
 export * from "./settings-types.js";
 
-export type Workshop = {
+export type SharedWorkshop = {
   title: string;
   start: string;
   end: string;
   workshopId: number;
+  active: boolean;
```
paologum, line 15: is there a reason that this isn't just the text status?
what if we want to specifically edit the status itself? i assume you did this for easier readability but if we edit workshops and we want to make a workshop inactive by editing it in the future route you will make won't this make it really difficult to discern if the `SharedWorkshop` itself doesn't have the status, so you will have to grab the workshop and have extra logic?

server/src/database.ts
```diff
@@ -49,6 +49,7 @@ Database Schema Details (Post-Migration):
   - title (text): Non-nullable, must be unique, the title of the workshop.
   - start (datetime): Non-nullable, the start date/time of the workshop. Make sure to always use ISO 8601 format. Easy way to get this is by using new Date().toISOString().
   - end (datetime): Non-nullable, the end date/time of the workshop. Make sure to always use ISO 8601 format. Easy way to get this is by using new Date().toISOString().
+  - status (string): Non-nullable, default is 'expired', indicates if the workshop is status or not. Possible values are 'status' and 'expired'.
```
paologum, line 52: change title to not have must be unique

server/src/database.ts
```diff
@@ -259,9 +262,31 @@ export async function deleteWorkshop(criteria: Partial<Workshop>) {
...
+ */
+export async function updateWorkshopStatusAndGetActive(): Promise<Workshop[]> {
+  const now = new Date().toISOString();
+  await db.updateTable("workshop").set({ status: "active" }).where("start", "<=", now).execute();
+  const endedWorkshops = await db
+    .updateTable("workshop")
+    .set({ status: "expired" })
+    .where("end", "<", now)
+    .where("status", "=", "active")
+    .returning("workshopId")
```
paologum, line 280: is it normal to update statuses on get requests to a table?
will this make overhead on those requests bigger than they need to be?

paologum (changes requested):

server/src/routes/login-routes.ts
paologum: i uploaded an excel file for new translations, we can add the other ones, but you can add new ones you make there and we can add the translations there so that we know things are done

server/db/migrations/2024.05.08-17.59.45-adjust-columns-in-verification-table.ts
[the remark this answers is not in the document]
asivath: previous migrations are not supposed to be changed. it doesn't really matter anyway since all migrations are past this point

shared/src/index.ts
asivath: i don't see a reason to let them edit the status from the frontend. the only real things someone should be able to edit are it's names and start/end time. it would make it easier to sort the workshops into categories on the frontend though so I'll add it.

paologum: i mean isn't the point of making it inactive but not expired to archive it until you decide on the exact time/ are undecided about items in the workshop, or maybe some other reason? not necessarily that you wanna delete it persay

server/src/database.ts
asivath: the alternative is doing two database calls instead of one which would involve more overhead. I don't think overhead is a concern anyway since this is not something a normal user would use and we're not going to be having tons of workshops. also this is likely to change in Jacqui's PR since the reason this is needed is to update users, but that won't be a thing in her PR.

paologum: how do we connect the user to a workshop in that pr again?

asivath: we would instead store user id's in the workshop table and if we wanted to know if a user was in a workshop we would perform a join

paologum: ok i guess that's okay then since jacqui's pr will take it out, and i'm assuming then the only updates to the status would be in crons?

shared/src/index.ts
asivath: inactive is meant for before the workshop has started, active is during the workshop, and expired is after the workshop. if you wanted to archive it or just keep it there, then you would just set a time far in the future., which you would have to do anyway to create the workshop

server/src/database.ts
asivath: if we did it with crons it would get kind of complicated. we would need to avoid redundant cron jobs, edit cron jobs when workshop times get edited, hard to detect failures. i think updates should be done whenever we try to fetch the data

paologum: hmm okay have you asked aaron about this?
this means every time any user wants to see what workshop they might be signed up to (in the future) we would check all the workshops? doesn't that seem like a ton of checks if we have users see their status?
for example with like 50 users and 10 workshops and with potential refreshing during a workshop, couldn't workshops cause the server to go crazy if they are all accessing and checking workshops?

asivath: i haven't talked to aaron about this but i don't think 50 users and 10 workshops is a crazy amount. better-sqlite3 is supposed to be able to handle thousands of queries per second, so the database itself wouldn't be the bottleneck

asivath: we could of course test this though with the stress test script and edit it to just keep trying to log in and get data.

shared/src/index.ts
paologum: hmm i feel like we should be more specific still about these statuses for clear usage when they happen
active -> during time frame
inactive -> not during time frame but not expired
archived -> show up in history or when you get all workshops but has the content decided but not the date/time ( this would not be shown to users for signing up, only for admins)
expired -> finished workshop

there is a use case i see where professor ogg might know the content and has datapacks planned but is not yet sure on when due to availability of himself or others if that makes sense. that use case would best fit into inactive but it depends on you picking a far fetched date and remembering that it's not actually scheduled for then at the same time

paologum, line 15: is there a reason for this since aaron said this is duplicate information?
since i'm assuming we could check active from `start` and `end`

asivath: aaron was talking about duplicate information in the db that would lead to needing to make updates in the db on fetches but i think sending this information to the frontend is fine since it makes it easier. the dates sent to the frontend are not normal date strings so they would have to be parsed into a different format than compared.

paologum (approved):

PR #260, opened 2024-09-16
moved about page pictures from app to server

asivath:
doesn't really seem to help with memory consumption but does get rid of some of the minification warnings


asivath:
this wasn't working initially because preview doesn't have it's own env variables, we probably need a task for that. also for using git pull instead of copying files, and checks for /confirm-deploy like trying to comment on a PR that doesn't exist

paologum (approved):

PR #264, opened 2024-09-22
Tsc 509 add database backend

lbostre:
- Added "account Type" column to user table. 
- Added admin route that allows admins to modify a user.
- Added tests to test modify user functionality


server/src/admin/admin-routes.ts
```diff
@@ -181,6 +182,40 @@ export const adminDeleteUser = async function adminDeleteUser(
...
+  if (!email || !validator.isEmail(email) || !accountType) {
+    reply.status(400).send({ error: "Missing/invalid required fields" });
+    return;
+  }
+  try {
+    const user = await checkForUsersWithUsernameOrEmail(username || email, email);
+    if (user.length == 0) {
+      reply.status(409).send({ error: "User does not exist." });
+      return;
+    }
```
paologum, line 209: should check if username or email exist before we `checkForUsernameOrEmail`
also need to register this in the actual routes and add body params

server/src/admin/admin-routes.ts
```diff
@@ -181,6 +182,40 @@ export const adminDeleteUser = async function adminDeleteUser(
   reply.send({ message: "User deleted" });
 };
 
+/**
+ * Admin sends a request to change a user's account type
+ * @param request
+ * @param reply
+ * @returns
+ */
+export const adminChangeAccountType = async function adminChangeAccountType(
```
paologum, line 191: need tests for this in `admin-routes.test.ts`

server/__tests__/admin-routes.test.ts
paologum, line 1: this file has an array of routes that you need to add your route to

paologum (changes requested):

server/src/admin/admin-routes.ts
```diff
@@ -180,6 +181,44 @@ export const adminDeleteUser = async function adminDeleteUser(
...
```
paologum, line 190: we probably don't want a route for every single change in an account, so this route shouldn't just be for `accountType` but for `isAdmin`

server/__tests__/admin-routes.test.ts
paologum, line 1: make sure you add the routes to the routes array

server/src/admin/admin-auth.ts
```diff
@@ -172,4 +182,9 @@ export const adminRoutes = async (fastify: FastifyInstance, _options: RegisterOp
     { schema: { body: adminCreateWorkshopBody }, config: { rateLimit: moderateRateLimit } },
     adminCreateWorkshop
   );
+  fastify.post(
+    "/user/account-type",
+    { schema: { body: adminChangeAccountTypeBody }, config: { rateLimit: moderateRateLimit } },
+    adminChangeAccountType
+  );
 };
```
paologum, line 190: you probably should use `fastify.patch("/user")` as the route

paologum (changes requested):

server/src/admin/admin-routes.ts
```diff
@@ -180,6 +181,43 @@ export const adminDeleteUser = async function adminDeleteUser(
   reply.send({ message: "User deleted" });
 };
 
+/**
+ * Admin sends a request to modify a user
+ * @param request
+ * @param reply
+ * @returns
+ */
+export const adminModifyUser = async function adminChangeAccountType(request: FastifyRequest, reply: FastifyReply) {
```
paologum, line 190: nitpicky but change `adminChangeAccountType` to `adminMondifyUser`

server/src/admin/admin-routes.ts
```diff
@@ -182,6 +183,43 @@ export const adminDeleteUser = async function adminDeleteUser(
...
+  try {
+    const user = await checkForUsersWithUsernameOrEmail(username, email);
+    if (user.length === 0) {
+      reply.status(409).send({ error: "User does not exist." });
+      return;
+    }
+
+    const updateData: { accountType?: string; isAdmin?: number } = {};
+    if (accountType) updateData.accountType = accountType;
+    if (isAdmin !== undefined) updateData.isAdmin = isAdmin;
```
paologum, line 214: you probably should make a type and make sure we only have certain set strings like `pro` and `default` so that we are for sure consistent

paologum (changes requested):

server/src/types.ts
```diff
@@ -19,6 +19,7 @@ export interface UserTable {
   invalidateSession: number;
   isAdmin: number;
   workshopId: number;
+  accountType: string;
```
paologum, line 22: this should be where the type is defined
i believe this should work
if we decide to change it in the future or open it up, a migration would need to be made but having this would enforce any adds to the user table

server/src/admin/admin-routes.ts
```diff
@@ -182,6 +183,44 @@ export const adminDeleteUser = async function adminDeleteUser(
...
 };
 
+/**
+ * Admin sends a request to modify a user
+ * @param request
+ * @param reply
+ * @returns
+ */
+export const adminModifyUser = async function adminModifyUser(request: FastifyRequest, reply: FastifyReply) {
+  type accountTypes = "pro" | "default";
```
paologum, line 193: this would move to types

paologum (changes requested):

server/src/types.ts
```diff
@@ -8,6 +8,8 @@ export interface Database {
   workshop: WorkshopTable;
 }
 
+export type accountTypes = "pro" | "default";
```
paologum, line 11: since `accountTypes` is a type and is only one type (not multiple types)  it should be pascal case and singular`AccountType`

server/src/admin/admin-routes.ts
```diff
@@ -182,6 +183,43 @@ export const adminDeleteUser = async function adminDeleteUser(
...
+ */
+export const adminModifyUser = async function adminModifyUser(request: FastifyRequest, reply: FastifyReply) {
+  const { username, email, accountType, isAdmin } = request.body as {
+    username: string;
+    email: string;
+    accountType?: accountTypes;
+    isAdmin?: number;
+  };
+
+  if (!email || !validator.isEmail(email) || (!accountType && isAdmin === undefined) || !username) {
```
paologum, line 200: you would need to add an `isAccountType` function here

paologum (changes requested):

server/__tests__/admin-routes.test.ts
```diff
@@ -244,7 +244,7 @@ const testAdminUser = {
   pictureUrl: "https://example.com/picture.jpg",
   isAdmin: 1,
   workshopId: 1,
-  accountType: "default"
+  accountType: "default" as AccountType
```
paologum, line 247: why does this need to be cast as `AccountType`?
would making `testAdminUser: User` make this better? i feel like there would be a better way to do this

paologum (approved):

PR #267, opened 2024-10-01
Tsc 528 edit workshop

asivath:
https://github.com/user-attachments/assets/f836c215-0398-4cd3-8f1f-711d0af996cf


server/src/admin/admin-routes.ts
```diff
@@ -609,3 +611,103 @@ export const adminCreateWorkshop = async function adminCreateWorkshop(
...
+    const newWorkshop = {
+      title: fieldsToUpdate.title ?? existingWorkshop.title,
+      start: fieldsToUpdate.start ?? existingWorkshop.start,
+      end: fieldsToUpdate.end ?? existingWorkshop.end
+    };
+    const identicalWorkshops = await findWorkshop(newWorkshop);
+    if (identicalWorkshops.length > 0) {
+      reply.status(409).send({ error: "Workshop with same title and dates already exists" });
+      return;
+    }
```
paologum, line 666: so now we are allowing identical titles ?

server/src/admin/admin-routes.ts
```diff
@@ -609,3 +611,103 @@ export const adminCreateWorkshop = async function adminCreateWorkshop(
...
+      startDate = new Date(start);
+      if (isNaN(startDate.getTime())) {
+        reply.status(400).send({ error: "Invalid start date" });
+        return;
+      }
+      fieldsToUpdate.start = startDate.toISOString();
+    }
+    if (end) {
+      const endDate = new Date(end);
+      if (isNaN(endDate.getTime()) || startDate.getTime() > endDate.getTime()) {
```
paologum, line 651: this should probably also not allow `start` and `end` to be the same time right

paologum (changes requested):

server/src/admin/admin-routes.ts
```diff
@@ -609,3 +611,103 @@ export const adminCreateWorkshop = async function adminCreateWorkshop(
...
+        reply.status(400).send({ error: "Invalid end date" });
+        return;
+      }
+      fieldsToUpdate.end = endDate.toISOString();
+    }
```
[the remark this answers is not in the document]
asivath: i can move it below the start check, but the end check needs the db query incase they don't give a start date

server/src/admin/admin-routes.ts
paologum: so now we can basically have the same workshop as long as one of the fields is different

paologum (approved):

PR #269, opened 2024-10-04
Tsc 486 generate under time settings

Changk820:


Changk820:
image
Without bookmark bar above, with bookmark bar below, slightly cut off my margin, but if I decrease the box above it's margin even more, looks a bit crammed
image

Changk820:
image
On just time page now

paologum (approved):

PR #270, opened 2024-10-04
Tsc 487 datapack tab between chart and settings

Michael-Knaack:
Added a Datapack tab between the home icon and chart tab on the navbar.  Works in both languages.


paologum (changes requested):
i think it should be before charts not after it

also make sure you upload images of what you changed on the front end whenever you make a pr

Screenshot 2024-10-05 at 1 48 00 PM

Michael-Knaack:
Screenshot 2024-10-11 011453
Swapped the location of datapacks on navbar

paologum:
i'll add the comment here but the meeting saw some bugs when selecting from the settings hover submenu and when generating a chart, the tab selected was incorrect, so you'll probably need to fix those things before you merge

paologum:
there's another small bug

https://github.com/user-attachments/assets/66e1c911-e1ea-442b-a8dd-6d4e45e79f2b

we monitor changes in settings that differ from the generated chart and push a snackbar that reflects that in the video above. since the tab changed, you'll also need to change the monitored tab that pushes the snackbar
i would suggest looking up that string to see where it is located in the repo

paologum:
one more thing is, and this might be slightly complicated but because we have a new "route" for datapacks, we need to show the discard,save changes on route changes so changing from `/datapack` to `/chart` or any difference in settings tab

for more context on this you can visit `DatapackProfile.tsx`

and look at this code
```
  useBlocker(
    ({ currentLocation, nextLocation }) =>
      isMountedRef &&
      state.datapackProfilePage.unsavedChanges &&
      currentLocation.pathname !== nextLocation.pathname &&
      !window.confirm(t("dialogs.confirm-changes.message"))
  );
  // for when user tries to leave page with unsaved changes
  useEffect(() => {
    const handleBeforeUnload = (event: BeforeUnloadEvent) => {
      if (state.datapackProfilePage.unsavedChanges) {
        event.preventDefault();
        event.returnValue = "";
      }
    };
    window.addEventListener("beforeunload", handleBeforeUnload);
    return () => {
      window.removeEventListener("beforeunload", handleBeforeUnload);
    };
  }, [state.datapackProfilePage.unsavedChanges]);
```
you may want to ask me, aditya, or jacqui if you have trouble with this or want more context

https://github.com/user-attachments/assets/eef594ca-c64c-4769-99c0-2a9ad92ecac6

paologum:
in this video, you can see a slight popup of the confirm changes when you generate a chart using a preset

https://github.com/user-attachments/assets/168ba37c-da47-4327-92b7-71fd1c6840b6

app/src/App.tsx
```diff
@@ -46,11 +47,15 @@ export default observer(function App() {
...
-    location.pathname === "/settings" &&
-    state.settingsTabs.selected !== "datapacks" &&
-    JSON.stringify(state.config.datapacks) !== JSON.stringify(state.unsavedDatapackConfig);
-
+  const checkUnsavedChanges = () => {
+    const isOnDatapacksTab = location.pathname === "/settings" && state.settingsTabs.selected === "datapacks";
+    const isOnDatapackPath = location.pathname === "/datapacks";
+    const hasUnsavedChanges = JSON.stringify(state.config.datapacks) !== JSON.stringify(state.unsavedDatapackConfig);
+    if (hasUnsavedChanges && !(isOnDatapackPath || isOnDatapacksTab)) {
+      return true;
```
paologum, line 55: there is an `isProcessDatapacks` boolean you can access that will not have that popup if it is currently configuring the new configuration

paologum (approved):

PR #272, opened 2024-10-07
Tsc 538 allow user to be in multiple workshops

JacquiFreeElectron:
### What's done: 

- created a joint table `usersWorkshops` to support many-to-many relationships between users and workshops
- dropped `workshopId` column from `users` table 
- implemented the frontend of edit and display user info / user-workshop relationships
- ~~instead of updating user, the handling ended workshops function now updates the `workshopHasEnded` in the `usersWorkshops` table~~

### What's not done:

- backend routes of edit user info
- need to add some test cases for add/delete user-workshop relationships


paologum:
add some demos

app/src/admin/AdminEditUserForm.tsx
```diff
@@ -0,0 +1,290 @@
...
+    email: data.email,
+    isAdmin: data.isAdmin,
+    pictureUrl: data.pictureUrl || null
+  });
+  const [selectedFile, setSelectedFile] = useState<File | null>(null);
+  const [originalUserInfo, setOriginalUserInfo] = useState(userInfo);
+  const workshops = data.workshopTitle ? data.workshopTitle : ["No registered workshop"];
+  const [currentWorkshops, setCurrentWorkshops] = useState(workshops);
+  const [selectedWorkshop, setSelectedWorkshop] = useState<string | null>(null);
+  const [openConfirmDialog, setOpenConfirmDialog] = useState(false);
```
paologum, line 42: can you make a seperate hook for all the state and handlers? then export it similar to `datapack-upload-form-hook.ts`

app/src/admin/AdminEditUserForm.tsx
```diff
@@ -0,0 +1,290 @@
...
+      [e.target.name]: e.target.value
+    });
+  };
+  const handleSaveChanges = () => {
+    if (selectedFile) {
+      const newAvatarUrl = URL.createObjectURL(selectedFile); // Preview URL for the uploaded image
+      setUserInfo({ ...userInfo, pictureUrl: newAvatarUrl });
+    }
+    setUnsavedChanges(false);
+    setIsEditing(false); // Quit edit mode
```
paologum, line 89: these handlers should also go there, i'm looking at this from an out perspective but i think it should be easily doable? you might need to pass in a few things, then you can return like `handlers, setters, state`

below too

server/src/admin/admin-routes.ts
```diff
@@ -110,8 +119,8 @@ export const adminCreateUser = async function adminCreateUser(request: FastifyRe
       pictureUrl: pictureUrl ?? null,
       isAdmin: isAdmin,
       emailVerified: 1,
-      invalidateSession: 0,
-      workshopId: 0
+      invalidateSession: 0
+      // workshopId: 0
```
paologum, line 123: comment

server/src/admin/admin-routes.ts
```diff
@@ -496,7 +510,22 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
...
+            const newRelationship = await checkWorkshopHasUser(userId, workshopId);
+            if (newRelationship.length !== 1) {
+              reply.status(500).send({ error: "Error adding user to workshop", invalidEmails: email });
+              return;
+            }
+          } else if (existingRelationship.length > 1) {
+            //TODO: add test
+            reply.status(500).send({ error: "Duplicated user-workshop relationship", invalidEmails: email });
+            return;
+          }
```
paologum, line 527: for robustness just make it an else

server/src/admin/admin-routes.ts
```diff
@@ -506,14 +535,22 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
...
+          uuid: randomUUID()
         });
         const newUser = await findUser({ email });
         if (newUser.length !== 1) {
           reply.status(500).send({ error: "Error creating user", invalidEmails: email });
           return;
         }
+        for (const eachUser of user) {
+          const { userId } = eachUser;
+          createUsersWorkshops({ userId: userId, workshopId: workshopId, workshopHasEnded: 0 });
```
paologum, line 547: you don't need an await here?

server/db/migrations/2024.09.22-14.19.29-add-users-workshops-join-table.ts
```diff
@@ -0,0 +1,21 @@
+import { type Kysely } from "kysely";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function up(db: Kysely<any>): Promise<void> {
+  await db.schema
+    .createTable("usersWorkshops")
+    .addColumn("workshopId", "integer", (col) => col.notNull())
+    .addColumn("userId", "integer", (col) => col.notNull())
+    .addColumn("workshopHasEnded", "integer", (col) => col.notNull().defaultTo(0))
```
paologum, line 9: i don't think this would be a good idea, because if we do this then everytime we edit the workshop (if we bring a past workshop to the front, we will have to check all of the users that have that workshop and edit them)
this could be something we want though but i feel like we can just check on the front end with all the workshops
aditya also might have thoughts on this

server/src/admin/admin-routes.ts
```diff
@@ -49,22 +52,28 @@ export const getUsers = async function getUsers(_request: FastifyRequest, reply:
...
+        const userWorkshops = await findUserInUsersWorkshops(userId);
+        const workshopTitle: string[] = [];
+        for (const userWorkshop of userWorkshops) {
+          const { workshopId } = userWorkshop;
           const workshop = await findWorkshop({ workshopId });
           if (workshop && workshop.length === 1) {
-            workshopTitle = workshop[0]?.title ?? "";
+            if (workshop[0]?.title) {
+              workshopTitle.push(workshop[0].title);
+            }
```
paologum, line 64: what do you think about splitting up workshops and users into two different routes
you can't actually see workshops until you open the form, so maybe it would be a good idea to not send workshop data until we need it. i'm not sure if the difference is negligible enough for us to just have it here though so let me know what you think.

server/src/routes/login-routes.ts
```diff
@@ -675,8 +678,8 @@ export const signup = async function signup(
       pictureUrl: null,
       emailVerified: 0,
       invalidateSession: 0,
-      isAdmin: 0,
-      workshopId: 0
+      isAdmin: 0
+      // workshopId: 0
```
paologum, line 682: comment

server/src/routes/login-routes.ts
```diff
@@ -800,8 +803,8 @@ export const googleLogin = async function googleLogin(
       pictureUrl: payload.picture,
       emailVerified: 1,
       invalidateSession: 0,
-      isAdmin: 0,
-      workshopId: 0
+      isAdmin: 0
+      //workshopId: 0
```
paologum, line 807: comment

server/src/types.ts
```diff
@@ -18,7 +19,7 @@ export interface UserTable {
   emailVerified: number;
   invalidateSession: number;
   isAdmin: number;
-  workshopId: number;
+  //workshopId: number;
```
paologum, line 22: comment

shared/src/index.ts
```diff
@@ -21,7 +21,7 @@ export type SharedUser = {
   pictureUrl: string | null;
   isGoogleUser: boolean;
   isAdmin: boolean;
-  workshopTitle?: string;
+  workshopTitle?: string[];
```
paologum, line 24: i would rename this `workshopsEnrolled`
for better understanding

server/src/database.ts
```diff
@@ -50,6 +51,11 @@ Database Schema Details (Post-Migration):
   - start (datetime): Non-nullable, the start date/time of the workshop. Make sure to always use ISO 8601 format. Easy way to get this is by using new Date().toISOString().
   - end (datetime): Non-nullable, the end date/time of the workshop. Make sure to always use ISO 8601 format. Easy way to get this is by using new Date().toISOString().
 
+- usersWorkshops Table:
+  - workshopId (integer): Non-nullable, links to the workshop table.
+  - userId (integer): Non-nullable, links to the users table.
+  - workshopHasEnded (integer): Non-nullable, default is 0, indicates if the workshop has ended
+
```
paologum, line 58: remove `workshopId` from these comments

paologum (changes requested):

JacquiFreeElectron:
https://github.com/user-attachments/assets/dadee106-77d2-417d-b106-5884a4927d33

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,151 @@
...
+  const [originalUserInfo, setOriginalUserInfo] = useState(userInfo);
+  const workshops = data.workshopEnrolled ? data.workshopEnrolled : ["No registered workshop"];
+  const [currentWorkshops, setCurrentWorkshops] = useState(workshops);
+  const [selectedWorkshop, setSelectedWorkshop] = useState<string | null>(null);
+  const [openConfirmDialog, setOpenConfirmDialog] = useState(false);
+
+  // Function to remove a workshop
+  const handleRemoveWorkshop = () => {
+    if (selectedWorkshop) {
+      const updatedWorkshops = currentWorkshops.filter((workshop) => workshop !== selectedWorkshop);
```
paologum, line 29: workshops can have the same name,
but they have to be unique in terms of end/start date AND name so this will not filter correctly for workshops of the same name

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,151 @@
...
+  };
+
+  const handleSaveChanges = () => {
+    if (selectedFile) {
+      const newAvatarUrl = URL.createObjectURL(selectedFile); // Preview URL for the uploaded image
+      setUserInfo({ ...userInfo, pictureUrl: newAvatarUrl });
+    }
+    setUnsavedChanges(false);
+    setIsEditing(false);
+  };
```
paologum, line 55: you should show me a demo of the profile pic uploading
additionally if they upload pro pic, it shouldn't preview after save changes, it should be previewed immediately after upload

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,151 @@
...
+    setOriginalUserInfo(userInfo);
+  };
+
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setUnsavedChanges(true);
+    setUserInfo({
+      ...userInfo,
+      [e.target.name]: e.target.value
+    });
+  };
```
paologum, line 46: to make this type safe, i would make a type on the front end that is like `EditableUserProperties` then set the type of the state of `UserInfo` and make sure that key `e.target.name` exists  in that type then set it

for dev robustness

app/src/components/edit-user-stats-hook.ts
paologum: shouldn't save changes also set original user info

shared/src/index.ts
```diff
@@ -21,7 +21,7 @@ export type SharedUser = {
   pictureUrl: string | null;
   isGoogleUser: boolean;
   isAdmin: boolean;
-  workshopTitle?: string[];
+  workshopEnrolled?: string[];
```
paologum, line 24: small thing but this needs to be `workshopsEnrolled` because it's multiple workshops

paologum: because titles are not unique as well, you might have to include more for uniqueness

paologum (changes requested):
the "graduation icon" should be before registered workshops, not before each workshop entry

server/src/admin/admin-routes.ts
```diff
@@ -540,7 +554,22 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
     for (const email of emailList) {
       const user = await checkForUsersWithUsernameOrEmail(email, email);
       if (user.length > 0) {
-        await updateUser({ email }, { workshopId });
+        for (const eachUser of user) {
+          const { userId } = eachUser;
+          const existingRelationship = await checkWorkshopHasUser(userId, workshopId);
+          if (existingRelationship.length == 0) {
+            createUsersWorkshops({ userId: userId, workshopId: workshopId });
```
paologum, line 561: await?

JacquiFreeElectron:
https://github.com/user-attachments/assets/cf644984-7df9-4484-b1e0-cbbf9f5532d2

server/src/admin/admin-routes.ts
```diff
@@ -472,7 +488,22 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
...
+            await createUsersWorkshops({ userId: userId, workshopId: workshopId });
+            const newRelationship = await checkWorkshopHasUser(userId, workshopId);
+            if (newRelationship.length !== 1) {
+              reply.status(500).send({ error: "Error adding user to workshop", invalidEmails: email });
+              return;
+            }
+          } else {
+            //TODO: add test
+            reply.status(500).send({ error: "Duplicated user-workshop relationship", invalidEmails: email });
+            return;
```
paologum, line 504: do you need to throw an error if the user is already in the workshop? i don't think you need to, you could just continue because they already are a part of the workshop? @asivath

paologum: also if you added test remove the TODO, if you haven't yet, then write the test

server/__tests__/admin-routes.test.ts
```diff
@@ -1223,7 +1253,14 @@ describe("getUsers", () => {
...
           emailVerified: true,
-          workshopTitle: testWorkshop.title
+          workshopsEnrolled: [
+            {
+              workshopId: testWorkshop.workshopId,
+              workshopTitle: testWorkshop.title,
+              start: testWorkshop.start,
+              end: testWorkshop.end
+            }
+          ] //
```
paologum, line 1263: // at the end

server/__tests__/admin-routes.test.ts
```diff
@@ -1702,7 +1740,12 @@ describe("adminAddUsersToWorkshop", () => {
     expect(response.statusCode).toBe(200);
   });
   it("should return 200 if successful and update old users", async () => {
```
paologum, line 1742: add test for when `checkWorkshopHasUser` (mentioned above) here is already in the junction table

server/__tests__/admin-routes.test.ts
```diff
@@ -1210,6 +1239,7 @@ describe("getUsers", () => {
   it("should return user with workshopTitle and one without", async () => {
```
paologum, line 1239: change this test title to reflect the test

server/src/admin/admin-routes.ts
```diff
@@ -52,22 +56,34 @@ export const getUsers = async function getUsers(_request: FastifyRequest, reply:
...
-        if (workshopId) {
+        const { hashedPassword, userId, ...displayedUser } = user;
+        const userWorkshops = await findUserInUsersWorkshops(userId);
+        const workshopsEnrolled: WorkshopEnrolled[] = [];
+        for (const userWorkshop of userWorkshops) {
+          const { workshopId } = userWorkshop;
           const workshop = await findWorkshop({ workshopId });
           if (workshop && workshop.length === 1) {
-            workshopTitle = workshop[0]?.title ?? "";
+            if (workshop[0]?.title) {
```
paologum, line 66: add these into the same if statement

server/__tests__/admin-routes.test.ts
paologum: additionally you should add your expect checks to adding new users, not just updating old ones. (the test above) `should return 200 if successful and add new users`

server/__tests__/admin-routes.test.ts
paologum, line 1: also need tests for 
```
 if (newRelationship.length !== 1) {
              reply.status(500).send({ error: "Error adding user to workshop", invalidEmails: email });
              return;
            }
```
when a new user is added and the other case when updating a user

server/src/admin/admin-routes.ts
```diff
@@ -472,7 +488,22 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
...
       if (user.length > 0) {
-        await updateUser({ email }, { workshopId });
+        for (const eachUser of user) {
+          const { userId } = eachUser;
+          const existingRelationship = await checkWorkshopHasUser(userId, workshopId);
+          if (existingRelationship.length == 0) {
+            await createUsersWorkshops({ userId: userId, workshopId: workshopId });
+            const newRelationship = await checkWorkshopHasUser(userId, workshopId);
+            if (newRelationship.length !== 1) {
+              reply.status(500).send({ error: "Error adding user to workshop", invalidEmails: email });
```
paologum, line 498: also @asivath if the first email is invalid for this reason, do we want to return all unadded emails or just the bad email and don't tell them the ones not added by not adding them to `invalidEmails`
i think we talked about this but i forgot what you had said

server/db/migrations/2024.09.22-14.19.29-add-users-workshops-join-table.ts
paologum, line 1: i can't bring up the pr preview because this migration is behind main since leyton pushed his migration which means this migration doesn't come next chronologically, it comes after leytons. so can you recreate this migration file to today's date so it comes after leyton since leyton is `CANON` now

paologum (changes requested):
there is something i might bring up in the meeting,
if we just remember which workshops that the admin deletes and DON'T put up a dialog until they press save changes or go through the unsaved changes dialog
that way they won't have to continuously press buttons for multiple deletions

server/src/routes/login-routes.ts
```diff
@@ -207,12 +208,20 @@ export const sessionCheck = async function sessionCheck(request: FastifyRequest,
...
       const workshop = await getAndHandleWorkshopEnd(workshopId);
       if (workshop && new Date(workshop.start) <= new Date()) {
-        workshopTitle = workshop.title;
+        const workshopEnrolled: WorkshopEnrolled = {
+          workshopId: workshopId,
+          workshopTitle: workshop.title,
+          start: workshop.start,
+          end: workshop.end
+        };
+        workshopsEnrolled.push(workshopEnrolled);
```
paologum, line 224: going to ask this in the meeting but i'll record my idea here for the time being, but i'm not sure if we need to have `WorkshopsEnrolled` on general users data.
We might just want to attach workshop ids then combine the two on the frontend because we will probably have to store all public workshops on the front end anyway

JacquiFreeElectron:
https://github.com/user-attachments/assets/2fd41330-c304-40a9-9b45-77df49732423

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,179 @@
...
+      setUserInfo(originalUserInfo); // Reset to original data
+      setSelectedFile(null); // Clear any selected file
+      setIsEditing(false);
+      setUnsavedChanges(false);
+      setShowDiscardDialog(false);
+      if (closeDialog) {
+        setMoreUsersInfoFormOpen(false);
+        setCloseDialog(false);
+      }
+    }
```
paologum, line 91: this method seems slightly confusing to me with how it's named and with the discard confirmed term used

i think you should just split this method into two methods, one that executes a discard changes action and just call it `discardChanges` and then the other case is not very clear so you might need to explain it for me

and also maybe could you explain why you chose this approach vs splitting into more defined functions

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,179 @@
...
+    } else {
+      setUserInfo(originalUserInfo); // Reset to original data
+      setSelectedFile(null); // Clear any selected file
+      setIsEditing(false);
+      setUnsavedChanges(false);
+      setShowDiscardDialog(false);
+      if (closeDialog) {
+        setMoreUsersInfoFormOpen(false);
+        setCloseDialog(false);
+      }
```
paologum, line 90: instead of having to remember to each `closeDialog` time with a variable that refreshes state, you should just add `setShowAdditionalUserInfoFormOpen` to `false` in each `handler` function that needs it.

shared/src/index.ts
```diff
@@ -21,10 +21,17 @@ export type SharedUser = {
   pictureUrl: string | null;
   isGoogleUser: boolean;
   isAdmin: boolean;
-  workshopTitle?: string[];
+  workshopsId?: number[];
```
paologum, line 24: `workshopIds`

otherwise `workshopsId` represents one workhop's id so it conveys a different meaning

paologum (changes requested):
Screenshot 2024-10-23 at 11 51 19 PM

dark mode should toggle like the others

Screenshot 2024-10-23 at 11 55 27 PM

should change when you change the theme, but here is a person with a workshop in dark mode

server/src/admin/admin-routes.ts
```diff
@@ -472,7 +485,21 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
     for (const email of emailList) {
       const user = await checkForUsersWithUsernameOrEmail(email, email);
       if (user.length > 0) {
-        await updateUser({ email }, { workshopId });
+        for (const eachUser of user) {
+          const { userId } = eachUser;
+          const existingRelationship = await checkWorkshopHasUser(userId, workshopId);
```
asivath, line 490: the variable `user` should always be only one user since email has to be unique, so this loop isn't needed

server/src/admin/admin-routes.ts
asivath: currently, before even starting to add/edit users in the database, we check for any invalid emails and just send invalid emails without adding any. this is slightly different since this could happen halfway through adding to the db. I think it would make more sense to continue with the batch update rather than stopping, and then send the invalidEmails at the end

server/src/admin/admin-routes.ts
```diff
@@ -483,14 +510,22 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
...
         }
+        for (const eachUser of newUser) {
+          const { userId } = eachUser;
+          await createUsersWorkshops({ userId: userId, workshopId: workshopId });
+          const newRelationship = await checkWorkshopHasUser(userId, workshopId);
+          if (newRelationship.length !== 1) {
+            reply.status(500).send({ error: "Error adding user to workshop", invalidEmails: email });
+            return;
+          }
+        }
```
asivath, line 528: same comment about the loop

server/src/types.ts
```diff
@@ -44,13 +44,21 @@ export interface WorkshopTable {
...
+
 export type User = Selectable<UserTable>;
 export type NewUser = Insertable<UserTable>;
 export type UpdatedUser = Updateable<UserTable>;
 
 export type Verification = Selectable<VerificationTable>;
 export type NewVerification = Insertable<VerificationTable>;
 
+export type UsersWorkshops = Selectable<UsersWorkshopsTable>;
+export type newUsersWorkshops = Insertable<UsersWorkshopsTable>;
```
asivath, line 60: capitalize this. also @paologum do you think we would ever need to update this table, or just deleting/creating?

server/src/database.ts
```diff
@@ -235,6 +236,42 @@ export async function checkForUsersWithUsernameOrEmail(username: string, email:
...
+    .execute();
+}
+
+export async function deleteWorkshopInUsersWorkshops(workshopId: number) {
+  return await db.deleteFrom("usersWorkshops").where("workshopId", "=", workshopId).execute();
+}
+
+export async function deleteUserInUsersWorkshops(userId: number) {
+  return await db.deleteFrom("usersWorkshops").where("userId", "=", userId).execute();
+}
```
asivath, line 269: same

server/src/database.ts
```diff
@@ -235,6 +236,42 @@ export async function checkForUsersWithUsernameOrEmail(username: string, email:
...
+    .execute();
+}
+
+export async function findUserInUsersWorkshops(userId: number) {
+  return await db
+    .selectFrom("usersWorkshops")
+    .selectAll()
+    .where((eb) => eb("userId", "=", userId))
+    .execute();
+}
```
asivath, line 261: you can combine these

app/src/components/edit-user-stats-hook.ts
JacquiFreeElectron: I built upon the existing code, as there was initially only one function managing whether the "Are you sure" dialog should open or not. I can split this into two parts for sure—one part to showing the dialog (one function) and the other part to handle the discard process. And as you suggested below I’ll create separate handler functions for each type of discard request (since each discard has a different effect depending on where it was triggered) instead of using a single function for all discards. But I think I still need to pass in some kind of boolean value to the `showDialog` handler so that it knows which ` discardChange` handler to call. Lmk if I’ve misunderstood anything or you have any better idea

server/src/admin/admin-routes.ts
JacquiFreeElectron: So this is what I should do:
1. Process the users and add/edit them in the database, if some emails are invalid, skip it and continue the loop.
2. Keep track of the invalid emails while process the users.
3. Once all users have been processed, return a response that includes the invalid emails (if any).
Is this correct?

asivath: yes

app/src/components/edit-user-stats-hook.ts
JacquiFreeElectron: After further consideration, I ended up with two "are you sure" dialogs (with the same content). I couldn't find a way to pass the correct handler function for different situations without using a state variable to keep track of closeDialog (now renamed to closeForm, since it actually closes the entire form). The issue arises because both clicking the close button of the form and clicking the discard button trigger showDiscardDialog, making it difficult to identify which action opened the dialog. Using a boolean or state variable felt similar to the previous approach, so I decided that a better solution would be to use two separate components, allowing me to assign different handlers to onYes. However, I still need a state variable to determine which dialog should be displayed—one dialog uses open={editState.showDiscardDialog}, while the other uses open={editState.showDiscardDialogAndCloseTheForm}. Overall, I find this approach to be more readable. what you think of this approach

JacquiFreeElectron: it looks like this now: 
```
<TSCYesNoPopup   //open when the close icon is clicked or on backdrop click
        open={editState.showDiscardDialogAndCloseForm} 
        title="You have unsaved changes. Are you sure you want to discard them?"
        onYes={() => handlers.handleDiscardChangesAndCloseTheForm}
        onNo={handlers.handleCloseDiscardDialog}
        onClose={handlers.handleCloseDiscardDialog}
      />
<TSCYesNoPopup //open when the Discard button is clicked
        open={editState.showDiscardDialog}
        title="You have unsaved changes. Are you sure you want to discard them?"
        onYes={() => handlers.handleDiscardChanges}
        onNo={handlers.handleCloseDiscardDialog}
        onClose={handlers.handleCloseDiscardDialog}
      />
```

paologum: yeah i think this is good, separating the two into two separate popups for two separate actions

paologum: yeah i think this is good, separating the two into two separate popups for two separate actions good work ! ty for explaining as well

server/src/types.ts
paologum: hmm i don't think we would need to change this imo

paologum:
this is still not dark mode toggled (use `ag-grid`) style switch like the others

Screenshot 2024-10-29 at 8 54 16 PM

paologum:
these should be the same for consistency 

Screenshot 2024-10-29 at 8 55 33 PM

Screenshot 2024-10-29 at 8 55 47 PM

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,196 @@
...
+  data: AdminSharedUser;
+};
+
+const useEditUser = ({ data }: UseUserStatsProps) => {
+  const workshops = data.workshopIds;
+  const [moreUsersInfoFormOpen, setMoreUsersInfoFormOpen] = useState(false);
+  const [isEditing, setIsEditing] = useState(false);
+  const [unsavedChanges, setUnsavedChanges] = useState(false);
+  const [showDiscardDialog, setShowDiscardDialog] = useState(false);
+  const [showDiscardDialogAndCloseForm, setshowDiscardDialogAndCloseForm] = useState(false);
```
paologum, line 16: change to `setShowDiscardDialogAndCloseForm`

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,196 @@
...
+import { SelectChangeEvent } from "@mui/material/Select";
+
+type UseUserStatsProps = {
+  data: AdminSharedUser;
+};
+
+const useEditUser = ({ data }: UseUserStatsProps) => {
+  const workshops = data.workshopIds;
+  const [moreUsersInfoFormOpen, setMoreUsersInfoFormOpen] = useState(false);
+  const [isEditing, setIsEditing] = useState(false);
```
paologum, line 13: for consistency across forms change this to `editMode` and `setEditMode`

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,196 @@
...
+  const [showDiscardDialogAndCloseForm, setshowDiscardDialogAndCloseForm] = useState(false);
+  const [userInfo, setUserInfo] = useState<EditableUserProperties>({
+    username: data.username,
+    email: data.email,
+    isAdmin: data.isAdmin,
+    pictureUrl: data.pictureUrl || undefined
+  });
+  const [selectedFile, setSelectedFile] = useState<File | null>(null);
+  const [originalUserInfo, setOriginalUserInfo] = useState(userInfo);
+  const [currentWorkshops, setCurrentWorkshops] = useState<number[] | undefined>(workshops);
```
paologum, line 25: when is `currentWorkshops` ever `undefined`

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,196 @@
...
+    if (e.target.files && e.target.files[0]) {
+      setSelectedFile(e.target.files[0]);
+    }
+    setUnsavedChanges(true);
+  };
+
+  const handleOpenConfirmDialog = (id: number) => {
+    setSelectedWorkshop(id);
+    setOpenConfirmDialog(true);
+  };
```
paologum, line 116: with the amount of dialogs i would rename `handleOpenConfirmDialog` to specifically reference the workshops like `handleOpenConfirmRemovalOfUserFromWorkshop`

additionally change the `setOpenConfirmRemovalOfUserFromWorkshopDialog`

it is long but because its ambiguity is gone it's easier to understand

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  const handleCloseConfirmDialog = () => {
+    setOpenConfirmDialog(false);
+    setSelectedWorkshop(null);
+  };
+
```
paologum, line 122: same here

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,196 @@
...
+  const handleCloseDiscardDialog = () => {
+    setShowDiscardDialog(false);
+    setshowDiscardDialogAndCloseForm(false);
+  };
```
paologum, line 126: for all three of these methods I would change the name to aptly describe them

app/src/components/edit-user-stats-hook.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  const handleCloseForm = () => {
+    if (unsavedChanges) {
+      //setCloseDialog(true);
+      //setShowDiscardDialog(true);
+      setshowDiscardDialogAndCloseForm(true);
+    } else {
+      handleDiscardChangesAndCloseTheForm();
+    }
+  };
```
paologum, line 136: comments and also be more specific about which form for all 3 of these methods as well

server/src/admin/admin-routes.ts
```diff
@@ -490,10 +497,23 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
...
+          await createUsersWorkshops({ userId: userId, workshopId: workshopId });
+          const newRelationship = await checkWorkshopHasUser(userId, workshopId);
+          if (newRelationship.length !== 1) {
+            invalidEmails.push(email);
+            continue;
+          }
+        } else {
+          duplicatedEmails.push(email);
+          continue;
+        }
```
paologum, line 516: i don't think we want to send errors on `duplicatedEmails`unless someone else said so
@asivath can you confirm?

shared/src/index.ts
```diff
@@ -658,8 +658,11 @@ export function assertSharedUser(o: any): asserts o is SharedUser {
   if (typeof o.isGoogleUser !== "boolean") throwError("User", "isGoogleUser", "boolean", o.isGoogleUser);
   if (typeof o.isAdmin !== "boolean") throwError("User", "isAdmin", "boolean", o.isAdmin);
   if (typeof o.uuid !== "string") throwError("User", "uuid", "string", o.uuid);
-  if (o.workshopTitle != null && typeof o.workshopTitle !== "string")
-    throwError("User", "workshopTitle", "number", o.workshopTitle);
+  if (o.workshopIds != null) {
```
paologum, line 661: `workshopIds` isn't able to be `null` it can only be `undefined`

paologum (changes requested):

app/src/admin/AdminShowAdditionalUserInfo.tsx
```diff
@@ -0,0 +1,265 @@
...
+    });
+
+    const timePart = date.toLocaleTimeString(undefined, {
+      hour: "numeric",
+      minute: "2-digit",
+      hour12: true
+    });
+
+    return `${datePart} at ${timePart}`;
+  };
```
paologum, line 51: because this is a format kept consistent in two areas, i would make a function in `non-action-util.ts` that accepts either a string or a `dayjs` object and formats it according to this (for a string) and the other way (for a dayjs) object

server/src/admin/admin-routes.ts
asivath: yep don't send

app/src/components/edit-user-stats-hook.ts
JacquiFreeElectron: when a user is not enrolled in any workshop

JacquiFreeElectron:
> handleDiscardChangesAndCloseTheForm
Used `backgroundColor: theme.palette.mode === "dark" ? "rgba(26, 34, 45, 0.7)" : "rgba(255, 255, 255, 0.7)"` instead cause ag-grid style switch doesn't work on `TableContainer`, the effect is satisfying imo: 
Screenshot 2024-11-01 at 10 14 42 PM
Screenshot 2024-11-01 at 10 15 01 PM

JacquiFreeElectron:
dark mode:
Screenshot 2024-11-01 at 10 30 00 PM
Other things stay the same as the last demo.
Tried my best to rename handlers/setters lmk if some of them don't make sense :)

paologum:
> > this is still not dark mode toggled (use `ag-grid`) style switch like the others
> 
> > 
> 
> > Screenshot 2024-10-29 at 8 54 16 PM
> 
> 
> 
> 
> 
> Used `backgroundColor: theme.palette.mode === "dark" ? "rgba(26, 34, 45, 0.7)" : "rgba(255, 255, 255, 0.7)"` instead cause ag-grid style switch doesn't work on `TableContainer`, the effect is satisfying imo: 
> 
> Screenshot 2024-11-01 at 10 14 42 PM
> 
> Screenshot 2024-11-01 at 10 15 01 PM
> 
> 

i sent you a slack message about figuring out if we can apply the theme here/how you initially tried it
but if we decide that we can't then you would need to put those colors in the actual palette (as two different colors i think)

server/__tests__/admin-routes.test.ts
```diff
@@ -1671,29 +1714,107 @@ describe("adminAddUsersToWorkshop", () => {
...
-        email: "test2@gmail.com"
-      },
-      { workshopId: 1 }
-    );
     expect(createUser).not.toHaveBeenCalled();
     expect(findUser).toHaveBeenCalledTimes(1); // 1st call is from the prehandler verifyAdmin
     expect(await response.json()).toEqual({ message: "Users added" });
     expect(response.statusCode).toBe(200);
   });
+  it("should return 500 if unable to update old user", async () => {
```
paologum, line 1731: same as above

paologum (changes requested):
some last cleanups, looks good!

app/src/state/non-action-util.ts
```diff
@@ -38,3 +39,27 @@ export function getDatapackProfileImageUrl(datapack: Datapack) {
     return devSafeUrl(`/datapack-images/default/${uuid}`);
   }
 }
+
+export function formatDate(input: string | dayjs.Dayjs): string {
```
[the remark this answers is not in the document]
JacquiFreeElectron: not yet. he uses `valueFormatter: (params) => dayjs(params.value).format("MMMM D, YYYY [at] h:mm A")`, I'm unsure how to call the function I created instead. Do i just do `valueFormatter: (params) =>formatDate(params.value)`?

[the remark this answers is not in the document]
paologum: i believe it's a string unless i'm wrong so i think that's good
just verify it's the same and you should be chilling!

server/src/admin/admin-routes.ts
```diff
@@ -497,18 +497,21 @@ export const adminAddUsersToWorkshop = async function addUsersToWorkshop(request
       reply.status(409).send({ error: "Invalid email addresses provided", invalidEmails: invalidEmails.join(", ") });
       return;
     }
+    const addedNewUserWorkshopRelationship = async (userId: number, workshopId: number, email: string) => {
```
paologum, line 515: i would change this to `addNewUserWorkshopRelationship`

app/src/components/edit-user-stats-hook.ts
```diff
@@ -102,28 +102,28 @@ const useEditUser = ({ data }: UseUserStatsProps) => {
...
     if (unsavedChanges) {
-      setOpenConfirmDiscardUserInfoChangeAndCloseForm(true);
+      setIsConfirmUserInfoChangesOnExitFormOpen(true);
     } else {
       handleDiscardUserInfoChangesThenCloseTheMoreUserInfoForm();
     }
   };
-  const handleOpenConfirmDiscardUserInfoChange = () => {
-    setOpenConfirmDiscardUserInfoChange(true);
+  const handleisConfirmDiscardUserInfoChangeOpen = () => {
```
paologum, line 125: you should still capitalize the i to `handleIs...`

paologum (changes requested):
two minor name changes, then i will approve!

paologum (approved):
i wanna make sure, but when a workshop

PR #273, opened 2024-10-11
Tsc 520 changes to account menus

sbal15:
pic1
image

removed dark/light mode & language from profile
added preferences tab

### DIFF
--- a/app/src/NavBar.tsx
+++ b/app/src/NavBar.tsx
@@ -261,6 +261,13 @@ export const NavBar = observer(function Navbar() {
                 <MenuItem>
                   <Typography sx={{ textAlign: "center", width: "100%" }}>{t(`LANGUAGES`)}</Typography>
                 </MenuItem>
+                <Divider
+                  sx={{
+                    borderBottomWidth: 0.75,
+                    borderColor: "gray",
+                    my: 1
+                  }}
+                />
                 {Object.entries(languageList).map(([key, value]) => (
                   <MenuItem
                     ref={dropRef}
--- a/app/src/NavBar.tsx
+++ b/app/src/NavBar.tsx
@@ -30,7 +30,6 @@ import { SettingsMenuOptionLabels, assertSettingsTabs } from "./types";
 import Color from "color";
 import { AccountMenu } from "./account_settings/AccountMenu";
 import { toJS } from "mobx";
-import LanguageIcon from "@mui/icons-material/Language";
 import i18next from "i18next";
 import { useTranslation } from "react-i18next";
 import Button from "@mui/material/Button";
--- a/app/src/NavBar.tsx
+++ b/app/src/NavBar.tsx
@@ -31,7 +31,6 @@ import { AccountMenu } from "./account_settings/AccountMenu";
 import { toJS } from "mobx";
 import i18next from "i18next";
 import { useTranslation } from "react-i18next";
-import MenuItem from "@mui/material/MenuItem";
 import React from "react";
 import languageList from "../translation/avaliable-language.json";
 import Switch from "@mui/material/Switch";
@@ -239,7 +238,7 @@ const NonUserSettings: React.FC = () => {
         backgroundColor: theme.palette.dark.light,
         border: `1px solid ${theme.palette.divider}`
       }}>
-      <MenuItem>
+      <TSCMenuItem>
         <CustomFormControlLabel
           width={120}
           control={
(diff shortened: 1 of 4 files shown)


paologum:
because language and dark mode are not directly correlated, i would probably move towards a hover menu button labeled `Language` that opens a second dropdown to the side that has the languages

sbal15:
image

sbal15:
image

app/src/NavBar.tsx
```diff
@@ -179,6 +160,70 @@ export const NavBar = observer(function Navbar() {
             }}
           />
         )}
+
+        <PopupState variant="popover" popupId="demo-popup-menu">
+          {(popupState) => (
+            <React.Fragment>
+              <Button variant="text" {...bindTrigger(popupState)}>
+                <SettingsIcon />
+                <Typography>{t(``)}</Typography>
```
paologum, line 169: if there is not translation just remove the `Typography`

app/src/NavBar.tsx
```diff
@@ -179,6 +160,70 @@ export const NavBar = observer(function Navbar() {
...
+              </Button>
+              <Menu {...bindMenu(popupState)}>
+                <MenuItem className="settings-sub-menu-item" onClick={() => {}}>
+                  <Typography>{t(``)}</Typography>
```
paologum, line 173: same here, also the above `onClick is empty so remove it

app/src/NavBar.tsx
```diff
@@ -179,6 +160,70 @@ export const NavBar = observer(function Navbar() {
             }}
           />
         )}
+
+        <PopupState variant="popover" popupId="demo-popup-menu">
```
paologum, line 164: remove the `demo-popup-menu`

app/src/NavBar.tsx
```diff
@@ -179,6 +160,70 @@ export const NavBar = observer(function Navbar() {
...
+                />
+                <MenuItem>
+                  <Typography sx={{ textAlign: "center", width: "100%" }}>{t(`LANGUAGES`)}</Typography>
+                </MenuItem>
+                <Divider
+                  sx={{
+                    borderBottomWidth: 0.75,
+                    borderColor: "gray",
+                    my: 1
+                  }}
```
paologum, line 203: move the `sx` to the `css` file

paologum (changes requested):

app/src/NavBar.tsx
sbal15: when i try to remove the onclick, it throws an error, would you know why?

paologum: does this menu item have any use if it is empty with an empty onClick function? what is the use of this

app/src/NavBar.tsx
```diff
@@ -179,6 +160,55 @@ export const NavBar = observer(function Navbar() {
...
+                  />
+                </MenuItem>
+                <Divider className="div-bar-top" />
+                <MenuItem>
+                  <Typography>{t(`LANGUAGES`)}</Typography>
+                </MenuItem>
+                <Divider className="div-bar-bottom" />
+                {Object.entries(languageList).map(([key, value]) => (
+                  <MenuItem
+                    ref={dropRef}
```
paologum, line 192: does this ref have a function? it just is being declared but never used

paologum (changes requested):

app/src/NavBar.tsx
sbal15: oops sorry, forgot to take it out, i was experimenting with the drop down menu

paologum:
tab selection seems to be buggy on the pr preview, there is probably a bug in this branch related to this because on local and dev sites this is not the case

https://github.com/user-attachments/assets/00f5389d-aa3e-4c0f-ac13-d7b6cfbab098

scripts/json-to-excel.ts
paologum, line 1: don't commit this

app/src/NavBar.tsx
```diff
@@ -110,19 +88,19 @@ export const NavBar = observer(function Navbar() {
...
+              <Tab
+                value={3}
+                disableRipple
+                label={t("navBar.settings")}
+                to="/settings"
+                component={Link}
+                ref={settingsRef}
+                {...anchorProps}
+              />
+              <Tab value={3} disableRipple label={t("navBar.help")} to="/help" component={Link} />
```
paologum, line 102: same tab value makes selection of tabs off

paologum (changes requested):
looks like you removed the mobile functionality from the pr, you will need to add that back

paologum (approved):

PR #274, opened 2024-10-11
Tsc 550 fix types and file system

paologum:
# Main

* large backend changes
* changed file system to be
```
uploads
├── private
│   └── [uuid]
├── public
│   └── [uuid]

```
* where uuid can be `server` or `workshop`
* changed `metadata` types to include the `server, workshop, or user` types
* combined `private_user` and `public_user` to be the same type
* added a new script called with `yarn migrate-datapacks`, which "migrates" the datapacks by reparsing them and rewriting/reusing the old cache (might want to add this to dev but i'm not sure yet)
* added `yarn dev-config --overwrite` to the `yarn dev` script so that the datapacks are fetched correctly
* `yarn dev-config` now moves datapacks in the `admin-config` and `asset/datapacks` dir to their correct places (via the types in `admin-config.json`) in either `public` or `private` when you run `yarn dev` or`yarn dev-config`
* made functions for getting directories and datapacks more modular for easier testing and understanding
* when fetching datapacks, it will fetch private user datapacks IF signed in, but I combined all of the public fetching into one, which will need optimization (fetches both server and uuid datapacks at the same time)
* removed indexes and uses datapack array to store datapacks
* ONLY use `addDatapack` and `removeDatapack` to modify this (or alternatively the setter)
* large testing changes to accomodate this
* don't store anything in memory anymore, will fetch directly from the file system.
* parses cached datapack json files in each uuid directory
Screenshot 2024-10-11 at 10 27 02 PM

* server datapack upload form now has an `isPublic` checkbox for making private and public datapacks (no editability) yet


server/src/routes/user-routes.ts
```diff
@@ -228,29 +179,30 @@ export const requestDownload = async function requestDownload(
...
 
   try {
-    await access(maybeEncryptedFilepath);
-    const file = await readFile(maybeEncryptedFilepath);
+    await access(encryptedFilepath);
+    const file = await readFile(encryptedFilepath);
 
-    if (await checkHeader(maybeEncryptedFilepath)) {
+    if (await checkHeader(encryptedFilepath)) {
+      console.log("here 3");
```
asivath, line 201: remove

server/src/routes/user-routes.ts
```diff
@@ -194,17 +144,18 @@ export const requestDownload = async function requestDownload(
   }
   // see if we have already encrypted the file
   try {
-    await access(maybeEncryptedFilepath);
-    const file = await readFile(maybeEncryptedFilepath);
-    if (await checkHeader(maybeEncryptedFilepath)) {
+    await access(encryptedFilepath);
```
asivath, line 147: redundant since you readfile anyway

server/src/util.ts
```diff
@@ -256,6 +256,18 @@ export async function verifyFilepath(filepath: string) {
...
+    filepath = path.resolve(filepath);
+    if (!filepath.startsWith(process.cwd())) {
+      return false;
+    }
+    return false;
+  } catch {
+    return true;
+  }
+}
+
```
asivath, line 270: what is going on here?

paologum: filepath is a `path.join` string and when we resolve i want to check if they have any weird `..` resolutions
since it doesn't exist i can't use realpath

asivath (approved):

PR #275, opened 2024-10-18
added route to service mappoints directly

Changk820:


server/src/index.ts
Changk820, line 1: Aditya said to incase you were wondering why I did this.

Changk820: For removing a route serving aboutImages

Changk820:
I changed the tests because I removed imagesDirectory throughout the repo. Can you take a look at the tests to make sure I didn't do something sus?

The way my code is getting the new map.img path is by taking a previously existing string I found and regexing that, then appending the image name.

Also some things I noticed:
It seems if public and private have the same map point images, only one will show up.

I'm not sure how to create my own map images. I just took a public data pack and uploaded as a private for now to test the map points functionality.

I'm pretty sure this new folder structure has broken the version checking mechanism. On server start up, my images folders keeps getting overwritten. I'm pretty sure originally, it was supposed to make a single version.info file, now every pack has a version.info file, but I feel like this shouldn't break the version checking. Might need to make a new issue for this. Maybe I'm going crazy though.

docs/server.md
paologum, line 1: this file isn't connected to the actual  docs, so you should change the docs branch itself instead of this, you can remove these changes. if you want you can get rid of this docs folder

server/src/index.ts
```diff
@@ -122,11 +122,18 @@ server.register(fastifyStatic, {
...
+  root: process.cwd() + "/assets/uploads",
+  prefix: "/assets/uploads",
+  decorateReply: false
+});
+
+// Serve the mappoint images from server/assets/uploads]
+server.register(fastifyStatic, {
+  root: process.cwd() + "/assets/decrypted",
+  prefix: "/assets/decrypted",
+  decorateReply: false
```
paologum, line 136: this doesn't exist anymore `assets/decrypted` is not something we support anymore

server/src/index.ts
```diff
@@ -122,11 +122,18 @@ server.register(fastifyStatic, {
...
-// Serve the about picture images from server/public/aboutPictures/
+// Serve the mappoint images from server/assets/uploads]
 server.register(fastifyStatic, {
-  root: process.cwd() + "/public/aboutPictures",
-  prefix: "/public/aboutPictures/",
-  decorateReply: false // first registration above already added the decorator
+  root: process.cwd() + "/assets/uploads",
+  prefix: "/assets/uploads",
+  decorateReply: false
+});
```
paologum, line 130: i wouldn't service the whole uploads folder using static, if you're gonna do this you have to make it specific. we don't want to make peoples files available to everyone

paologum (changes requested):

paologum:
between public and private, there should only ever be one datapack with a unique title. each datapack has public and uuid to make it easily understandable where the datapack is in the file system. you still need to create a route that SPECIFICALLY fetches the image. not a general route that gives the user the options to select from a large amount of files

server/src/index.ts
Changk820: Is  the new change what you meant?

server/src/index.ts
```diff
@@ -122,18 +122,27 @@ server.register(fastifyStatic, {
...
+  allowedPath: (pathName, _root, req) => {
+    const uuid = req.session.get("uuid");
+    if (!uuid) {
+      return false;
+    }
+
+    const uuidFolder = pathName.split("/")[1];
+    return uuidFolder === uuid;
+  },
+  decorateReply: false,
```
paologum: this needs to be specific again. this needs to ONLY service the map pack images in the decrypted folder if they exist

server/src/index.ts
```diff
@@ -120,11 +121,44 @@ server.register(fastifyStatic, {
...
+  session: {
+    get: (key: string) => string | undefined;
+  };
+}
+
+/* Utility to validate the path of mappoint images for public and private
+ * Path must contain 'MapImages' and end with an image extension */
+const isValidMapImagePath = (pathName: string): boolean => {
+  const pathSegments = pathName.split("/");
+  const mapImagesIndex = pathSegments.indexOf("MapImages");
```
paologum, line 134: check that the second to last directory is `MapImages`

server/src/index.ts
```diff
@@ -120,11 +121,44 @@ server.register(fastifyStatic, {
...
+};
+
+const assetPaths = ["public", "private"];
+assetPaths.forEach((type) => {
+  server.register(fastifyStatic, {
+    root: path.join(process.cwd(), `assets/uploads/${type}`),
+    prefix: `/assets/uploads/${type}`,
+    decorateReply: false,
+    allowedPath: (pathName, _root, req: Request) => {
+      const fullPath = path.join(_root, pathName);
```
paologum, line 156: verify

server/src/index.ts
```diff
@@ -120,11 +121,44 @@ server.register(fastifyStatic, {
...
+  const [uuidFolder] = pathSegments.slice(1); // Extract UUID folder
+  // Ensure UUID matches and path is valid
+  return uuidFolder === uuid && isValidMapImagePath(pathName);
+};
+
+const assetPaths = ["public", "private"];
+assetPaths.forEach((type) => {
+  server.register(fastifyStatic, {
+    root: path.join(process.cwd(), `assets/uploads/${type}`),
+    prefix: `/assets/uploads/${type}`,
```
paologum, line 153: tbh i would probably lean towards not requiring a prefix of `/assets/uploads/` so others don't know our filesystem pathing. i would instead prefix like `${type}/${imageName}` or however, then construct it in your method
that way they don't actually need to know the path and we can just change the path if we change our filesystem
this ends up basically being a regular route but i think this would make the most sense
lmk if that makes sense

paologum: i'm mostly saying this because how you're checking how the file is constructed is basically making the path itself, so given an image name and a type then you can service it by constructing the path
sorry my directions for this task were ambiguous that's my bad

server/src/index.ts
```diff
@@ -120,11 +121,49 @@ server.register(fastifyStatic, {
...
+    allowedPath: (pathName, _root, req: Request) => {
+      const fullPath = path.join(_root, pathName);
+      // Verify the full path to prevent path traversal.
+      const isVerified = verifyFilepathSync(fullPath);
+      if (!isVerified) {
+        return false;
+      }
+      // Validate path based on route type | Private: Check UUID and path | Public: Only check path
+      const isAllowed = type === "private" ? isAllowedPrivatePath({ pathName, req }) : isValidMapImagePath(pathName);
+      return isAllowed && fs.existsSync(fullPath);
```
paologum, line 164: not sync

server/src/index.ts
```diff
@@ -120,11 +121,49 @@ server.register(fastifyStatic, {
...
+const assetPaths = ["public", "private"];
+assetPaths.forEach((type) => {
+  server.register(fastifyStatic, {
+    root: path.join(process.cwd(), `assets/uploads/${type}`),
+    prefix: `/assets/uploads/${type}`,
+    decorateReply: false,
+    allowedPath: (pathName, _root, req: Request) => {
+      const fullPath = path.join(_root, pathName);
+      // Verify the full path to prevent path traversal.
+      const isVerified = verifyFilepathSync(fullPath);
```
paologum, line 158: not sync

server/src/util.ts
```diff
@@ -259,6 +259,19 @@ export async function verifyFilepath(filepath: string) {
...
+    filepath = path.resolve(filepath);
+    if (!filepath.startsWith(root)) {
+      return false;
+    }
+  } catch {
+    return false;
+  }
+  return true;
+}
+
```
paologum, line 274: we shouldn't need this

Changk820: The original one was async. Would it not cause problems if I attempt to use it synchronously?

VS code complained at me, doing this

`
    allowedPath : async (pathName, _root, req: Request) => {
      const fullPath = path.join(_root, pathName);
      // Verify the full path to prevent path traversal.
      const isVerified = await verifyFilepath(fullPath);
      if (!isVerified) {
        return false;
      }
      // Validate path based on route type | Private: Check UUID and path | Public: Only check path
      const isAllowed = type === "private" ? isAllowedPrivatePath({ pathName, req }) : isValidMapImagePath(pathName);
      return isAllowed && fs.existsSync(fullPath);
    }
  }); 
`

allowed path says
`No overload matches this call.
  Overload 1 of 3, '(plugin: FastifyPluginCallback<FastifyStaticOptions, RawServerDefault, FastifyTypeProviderDefault, FastifyBaseLogger>, opts?: FastifyRegisterOptions<...> | undefined): FastifyInstance<...> & PromiseLike<...>', gave the following error.
    Type '(pathName: string, _root: string, req: Request) => Promise<boolean>' is not assignable to type '(pathName: string, root: string, request: FastifyRequest<RouteGenericInterface, RawServerDefault, IncomingMessage, FastifySchema, FastifyTypeProviderDefault, unknown, FastifyBaseLogger, ResolveFastifyRequestType<...>>) => boolean'.
      Type 'Promise<boolean>' is not assignable to type 'boolean'.
  Overload 2 of 3, '(plugin: FastifyPluginAsync<FastifyStaticOptions, RawServerDefault, FastifyTypeProviderDefault, FastifyBaseLogger>, opts?: FastifyRegisterOptions<...> | undefined): FastifyInstance<...> & PromiseLike<...>', gave the following error.
    Type '(pathName: string, _root: string, req: Request) => Promise<boolean>' is not assignable to type '(pathName: string, root: string, request: FastifyRequest<RouteGenericInterface, RawServerDefault, IncomingMessage, FastifySchema, FastifyTypeProviderDefault, unknown, FastifyBaseLogger, ResolveFastifyRequestType<...>>) => boolean'.
  Overload 3 of 3, '(plugin: FastifyPluginCallback<FastifyStaticOptions, RawServerDefault, FastifyTypeProviderDefault, FastifyBaseLogger> | FastifyPluginAsync<...> | Promise<...> | Promise<...>, opts?: FastifyRegisterOptions<...> | undefined): FastifyInstance<...> & PromiseLike<...>', gave the following error.
    Type '(pathName: string, _root: string, req: Request) => Promise<boolean>' is not assignable to type '(pathName: string, root: string, request: FastifyRequest<RouteGenericInterface, RawServerDefault, IncomingMessage, FastifySchema, FastifyTypeProviderDefault, unknown, FastifyBaseLogger, ResolveFastifyRequestType<...>>) => boolean'.ts(2769)`

server/src/index.ts
Changk820: The signature '(path: PathLike, callback: (exists: boolean) => void): void' of 'fs.exists' is deprecated.ts(6387)

server/src/util.ts
paologum: if you can't do async, it would probably be better to make your own route and then make the method async

server/src/index.ts
paologum: we already have a file util function that is asynchronous so use that
also existsSync only has a synchronous you have to use `realpath` or `access` if you wanted to do your own checking but that's not enough so just use our handler

app/src/settings_tabs/map_points/MapViewer.tsx
```diff
@@ -99,7 +99,7 @@ export const MapViewer: React.FC<MapProps> = observer(({ name, isFacies }) => {
                     id="map"
                     ref={imageRef}
                     style={mapStyle}
-                    src={devSafeUrl(mapData.img)}
+                    src={devSafeUrl(mapData.img.replace("/assets/uploads", "/getMapImages"))}
```
paologum, line 102: just store the file image name here and then instead of reconstructing it just construct it here
there is no reason to store the filepath in `img` if we don't need ut

paologum (changes requested):

server/src/index.ts
```diff
@@ -120,11 +121,50 @@ server.register(fastifyStatic, {
...
+  const [uuidFolder] = pathSegments.slice(1); // Extract UUID folder
+  // Ensure UUID matches and path is valid
+  return uuidFolder === uuid && isValidMapImagePath(pathName);
+};
+
+const assetPaths = ["public", "private"];
+assetPaths.forEach((type) => {
+  server.register(fastifyStatic, {
+    root: path.join(process.cwd(), `assets/uploads/${type}`),
+    prefix: `/getMapImages/${type}`,
```
paologum, line 153: routes shouldn't have get in them because the request already defines that in the type of request (`GET`, `POST`, and `PATCH`)
so just name this `/map-images/${uuid}/${datapack}/${image}`
and use these to service the image

app/src/settings_tabs/map_points/MapViewer.tsx
Changk820: How would I know where the image is stored with only the name of the map?

I don't know how to get the exact datapack an image is in

server/src/routes/routes.ts
```diff
@@ -485,3 +485,59 @@ export const fetchDatapackCoverImage = async function (
...
+ * Path must have 'MapImages' as second to last string in path and end with an image extension */
+const isValidMapImagePath = (pathName: string): boolean => {
+  const pathSegments = pathName.split("/");
+  const isImage = /\.(png|jpg|jpeg|gif|webp)$/i.test(pathName);
+  const isSecondToLastSegment = pathSegments[pathSegments.length - 2] === "MapImages";
+  return isSecondToLastSegment && isImage;
+};
+
+// Utility to validate private access based on UUID
+const isAllowedPrivatePath = ({ pathName, req }: { pathName: string; req: Request }): boolean => {
```
paologum, line 505: you should be able to just use `FastifyRequest` to type out `Request` and don't need to declare a new interface

server/src/routes/routes.ts
```diff
@@ -485,3 +485,59 @@ export const fetchDatapackCoverImage = async function (
...
+  try {
+    const { type } = request.params;
+    const rawPath = request.params["*"] as string;
+    const baseDir = path.join(process.cwd(), "assets", "uploads", type);
+    const fullPath = path.join(baseDir, path.normalize(rawPath));
+    const isVerified = await verifyFilepath(fullPath);
+    if (!isVerified) {
+      return reply.status(403).send({ error: "Forbidden: invalid path" });
+    }
+    const pathName = `/assets/uploads/${type}/${rawPath}`;
```
paologum, line 526: `/assets/uploads` should be changed to `assetconfigs.uploadDirectory`

paologum (changes requested):
just a couple changes!

server/src/parse-map-packs.ts
```diff
@@ -113,7 +113,8 @@ export function processLine(
         );
       }
       map.name = formatColumnName(String(info[1]));
-      map.img = `/${assetconfigs.imagesDirectory}/${String(info[2])}`;
+      const trimmedPath = map_info.replace(/\/map-packs\/.*$/, "");
+      map.img = `/${trimmedPath}/MapImages/${String(info[2])}`;
```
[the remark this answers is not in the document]
Changk820: I can't do:
      // const trimmedPath = path.dirname(path.dirname(map_info));
This is what caused the tests to fail.

It expected
      "img": "/server/__tests__/__data__/parse-map-packs-test-1.txt/MapImages/AfricaBight.png"

Using this commented out code, I give
       "img": "/server/__tests__/MapImages/AfricaBight.png"

I think what happened was the path name got duplicated, so what I was doing was removing that duplicated part, then tacking the "MapImages/AfricaBight.png" back on at the end.

Or I could just change the tests, going back twice in the directory seems to still work, just tests fail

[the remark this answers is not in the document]
paologum: the tests are just intended to be the correct version, if that's what you expect then you should change the expected to be what it should be

PR #277, opened 2024-10-19
Tsc 510 upload size limits 3000 chars

lbostre:
- Pro users or Admins can upload datapacks over 3000 characters
- Regular users cannot


server/src/routes/user-routes.ts
```diff
@@ -296,6 +298,11 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
           reply.status(400).send({ error: `Empty file cannot be uploaded` });
           return;
         }
+        if (uploadedFile.file.bytesRead > 3000 && !isProOrAdmin) {
+          await rm(filepath, { force: true });
+          reply.status(400).send({ error: `Regular users cannot upload datapacks over 3000 characters.` });
+          return;
+        }
```
paologum, line 305: this is bytes read and not characters, and unless i misunderstood professor, he wants characters to be the limitation, not bytes

server/src/util.ts
```diff
@@ -71,6 +71,7 @@ export async function copyDirectory(src: string, destination: string): Promise<v
  * @returns
  */
 export async function grabFilepaths(files: string[], topDirectory: string, botDirectory: string): Promise<string[]> {
+  topDirectory = topDirectory.replace("/", "").replaceAll("\\", "/");
```
paologum, line 74: why do you replace / with ""

i feel like this could go wrong with different inputs 
if you are trying to get rid of leading slashes, i would do it more robustly

server/src/routes/user-routes.ts
```diff
@@ -261,6 +261,8 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
   const fields: Record<string, string> = {};
   let uploadedFile: MultipartFile | undefined;
   const userDir = await getPrivateUserUUIDDirectory(uuid);
+  const user = await findUser({ uuid });
+  const isProOrAdmin = user.length > 0 && (user[0]?.accountType === "pro" || user[0]?.isAdmin);
```
paologum, line 265: you should put this and the `getPrivateUserUUIDDirectory` in a try catch

paologum (changes requested):

server/src/routes/user-routes.ts
```diff
@@ -260,9 +260,17 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+  let userDir: string = "";
+  let user;
+  let isProOrAdmin;
+  try {
+    userDir = await getPrivateUserUUIDDirectory(uuid);
+    user = await findUser({ uuid });
+    isProOrAdmin = user.length > 0 && (user[0]?.accountType === "pro" || user[0]?.isAdmin);
+  } catch (e) {
+    reply.status(401).send({ error: "Could not find private user directory or user with error " + e });
+  }
```
paologum, line 272: small thing, but since `isProOrAdmin` can be defined as a `const`, it would be better so that it's visually easier to know that it shouldn't change

paologum (changes requested):

server/src/routes/user-routes.ts
```diff
@@ -283,6 +283,15 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
   };
+  let userDir: string = "";
+  let user: User[] = [];
+  try {
+    userDir = await getPrivateUserUUIDDirectory(uuid);
+    user = await findUser({ uuid });
+  } catch (e) {
+    reply.status(401).send({ error: "Could not find private user directory or user with error " + e });
+  }
+  const isProOrAdmin = user.length > 0 && (user[0]?.accountType === "pro" || user[0]?.isAdmin);
```
paologum, line 294: ok last thing, i keep realizing things after looking over the code, but if a user doesn't exist so if `user === null` it shouldn't continue so there should be a reply back and a return before `isProOrAdmin` probably.

server/src/routes/user-routes.ts
```diff
@@ -283,7 +283,18 @@ export const uploadDatapack = async function uploadDatapack(request: FastifyRequ
...
+    user = await findUser({ uuid });
+    if (user.length == 0 || !user) {
+      reply.status(401).send({ error: "Could not find user." });
+      return;
+    }
+    userDir = await getPrivateUserUUIDDirectory(uuid);
+  } catch (e) {
+    reply.status(401).send({ error: "Could not find private user directory or user with error " + e });
+  }
   const isProOrAdmin = user.length > 0 && (user[0]?.accountType === "pro" || user[0]?.isAdmin);
```
paologum, line 298: for readability user should not be an array of users, it should just be one user, there's no reason to put the scope as an array of users, then you won't also need `?` on line `:298`

server/assets/config.json
```diff
@@ -13,5 +13,6 @@
   "uploadDirectory": "assets/uploads",
   "publicDirectory": "public",
   "privateDatapacksDirectory": "assets/uploads/private",
-  "publicDatapacksDirectory": "assets/uploads/public"
-}
\ No newline at end of file
+  "publicDatapacksDirectory": "assets/uploads/public",
+  "adminConfigPath": "assets/admin-config.json"
+}
```
paologum, line 18: shouldn't add this

server/assets/config.json
```diff
@@ -13,5 +13,5 @@
   "uploadDirectory": "assets/uploads",
   "publicDirectory": "public",
   "privateDatapacksDirectory": "assets/uploads/private",
-  "publicDatapacksDirectory": "assets/uploads/public"
-}
\ No newline at end of file
+  "publicDatapacksDirectory": "assets/uploads/public",
```
paologum, line 16: remove this `,` i believe it can't read it with this
`yarn dev:config`

paologum (approved):
i will approve this but check your slack, i have a new task that you should prioritize that makes this process less confusing

PR #282, opened 2024-11-04
Upload Error Limit on Frontend for Regular Users 

lbostre:
4eaab863f0ac5f5463acf0a9bb036475
- error for when a regular user tries to upload a datapack over 3000 characters


app/src/state/actions/general-actions.ts
```diff
@@ -265,7 +265,15 @@ export const uploadUserDatapack = action(
         }
         pushSnackbar("Successfully uploaded " + title + " datapack", "success");
       } else {
-        displayServerError(data, ErrorCodes.INVALID_DATAPACK_UPLOAD, ErrorMessages[ErrorCodes.INVALID_DATAPACK_UPLOAD]);
+        if (response.status === 403) {
+          displayServerError(data, ErrorCodes.DATAPACK_TOO_LARGE, ErrorMessages[ErrorCodes.DATAPACK_TOO_LARGE]);
```
paologum, line 269: use `pushError` not `displayServerError` just because you don't actually need to log the server error

app/src/util/error-codes.ts
```diff
@@ -3,6 +3,7 @@ import i18n from "../../i18n";
 export enum ErrorCodes {
   SERVER_RESPONSE_ERROR = "SERVER_RESPONSE_ERROR",
   INVALID_DATAPACK_INFO = "INVALID_DATAPACK_INFO",
+  DATAPACK_TOO_LARGE = "DATAPACK_TOO_LARGE",
```
paologum, line 6: can you change this to be more specific to the error so like `REGULAR_USER_UPLOAD_DATAPACK_TOO_LARGE`
also change the translation key

paologum (changes requested):

paologum (approved):

PR #286, opened 2024-11-12
Tsc 566 add priority to datapacks display

paologum:
# Main

* fixed a bug with migrating datapacks and added it to the `dev` command
* didn't add it to the `prod` command, since migrating can have issues and we should probably not automate it if it has the potential to mess data up since it's a makeshift and original migration workflow
* added priorities to datapacks (required)
* can rearrange in the `ag-grid`, will debounce requests after three seconds
* can edit singular cell's priority
* can reset priorities back to one index (this was my solution since dragging around can infinitely increase numbers
* priority batch updates only for admin, probably thinking that users should just be individual edits on the `DatapackProfile` tab and not similar to the admin priority arrangement with `ag-grid-react`
* added filenames to respective column in `AdminDatapackConfig.tsx`

# Singular Cell Edit and Row Drag and Sorted Display

https://github.com/user-attachments/assets/d4a0643f-9bd6-4483-8c8e-1787c2eff9d1

# More Demo and Reset Priorities

https://github.com/user-attachments/assets/9ae5bc7e-36bc-4ebb-a750-fcf496fb19e3


paologum:
https://github.com/user-attachments/assets/35aefb9b-8cf6-4428-88af-12fc6bbb004f

asivath (approved):

PR #290, opened 2024-11-18
Limit fields backend and tests

lbostre:
- backend limits fields for datapack upload
- tests included


paologum (changes requested):
notes, contact, description, references doesn't have a limit on the frontend and backend

paologum (approved):

PR #298, opened 2024-12-29
Tsc 578 help rework

JacquiFreeElectron:
I took out `Contributor` because I think that's the same as the `About` page. we can move `About` in the help page if Prof. Ogg wants.

For the Quick Start Guide, I've placed a beacon before the guide starts, while for other tours, I've disabled this feature. Let me know which style you prefer, and I'll make sure they're consistent across all sections.

https://github.com/user-attachments/assets/dbe84e38-24b1-49ff-8156-57140497994d


paologum:
i think that the last step should be `finish tour` and not `last`
additionally, i'm going to assume that ogg/aaron would want these not to be hidden in the help tab because these are VERY useful so i would assume there would be a help or tour button of each (datapacks, settings, etc) in each tab that would start the tour, but i think for right now it's ok here. just in the future or on friday i wanna assume they would want that

paologum:
last front end thing, 
image

this should be capitalized and probably named more appropriately to `View file format info`

app/src/App.tsx
```diff
@@ -134,6 +162,51 @@ export default observer(function App() {
...
+            continuous
+            run={state.isSettingsTourOpen}
+            steps={settingsTour}
+            callback={handleSettingsTourCallback}
+            locale={{ skip: "Quit Tour" }}
+            showProgress
+            styles={{
+              options: {
+                zIndex: 200,
+                primaryColor: "#6693C9",
```
paologum, line 205: for `primaryColor` are you able to use the theme?  we would prefer that so that the theme stays consistent with different palettes

app/src/FileFormatInfo.tsx
```diff
@@ -0,0 +1,1189 @@
...
+import { observer } from "mobx-react-lite";
+import patternWidthExample from "./file_format_guide_images/patternWidthExample.png";
+import transect_ex1 from "./file_format_guide_images/transect_ex1.png";
+import transect_load_template_with_error from "./file_format_guide_images/transect_load_template_with_error.png";
+import transect_save_template from "./file_format_guide_images/transect_save_template.png";
+import transect_template_missing_point from "./file_format_guide_images/transect_template_missing_point.png";
+import transect_template_space_merged from "./file_format_guide_images/transect_template_space_merged.png";
+import transect_template_space from "./file_format_guide_images/transect_template_space.png";
+import transect_template_with_error from "./file_format_guide_images/transect_template_with_error.png";
+import transect_template from "./file_format_guide_images/transect_template.png";
```
paologum, line 11: to prevent long build times and the like, we should store these on the server then service them from the app

app/src/Help.tsx
```diff
@@ -28,49 +46,109 @@ export const Help = observer(function Help() {
             <Typography>Quick Start Guide</Typography>
           </AccordionSummary>
           <AccordionDetails>
-            <Typography>Put QSG here</Typography>
+            <Typography style={{ marginBottom: "20px" }}>
+              Welcome to Time Scale Creator Online! In this Quick Start Guide, you will quickly learn about the
+              functions of each tab on the Nav bar. Click the &apos;Start QSG&apos; button to begin.
```
paologum, line 51: tours should also be multilingual so this should be a translation (since they are really important)

app/src/Help.tsx
```diff
@@ -28,49 +46,109 @@ export const Help = observer(function Help() {
...
+            <TSCButton buttonType="primary" onClick={() => runTour("datapacks")}>
+              Start Datapacks Tour
+            </TSCButton>
+            <Divider style={{ marginTop: "20px" }} variant="middle" />
+            <Typography style={{ marginBottom: "20px" }}>
+              In the settings tour you will explore the Settings page in depth. Click the button to start your tour!
+            </Typography>
+            <TSCButton buttonType="primary" onClick={() => runTour("settings")}>
+              Start Settings Tour
+            </TSCButton>
```
paologum, line 75: same

app/src/state/state.ts
```diff
@@ -130,6 +130,10 @@ export type State = {
   presetColors: string[];
   isProcessingDatapacks: boolean;
   unsavedDatapackConfig: DatapackConfigForChartRequest[];
+  isQSGOpen: boolean;
+  isDatapacksTourOpen: boolean;
+  isSettingsTourOpen: boolean;
+  isWorkshopsTourOpen: boolean;
```
paologum, line 136: can you make this all in one object in state like
```
guides: {
...
}

```

app/src/tours.ts
paologum, line 1: this'll be crazy but these will probably need translation (you don't need to do the translation for chinese yourself but just the english ones and connect/put them in the `en.json`

paologum (changes requested):
looks really good i really think this is super beneficial!

app/src/tours.ts
JacquiFreeElectron: np, I can do the translation for Chinese too

JacquiFreeElectron:
Screenshot 2025-01-02 at 5 31 20 PMScreenshot 2025-01-02 at 5 31 29 PM

app/src/FileFormatInfo.tsx
```diff
@@ -0,0 +1,1181 @@
+import React from "react";
+import { observer } from "mobx-react-lite";
+
+const serverURL = `${import.meta.env.VITE_SERVER_URL ?? "http://localhost:3000"}/public/file_format_guide_images`;
```
paologum, line 4: you should use `devSafeUrl` here, also on the pr preview the images do not load

paologum (changes requested):
dark mode should have a secondary background color palette for the guide tooltips

additionally, switching tabs (when you click settings or datapacks tours) do not change the tab you are on. it still says you are on help when on settings tab for instance.

Screenshot 2025-01-03 at 12 35 14 PM

JacquiFreeElectron:
the images load on pr preview now: Screenshot 2025-01-03 at 11 55 05 PM
dark mode: 
Screenshot 2025-01-03 at 11 56 43 PM
Tab bug is fixed now: 
Screenshot 2025-01-03 at 11 57 22 PM

paologum (approved):

PR #302, opened 2025-01-05
Tsc 553 calendar for admin

Ganthry:


paologum:
you should add a demo and pictures of your pr in light mode and dark mode
any description if needed as well

Ganthry:
https://github.com/user-attachments/assets/12c06c62-22a3-4e58-9a0e-3618eaf27b3b

paologum:
the sizing of the different words shouldn't affect the display/configuration of the spacing between the the month and filter, these should be consistent

https://github.com/user-attachments/assets/7221154f-b95a-4057-a371-5d6f9b06a15a

paologum:
also to indicate hoverability, the events should be hoverable 
additionally, workshops are most likely going to be a couple hours in one day, so this calendar should support times using hours and minutes with the associated display

paologum:
additionally, the month button to go left and right should be proportional to the size of the icon. since the icon is smaller, the surrounding button shouldn't be so big. also it shouldn't have a button hover background that is a square, it should probably be circular.

https://github.com/user-attachments/assets/cfcaa77d-e828-4aff-a1b4-36c427428cbe

hovering over the filter should have hover background similar to this button in `MapPoints`



https://github.com/user-attachments/assets/87a10f19-9470-420e-b957-4baf1a1670e0

app/src/Workshops.tsx
```diff
@@ -26,6 +31,10 @@ import {
 import { useNavigate } from "react-router";
 import { useTranslation } from "react-i18next";
 import { NotLoggedIn } from "./NotLoggedIn";
+import { Calendar as BigCalendar, CalendarProps, momentLocalizer, ToolbarProps } from "react-big-calendar";
+import { SelectChangeEvent } from "@mui/material/Select";
+import moment from "moment";
```
paologum, line 36: we already import `dayJs` for this project and format our dates with that. is the calendar only possible with `moment`? if possible with `dayjs` than you should stay with `dayJs`

app/translation/zh.json
```diff
@@ -78,6 +79,22 @@
...
+                    "4": "四月",
+                    "5": "五月",
+                    "6": "六月",
+                    "7": "七月",
+                    "8": "八月",
+                    "9": "九月",
+                    "10": "十月",
+                    "11": "十一月",
+                    "12": "十二月"
+                }
```
paologum, line 97: normally we don't add the chinese translations ourselves but since you already put it in i would check with @JacquiFreeElectron to make sure these are correct translations (mainly `all` because `months` is pretty straightforward)

app/src/Workshops.tsx
paologum: i believe the `react-big-calendar` supports dayJs localizer so you should use that

app/src/Workshops.tsx
```diff
@@ -184,11 +193,85 @@ export const Workshops: React.FC = observer(() => {
...
+        components={{ toolbar: CustomToolbar }}
+        onSelectEvent={(event) => setWorkshopAndNavigate(event as { title: string })}
+        dayPropGetter={(date) => {
+          const newStyle = {
+            backgroundColor: theme.palette.secondaryBackground.main,
+            borderRadius: "0px",
+            border: "none"
+          };
+          if (date.toDateString() === new Date().toDateString()) {
+            newStyle.backgroundColor = "#aecfeb";
```
paologum, line 253: if it's a new color, you should add it to the theme. you should make dark and light mode different for this color because it is a little too bright in dark mode

app/src/Workshops.tsx
```diff
@@ -184,11 +193,85 @@ export const Workshops: React.FC = observer(() => {
   const upcomingWorkshops = getUpcomingWorkshops(dummyWorkshops);
   const pastWorkshops = getPastWorkshops(dummyWorkshops);
 
+  const [calendarWorkshops, setCalendarWorkshops] = useState("All");
```
paologum, line 196: i would call this `calendarWorkshopFilter`

app/src/Workshops.tsx
```diff
@@ -184,11 +193,85 @@ export const Workshops: React.FC = observer(() => {
...
+
   // TODO: change this when implement the backend
-  function setWorkshopAndNavigate(workshop: Workshop) {
-    navigate(getNavigationRouteForWorkshopDetails(workshop.workshopId));
+  function setWorkshopAndNavigate(event: { title: string }) {
+    const workshop = dummyWorkshops.find((w) => w.title === event.title);
```
paologum, line 200: workshops can have the same name, the unique identifier is the workshop id

paologum (changes requested):
just some changes/comments!

Ganthry:
https://github.com/user-attachments/assets/575ebddd-1f87-4d58-bb8c-df442aecdc72

The calendar only starts open when there is an event in the next week. Not sure on the placement of the switch or if it should be somewhere else/look different

paologum:
i would probably put it in the same row as `Registered Workshops` with registered workshops in the middle and the switch at the right (css wise, `justify-content: space-between` with "three" elements, blank, registered, switch)
there might be another way to do this but this seems the simplest for me to explain

paologum:
https://github.com/user-attachments/assets/51ffeda3-cf9b-4597-ba63-157cbc16244b

since one day workshops are the norm, you should have media queries with css to properly adjust the font etc to display the name and/or the time
ideally we want the name to always be visible and not to clip
to do so i would adjust the font size etc
as you can see in the video, there still is space in the text header at the top of the blue
i would give it a smaller font size, the title a line height of 2 and if you can fit the time do so, if not it's ok to clip it.
if the title goes past the line height of 2 use ellipsis to overflow (you can use webkit clamp like other places in the repo if you can't seem to figure it out)

paologum:
image

is this intended?

paologum:
image

this is iphone SE btw, advanced typescript looks like it's off centered?

app/src/Workshops.tsx
```diff
@@ -184,10 +224,231 @@ export const Workshops: React.FC = observer(() => {
...
+        ? `${t(`workshops.calendar.months.${monthIndex}`)} ${year}`
+        : calendarView === "week"
+          ? `${t(`workshops.calendar.months.${monthIndex}`)} ${label.split(" ")[1]} - ${label.split(" ")[3]}`
+          : `${t(`workshops.calendar.days.${label.split(" ")[0].toLowerCase()}`)}, ${t(`workshops.calendar.months.${monthIndex}`)} ${label.split(" ")[2]}`;
+    };
+
+    return (
+      <Box
+        className="calendar-toolbar"
+        sx={{ display: "flex", alignItems: "center", flexDirection: { xs: "column", md: "row" } }}>
```
paologum, line 270: you might as well add the sx to the class (besides the `flexDirection` if you don't want to do the media queries)

app/src/Workshops.tsx
```diff
@@ -184,10 +224,231 @@ export const Workshops: React.FC = observer(() => {
...
+        start: current.isSame(event.start, "day")
+          ? event.start
+          : dayjs(current.format("YYYY-MM-DD") + " " + startTime).toDate(),
+        end: current.isSame(event.end, "day") ? event.end : dayjs(current.format("YYYY-MM-DD") + " " + endTime).toDate()
+      });
+      current = current.add(1, "day").startOf("day");
+    }
+
+    return days;
+  });
```
paologum, line 451: image

this layout is kind of confusing, wouldn't you want the week to fill from 9 am, full day, full day, then on Friday end on 6 pm? for react basics

paologum (commented):
image

image

image


the white on the black contrast is a little to contrasting imo, could you try a lighter gray like this notion calendar

paologum:
image


i am slightly confused, if this is 9am 29th to 6pm 30th, shouldn't it be filling out the 29th from the 9am block forward and filling 30th up until 6 pm?

Ganthry:
> image
> 
> i am slightly confused, if this is 9am 29th to 6pm 30th, shouldn't it be filling out the 29th from the 9am block forward and filling 30th up until 6 pm?

for this I went off of how google calendar does it. 

image


If an event takes more than a day, in week view its displayed as a banner at the top of the view with the times included. If an event was single day it would take up the given time/space of that day. I had the other implementation you're talking about but that led to weird cases like this when multi day events overlap. 

image

paologum:
ah ok i see i would probably go full google calendar and not put the end time as well

paologum:
also we could save some space by putting those on the same line, i'm not 100% sure how integrated having two lines is in your code, but if there is an easy solution to have it same line then wrap I would do that, otherwise if it's really annoying just keep it and let me know that you're keeping it

Ganthry:
> also we could save some space by putting those on the same line, i'm not 100% sure how integrated having two lines is in your code, but if there is an easy solution to have it same line then wrap I would do that, otherwise if it's really annoying just keep it and let me know that you're keeping it
> 
> image

I put them on one row in week view for long events and got rid of the end time in this case

Ganthry:
Turns out it already builds down and expands in this case, does this look good?
image

paologum:
yeah that looks good, but it shouldn't have a min width i don't think that is default taking up empty space

Ganthry:
> yeah that looks good, but it shouldn't have a min width i don't think that is default taking up empty space

image
image

Is this better?

PR #303, opened 2025-01-06
Tsc 585 send basic datapack data on page load and details in background

asivath:
- only send metadata on page load
- send full datapack when the user clicks the datapack
- fetch for datapackprofile and also presets
- i removed some info for the expanded datapack cards since they wouldn't be available
- there were also some other bugs i found that I fixed which were: can't click generate on presets twice in a row, editing the title of a datapack failed to remove the old datapack from state


paologum:
some bugs with clicking off before loading resolves
clicking to add a datapack that hasn't been loaded than clicking off. it should probably assume the user wants it to be added and wait til it loads in. currently from this there seems to be a dismissing of selection if not loaded in (only sometimes)

additionally the datapack selection has some weird bugs i can't really describe well, you should just watch
also deselecting all datapacks doesn't pop up with the prompt ie the second video maybe not your changes but could you change it so it still asks (i think this is intended to happen)

https://github.com/user-attachments/assets/661d7409-9c86-43e4-ac65-1f51861b856c

# Deselecting

https://github.com/user-attachments/assets/80bf2adf-3649-4bdd-9e27-e974f96cf58e

paologum:
https://github.com/user-attachments/assets/80efca1d-bc04-4d8d-9e95-d1057a4ec05a


okay after more experimenting, (i haven't looked at the code yet) i'm pretty sure what is happening is it goes

user clicks -> loads -> clicks away from datapack.tsx while loading background-> still background loading (user free to do anything) -> loads background complete -> pop up (user locked)

which means the user is able to do things while the background loading occurs
we don't want the user to do anything if the config isn't 100% which is a problem

if

user clicks -> loads -> clicks away from datapacks while loading background -> returns -> weird final state that sometimes completes like in the video, config is odd

app/src/settings_tabs/Datapack.tsx
```diff
@@ -202,15 +213,49 @@ const DatapackGroupDisplay: React.FC<DatapackGroupDisplayProps> = observer(
     const isPublicContributed = header === t("settings.datapacks.title.contributed");
     const shouldSplitIntoTwoCol = isOfficial || isPublicContributed;
     const visibleLimit = shouldSplitIntoTwoCol ? 12 : 6;
-    const visibleDatapacks: (Datapack | null)[] = showAll ? datapacks : datapacks.slice(0, visibleLimit);
+    const visibleDatapacks: (DatapackMetadata | null)[] = showAll ? datapacks : datapacks.slice(0, visibleLimit);
     const skeletons = Array.from({ length: extraLoadingSkeletons }, () => null);
-    const onChange = (newDatapack: DatapackConfigForChartRequest) => {
-      if (state.unsavedDatapackConfig.includes(newDatapack)) {
+    const onChange = async (newDatapack: DatapackConfigForChartRequest) => {
+      if (state.unsavedDatapackConfig.some((datapack) => datapack.title === newDatapack.title)) {
```
paologum, line 219: if you just use title doesn't that pose problems for datapacks of the same title but different types? isn't that why `includes` was used

app/src/settings_tabs/Datapack.tsx
```diff
@@ -202,15 +213,49 @@ const DatapackGroupDisplay: React.FC<DatapackGroupDisplayProps> = observer(
...
+              case "workshop": {
+                const metadata = state.datapackMetadata.find(
+                  (d) => d.title === newDatapack.title && d.type === "workshop"
+                );
+                if (!metadata) return;
+                assertWorkshopDatapack(metadata);
+                datapack = await actions.fetchWorkshopDatapack(metadata.uuid, newDatapack.title);
+                break;
+              }
+            }
```
paologum, line 244: because this is so similar to the other function in `DatapackProfile` i think you should abstract this to a method and accept the params for `type` and optional `recaptcha` method

app/src/settings_tabs/Datapack.tsx
```diff
@@ -50,115 +53,123 @@ export const Datapacks = observer(function Datapacks() {
...
             className={styles.buttons}
-            onClick={() => {
-              setFormOpen(!formOpen);
+            onClick={async () => {
+              setLoading(true);
+              while (state.loadingDatapacks) {
+                await new Promise((resolve) => setTimeout(resolve, 100));
+              }
+              setLoading(false);
+              await actions.processDatapackConfig(toJS(state.unsavedDatapackConfig));
```
paologum, line 152: instead of having two loaders, you could just disable the confirm button until all selected datapacks have loaded in

i can see both ways though just a comment

app/src/state/actions/general-actions.ts
```diff
@@ -465,10 +435,10 @@ export const processDatapackConfig = action(
   async (datapacks: DatapackConfigForChartRequest[], settingsPath?: string, force?: boolean) => {
     if (datapacks.length === 0) {
       setEmptyDatapackConfig();
-      return;
+      return false;
```
paologum, line 438: why is this false? if we reset it isn't a bad case

app/src/state/actions/user-actions.ts
```diff
@@ -81,32 +84,35 @@ export const handleDatapackEdit = action(
...
-    return null;
+export const refetchDatapack = action(
+  async (editedDatapack: DatapackUniqueIdentifier, originalDatapack: DatapackUniqueIdentifier) => {
+    const userDatapack = await fetchUserDatapack(editedDatapack.title);
+    if (userDatapack) {
+      removeDatapack(originalDatapack);
+      addDatapack(userDatapack);
+      return userDatapack;
+    } else {
+      return null;
```
paologum, line 112: if this function is only for refetching `user` datapacks, then it should be renamed
otherwise you should switch it so it handles all types of refetches (`server`, `workshop` too)

it'd probably be useful for when i make the `edit` pr for them to have this function since i'll probably refetch

app/translation/en.json
```diff
@@ -101,6 +101,10 @@
                 }
             }
         },
+        "presets": {
+            "BASIC": "BASIC PRESETS",
+            "MAP-PACK": "MAP-PACK PRESETS"
+        },
```
paologum, line 107: this should be removed

server/src/routes/user-auth.ts
```diff
@@ -100,4 +109,12 @@ export const userRoutes = async (fastify: FastifyInstance, _options: RegisterOpt
...
     editDatapackMetadata
   );
+  fastify.get(
+    "/workshop/:workshopUUID/datapack/:datapackTitle",
+    {
+      config: { rateLimit: looseRateLimit },
+      schema: { params: fetchWorkshopDatapackParams }
+    },
+    fetchWorkshopDatapack
+  );
```
paologum, line 119: added to tests, also probably should have tests for this since `user-routes` is a highly tested file

paologum (changes requested):
now that i think about it, do websites lean towards copying the original for the user to edit? or editing the original and when reloading or clearing, refetching?

app/src/DatapackProfile.tsx
```diff
@@ -33,52 +33,89 @@ import {
...
-    const datapack = state.datapacks.find((d) => d.title === id && d.type === query.get("type"));
-    return datapack;
+    if (!datapack) {
+      switch (queryType) {
+        case "user":
+          await loadRecaptcha();
+          return await actions.fetchUserDatapack(id);
+        case "official":
+          return await actions.fetchOfficialDatapack(id);
+        case "workshop": {
```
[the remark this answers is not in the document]
paologum: oops this doesn't work nvm

app/src/Presets.tsx
```diff
@@ -69,18 +71,30 @@ const TSCPresetHighlights = observer(function TSCPresetHighlights({
...
+                          }
+                          setLoading(true);
+                          const fetchedDatapack = await actions.fetchOfficialDatapack(dp.name);
+                          if (!fetchedDatapack) {
+                            actions.pushError(ErrorCodes.UNABLE_TO_FETCH_DATAPACKS);
+                            return;
+                          }
+                          actions.addDatapack(fetchedDatapack);
+                          datapacks.push(_.cloneDeep(fetchedDatapack));
+                        }
```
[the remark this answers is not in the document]
asivath: it's on main removing it does not help

app/src/DatapackProfile.tsx
```diff
@@ -157,14 +194,16 @@ export const DatapackProfile = observer(() => {
         actions.pushError(ErrorCodes.DATAPACK_ALREADY_EXISTS);
         return;
       }
-      const result = await actions.handleDatapackEdit(datapack, state.datapackProfilePage.editableDatapackMetadata);
+      const editedDatapack = await actions.handleDatapackEdit(
+        datapack,
+        state.datapackProfilePage.editableDatapackMetadata
+      );
+      if (editedDatapack) setDatapack(editedDatapack);
+      actions.setDatapackProfilePageEditMode(false);
```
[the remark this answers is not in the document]
asivath: probably not

app/translation/en.json
asivath: are these not missing? they're missing on main

paologum: they aren't supposed to be translated they are going to be editable by ogg in the future

asivath:
i've made it so that if you switch to another tab while a datapack is loading it shows a loader while the datapack is fetching and then the popup for confirming the change occurs. 

idk which is used more, we could ask in the meeting, but I lean to the second one since it keeps the app in sync with the server and is simpler

app/src/Presets.tsx
```diff
@@ -69,18 +68,30 @@ const TSCPresetHighlights = observer(function TSCPresetHighlights({
...
+                            datapacks.push(stateDatapack);
+                            continue;
+                          }
+                          actions.setshowLoadingDatapacksLoader(true);
+                          const fetchedDatapack = await actions.fetchOfficialDatapack(dp.name);
+                          if (!fetchedDatapack) {
+                            actions.pushError(ErrorCodes.UNABLE_TO_FETCH_DATAPACKS);
+                            return;
+                          }
+                          actions.addDatapack(fetchedDatapack);
```
paologum, line 87: won't this push two errors? if it doesn't return than it had to have pushed some sort of error

app/src/settings_tabs/Datapack.tsx
```diff
@@ -50,118 +52,121 @@ export const Datapacks = observer(function Datapacks() {
...
-            onClick={async () => {
-              await actions.processDatapackConfig([]);
-            }}>
-            <DeselectIcon />
-          </IconButton>
-        </CustomTooltip>
-        <div>
-          <Typography className={styles.h}>{t("settings.datapacks.see-info-guidance")}</Typography>
-          <Typography className={styles.dh}>{t("settings.datapacks.add-datapack-guidance")}</Typography>
+    <>
```
paologum, line 55: unless i'm reading this wrong is this change just an unused react fragment?

app/src/settings_tabs/Datapack.tsx
```diff
@@ -205,15 +210,26 @@ const DatapackGroupDisplay: React.FC<DatapackGroupDisplayProps> = observer(
...
       } else {
+        actions.setLoadingDatapacks(true);
+        if (!doesDatapackAlreadyExist(newDatapack, state.datapacks)) {
+          const datapack = await actions.fetchDatapack(newDatapack.type, newDatapack.title);
+          if (!datapack) {
+            actions.setLoadingDatapacks(false);
+            actions.pushError(ErrorCodes.UNABLE_TO_FETCH_DATAPACKS);
+            return;
+          }
+          actions.addDatapack(datapack);
```
paologum, line 229: double error again?

app/src/state/non-action-util.ts
```diff
@@ -9,45 +10,51 @@ import { devSafeUrl } from "../util";
 import dayjs from "dayjs";
 import { Workshop } from "../Workshops";
 
+export function compareExistingDatapacks(a: DatapackUniqueIdentifier, b: DatapackUniqueIdentifier) {
+  return (
+    a.title === b.title && a.type === b.type && (isUserDatapack(a) && isUserDatapack(b) ? a.uuid === b.uuid : true)
+  );
+}
```
paologum, line 17: workshop datapacks have uuids

paologum (changes requested):

app/src/DatapackProfile.tsx
```diff
@@ -33,52 +32,76 @@ import {
...
+      console.error("Error fetching datapack", e);
+      displayServerError(e, ErrorCodes.UNABLE_TO_FETCH_DATAPACKS, ErrorMessages[ErrorCodes.UNABLE_TO_FETCH_DATAPACKS]);
+    });
+  }, [queryType, id, isMetadataInitialized, datapack]);
   useEffect(() => {
     return () => {
       actions.setDatapackProfilePageEditMode(false);
     };
   }, []);
+  if (loading) return <TSCDialogLoader headerText="Fetching Datapack" open={true} />;
```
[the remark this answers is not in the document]
paologum: there's already functionality somewheree for a blank loader as an option on this i think

app/src/DatapackProfile.tsx
```diff
@@ -33,52 +32,92 @@ import {
...
+      if (shouldLoadRecaptcha) {
+        removeRecaptcha();
+      }
+      // if the user navigates away from the page quickly, the recaptcha script will be removed but the fetch will still be in progress causing an error as well as a zombie fetch request so we need to abort the fetch request
+      // a second issue is that due to strict mode the initial fetch will be aborted because React will call the cleanup before fetching is done causing the page to flicker to the 404 page for a split second
+      // the second fetch in the second useEffect will be successful and the page will render correctly
+      // this is not an issue in prod since strict mode is disabled
+      controller.abort();
+    };
+  }, [queryType, id, isMetadataInitialized]);
```
asivath, line 113: what do you think about this? this is also a problem for fetch private official datapacks but I haven't fixed it because idk if this is the solution we want to go with. we could just load recaptcha for the whole app and not worry about mounting/unmounting but than we need a way to track if recaptcha is finished loading and waiting for that which is not a terrible solution but we might ask aaron about loading through the entire app.

app/src/state/actions/general-actions.ts
```diff
@@ -60,37 +62,75 @@ import { cloneDeep } from "lodash";
...
+            datapack = await actions.fetchPublicUserDatapack(title, metadata.uuid, options);
+          } else {
+            datapack = await actions.fetchUserDatapack(title, options);
+          }
+        } else {
+          // if the datapack title is being edited then it won't exist in metadata yet, but it will be the user's datapack
+          datapack = await actions.fetchUserDatapack(title, options);
+        }
+        break;
+      }
```
asivath, line 94: lmk if you can think of a better solution for this as well

app/src/state/actions/general-actions.ts
```diff
@@ -60,13 +61,37 @@ import { cloneDeep } from "lodash";
...
+    switch (type) {
+      case "user":
+        if (useRecaptcha) await loadRecaptcha();
+        datapack = await actions.fetchUserDatapack(title);
+        break;
+      case "official":
+        datapack = await actions.fetchOfficialDatapack(title);
+        break;
+      case "workshop": {
+        if (useRecaptcha) await loadRecaptcha();
```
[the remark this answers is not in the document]
asivath: i've changed my mind on this i don't think actions should be loading the recaptcha in. i think it should be the responsibility of the component that is trying to use these functions

app/src/settings_tabs/Datapack.tsx
```diff
@@ -18,36 +18,42 @@ import ViewCompactIcon from "@mui/icons-material/ViewCompact";
...
 
 export const Datapacks = observer(function Datapacks() {
   const { state, actions } = useContext(context);
   const [formOpen, setFormOpen] = useState(false);
   const { t } = useTranslation();
-
+  const shouldLoadRecaptcha =
+    state.isLoggedIn &&
+    (state.user.isAdmin ||
+      state.datapackMetadata.some((dp) => isWorkshopDatapack(dp) || (isUserDatapack(dp) && !dp.isPublic)));
```
paologum, line 42: if we use it multiple places, we should probably just have a getter/function somewhere in util that returns a true or false on loading recaptcha

app/src/state/actions/general-actions.ts
paologum: solution for fetching?

paologum (changes requested):

app/src/DatapackProfile.tsx
```diff
@@ -33,52 +32,92 @@ import {
...
 export const DatapackProfile = observer(() => {
   const { state, actions } = useContext(context);
   const { id } = useParams();
+  const query = new URLSearchParams(useLocation().search);
+  const queryType = query.get("type");
   const navigate = useNavigate();
   const [tabIndex, setTabIndex] = useState(0);
+  const [datapack, setDatapack] = useState<Datapack | undefined>(
+    state.datapacks.find((d) => d.title === id && d.type === queryType)
+  );
```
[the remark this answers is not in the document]
asivath: yeah I added uuid to this

app/src/state/actions/general-actions.ts
asivath: nvm i changed how this function works

app/src/settings_tabs/Datapack.tsx
asivath: do you mean the boolean itself since the logic is different between the two uses

app/src/DatapackProfile.tsx
asivath: the error gets caught in the fetch action

app/src/settings_tabs/Datapack.tsx
paologum: ok yeah they have two different uses, but this should instead of `isWorkshopDatapack(dp) || (isUserDatapack(dp) && !dp.isPublic` should be `isWorkshopDatapack(dp) || isOwnedByUser(dp)`
the public or not public doesn't really matter if i'm thinking of this correctly

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+        {
+            "type": "chrome",
+            "request": "launch",
+            "skipFiles": ["${workspaceFolder}/.yarn"],
+            "name": "Debug app",
+            "preLaunchTask": "app-dev",
+            "url": "http://localhost:5173",
+            "webRoot": "${workspaceFolder}/app"
+        }
+    ]
```
paologum, line 16: is this slow build time fix

app/src/DatapackProfile.tsx
```diff
@@ -33,52 +31,107 @@ import {
...
+      }
+      // if the user navigates away from the page quickly, the recaptcha script will be removed but the fetch will still be in progress causing an error as well as a zombie fetch request so we need to abort the fetch request
+      // a second issue is that due to strict mode the initial fetch will be aborted because React will call the cleanup before fetching is done causing the page to flicker to the 404 page for a split second
+      // the second fetch in the second useEffect will be successful and the page will render correctly
+      // this is not an issue in prod since strict mode is disabled
+      if (import.meta.env.MODE !== "development") {
+        // if we're in prod, we do want to abort to prevent errors and zombie fetch requests
+        // but if we're in dev, skip aborting so we don’t see that flicker
+        controller.abort();
+      }
```
paologum, line 125: hmm wait this will flicker on prod?

paologum: oh wait because prod doesn't have strict mode nvm i see

.vscode/launch.json
asivath: this is to allow debugging the app in vscode but i'll add that here instead of my other branch

app/src/settings_tabs/Datapack.tsx
asivath: if a user dp is public then it uses the route that doesn't require recaptcha. the reason DatapackProfile needs the `isOwnedByUser` check is because they would be able to edit their own datapack so we do need recaptcha and public doesn't matter. the cases where we need recaptcha are if it's an admin to fetch private official datapacks, if it's a private user datapack, or if it's a workshop datapack. are there more cases i'm not thinking of?

paologum (approved):

PR #304, opened 2025-01-09
Tsc 586 abstract upload datapacks

paologum:
refactoring and adding tests.

* added some tests to `user-handler` and `upload-handler` to add them to coverage
* mainly added refactoring into uploading a datapack so EVERY route that uploads now and in the future can call one function `processAndUploadDatapack` to upload
* the way i abstracted parts of these functions is up for debate but i thought these were the best ideas


server/src/upload-datapack.ts
```diff
@@ -0,0 +1,104 @@
...
+    datapackMetadata
+  };
+};
+export const processAndUploadDatapack = async (uuid: string, parts: AsyncIterableIterator<Multipart>) => {
+  const user = await findUser({ uuid }).catch(() => {
+    return [];
+  });
+  if (!uuid || !user || !user[0]) {
+    return { code: 404, message: "Error finding user" };
+  }
```
asivath, line 56: this won't work for Kevin's treatise datapack thing since there's isn't a treatise user. is he not going to use this code? also the empty uuid check should be before using it in findUser

server/src/upload-handlers.ts
```diff
@@ -335,3 +342,85 @@ export async function fetchDatapackProfilePictureFilepath(uuid: string, datapack
...
+      await rm(tempProfilePictureFilepath, { force: true });
+    }
+    if (filepath) {
+      await rm(filepath, { force: true });
+    }
+  }
+  const user = await findUser({ uuid }).catch(() => []);
+  if (!uuid || !user || !user[0]) {
+    return { code: 404, message: "User not found" };
+  }
```
asivath, line 368: why is the parameter uuid string | undefined when you return an error if it's not defined?

server/src/user/user-handler.ts
```diff
@@ -274,7 +274,9 @@ export async function deleteUserDatapack(uuid: string, datapack: string): Promis
   if (!(await verifyFilepath(datapackPath))) {
     throw new Error("Invalid filepath");
   }
-  await rm(datapackPath, { recursive: true, force: true });
+  await rm(datapackPath, { recursive: true, force: true }).catch((e) => {
+    logger.error(e);
+  });
```
asivath, line 279: shouldn't this rethrow the error so caller functions know of the failure

server/src/upload-datapack.ts
paologum: he'll have to change it because i can't conform to his treatise way without actually merging it so whenever the merge happens the other person will need to make the appropriate changes but they shouldn't be hard

and to be honest it might even be helpful to create a fourth type of datapack called treatise datapacks

server/src/upload-handlers.ts
paologum: so i can throw the error here and make the request in the route cleaner
since we are already checking for findUser and the such here i thought it would make it easier to read instead of checking it here and there (since i don't actually check for `uuid`'s existence i check for `user` existence)

server/src/user/user-handler.ts
paologum: true

server/src/upload-datapack.ts
```diff
@@ -0,0 +1,104 @@
...
+        const result = await grabAndVerifyWorkshopUUID(datapackMetadata);
+        if (result.code !== 200) {
+          return result;
+        }
+      } catch (e) {
+        return { code: 500, message: "Failed to verify workshop UUID" };
+      }
+    }
+    if (await doesDatapackFolderExistInAllUUIDDirectories(uuidDirectoryToDownloadTo, datapackMetadata.title)) {
+      return { code: 500, message: "Datapack with the same title already exists" };
```
asivath, line 84: this should be 409 Conflict

server/src/upload-handlers.ts
```diff
@@ -335,3 +342,85 @@ export async function fetchDatapackProfilePictureFilepath(uuid: string, datapack
...
+        // store it temporarily in the /tmp directory
+        // this is because we can't check if the file should overwrite the existing file until we verify it
+        filepath = join(tmpdir(), storedFileName);
+        if (!checkFileTypeIsDatapack(file)) {
+          await cleanupTempFiles();
+          return { code: 415, message: "Invalid file type for datapack file" };
+        }
+        if (file.file.bytesRead > 3000 && !isProOrAdmin) {
+          await cleanupTempFiles();
+          return { code: 400, message: "File is too large" };
```
asivath, line 386: this should be 413 Payload Too Large

asivath (approved):
i consider those http error codes to be non blocking issues so i'm just going to approve

PR #305, opened 2025-01-10
Tsc 590 add editing for all datapacks

paologum:
# Changes

* added loading variable to track private official datapack loading
* `fetchWorkshops` and `extractDatapackMetadataFromDatapack` functions taken from #303 (Aditya's pr)
* can edit workshops and official datapacks AS ADMIN
* currently can only edit workshops when workshops are currently going on. the pr got big so i'll add functionality to visit or display all workshop datapacks on a future date
* abstracted workshop routes to their own folder/file area to anticipate all users being able to make workshops
* fixed back button on `DatapackProfile.tsx` to go to the correct place instead of straight to `/settings`

# Main Part lol
* made edit handlers general and not specific to users
* abstracted edit handlers to a cloud folder/file area to make easier to read and to group to a more appropriate area
* workshop datapacks cannot be changed from public to private

# Videos

https://github.com/user-attachments/assets/63504b9b-cb65-4263-ac16-6268452b0353


app/src/state/actions/shared-actions.ts
```diff
@@ -0,0 +1,203 @@
...
+    pushError(ErrorCodes.SERVER_RESPONSE_ERROR);
+  } finally {
+    setEditRequestInProgress(false);
+  }
+});
+
+export const replaceProfileImageFile = action(
+  async (datapackUniqueIdentifier: DatapackUniqueIdentifier, file: File) => {
+    try {
+      setEditRequestInProgress(true);
```
asivath, line 168: probably want same logged in check for this function and the one above like in handleDatapackEdit

app/src/state/non-action-util.ts
```diff
@@ -52,7 +59,7 @@ export function getDatapackProfileImageUrl(datapack: Datapack) {
   if (datapack.datapackImage) {
     return devSafeUrl(`/datapack-images/${datapack.title}/${uuid}`);
   } else {
-    return devSafeUrl(`/datapack-images/default/${uuid}`);
+    return devSafeUrl(`/datapack-images//${uuid}`);
```
asivath, line 62: double /

server/src/cloud/edit-handler.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+  if ("datapackImage" in newDatapack) {
+    const sourceFilepath = await getTemporaryFilepath(uuid, newDatapack.datapackImage!);
+    // check to see if a temp file was uploaded
+    if (await verifyFilepath(sourceFilepath)) {
+      await changeProfilePicture(
+        uuid,
+        oldDatapackTitle,
+        await getTemporaryFilepath(uuid, newDatapack.datapackImage!)
```
asivath, line 64: this is sourceFilepath

server/src/cloud/general-cloud-requests.ts
```diff
@@ -0,0 +1,41 @@
...
+): Promise<OperationResult> {
+  const response = await processEditDatapackRequest(parts, uuid).catch(() => {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  });
+  if (!response) {
+    return { code: 500, message: "Failed to process request" };
+  }
+  if (isOperationResult(response)) {
+    return response;
+  }
```
asivath, line 23: it's fine to not execute the finally block in this case?

shared/src/index.ts
```diff
@@ -611,6 +611,14 @@ export type DatapackPriorityUpdateSuccess = {
 
 export type DefaultChronostrat = "USGS" | "UNESCO";
 
+export function isOfficialUUID(uuid: string): boolean {
+  return uuid === "official";
+}
+export function isWorkshopUUID(uuid: string): boolean {
+  const workshopPrefix = "workshop-";
+  return uuid.startsWith(workshopPrefix) && !isNaN(Number(uuid.slice(workshopPrefix.length)));
+}
```
asivath, line 620: should handle if it's just `workshop-` since Number would convert the empty string to 0. if you want to be really robust this should also handle decimals and negative numbers. also getWorkshopIdFromUUID should use this function to validate instead of having two different ways of parsing/validation

server/src/user/user-handler.ts
```diff
@@ -511,3 +437,14 @@ export function convertNonStringFieldsToCorrectTypesInDatapackMetadataRequest(fi
...
+export function verifyWorkshopValidity(workshopUUID: string, userId: number): OperationResult {
+  const workshopId = getWorkshopIdFromUUID(workshopUUID);
+  if (!workshopId) {
+    return { code: 400, message: "Invalid workshop UUID" };
+  }
+  if (!isUserInWorkshopAndWorkshopIsActive(workshopId, userId)) {
+    return { code: 401, message: "User does not have access to this workshop" };
+  }
+  return { code: 200, message: "Success" };
+}
```
asivath, line 450: this feels like it belongs in workshop-util.ts. also should add a jsdoc since it's not super clear what it does from the name

asivath (commented):
i think some of the names of files/folders are confusing. 
- what is shared-actions.ts for? from that title I think it's for actions that are shared across the app, but other action files have actions that are shared across the app. what makes these actions specifically "shared"?
- what makes something a cloud request that goes in general-cloud-requests.ts? how is it different from any other kind of request?
- why is edit-handler.ts in the cloud folder? what files would go in the cloud folder? 

also you should move workshop-util.ts to the workshop folder

paologum:
so we have admin and user actions and `shared-actions.ts` is for both admins and users
i can rename it to `shared-admin-user.actions` but with the idea that shared is between server and app, then i thought it wouldn't be confusing calling it `shared-actions` because that is the only thing that gets shared but i can see what you mean

everything cloud related goes in the `cloud`folder. i want all file operations that are helper functions in there. it generalizes operations made my different users/admins because there aren't general areas for "all user" file operations and sometimes we need multiple different files. it's all going to be moved inevitably here so `upload-datapack` and `upload-handler` will also move here but this will change more and it's very small so i just focused on the big changes here. having `edit-handler` and all these floater handlers just floating around seems odd so i chose this. i can move the others in this pr if you want

a cloud request is just a file operation or file change that specifically groups multiple handlers together
for instance if we have multiple handlers in the same file and the grouping function that joins all the handler functions together ( which happens a lot if we abstract and the handlers are located in the same file ) makes testing harder because you have to physically mock them (you cant mock normally unless you "hack" it by self importing which i didn't like since they exist in the same file) i deemed that abstracting into handlers with the specific cloud function then grouping those handlers into a grouping function. the reason we don't just use the grouping in the route itself, is because we have multiple routes using the same functions.

app/src/state/non-action-util.ts
paologum: you need it to be empty so that it sends the default
titles can't be empty so having a default name as the default means that datapacks with the default name can't get their image

server/src/cloud/general-cloud-requests.ts
paologum: yes because we don't have access to the temp files generated by response if it returns an `OperationResult`

shared/src/index.ts
paologum: well if it is negative it wouldn't pass, and we shouldn't allow uuids with negatives right?
i'll change this to handle decimals and the empty string at least for now

server/src/workshop/workshop-util.ts
```diff
@@ -1,13 +1,16 @@
+import { isWorkshopUUID } from "@tsconline/shared";
+import { OperationResult } from "../types";
+import { isUserInWorkshopAndWorkshopIsActive } from "../database";
```
asivath, line 3: needs to end in .js, causing preview to fail

asivath (commented):
i like `shared-admin-user.actions` thats way more clear. i still think cloud is confusing though since nothing is inherently "cloud" and cloud is kind of a generic name. if you want it to be for file operations and handlers what about `file-handlers`.

asivath (approved):

PR #307, opened 2025-01-12
Tsc 450 dual column comparison frontend

jay-jhk-lee:
https://github.com/user-attachments/assets/5aad29ba-89cf-4fef-a1f9-03ae6a1c1bb0


app/src/settings_tabs/advanced_settings/OverlaySettings.tsx
```diff
@@ -0,0 +1,360 @@
...
+    <StyledScrollbar>
+      <Box className="data-mining-settings-container">
+        <Typography className="advanced-settings-header" variant="h6">
+          {t("settings.column.overlay-menu.title")}
+        </Typography>
+        <CustomDivider className="settings-header-divider" />
+        <div className="data-mining-settings-content">
+          <Box className="overlay-settings-container">
+            <Box className="overlay-name-and-checkbox">
+              <Typography>Current Overlay</Typography>
```
paologum, line 80: i would still add the translations just in case

app/src/util/util.ts
```diff
@@ -1,6 +1,13 @@
...
 import Color from "color";
 
+export function discardTscPrefix(name: string | null) {
+  if (!name) {
+    return "";
+  }
+  return name.split(":")[1];
+}
+
 /**
```
paologum, line 11: can you write a comment on what this is accomplishing

app/src/state/actions/util-actions.ts
```diff
@@ -84,3 +84,41 @@ export function checkDatapackValidity(file: File) {
...
+    return false;
+  }
+  let reachedFirstRef = false;
+  while (!reachedFirstRef) {
+    if (
+      !(
+        (dccColumn.minAge <= userTopAge && dccColumn.maxAge >= userBaseAge) ||
+        (dccColumn.minAge > userTopAge && dccColumn.minAge < userBaseAge) ||
+        (dccColumn.maxAge < userBaseAge && dccColumn.maxAge > userTopAge)
+      )
```
paologum, line 99: `if (dccColumn.maxAge < userTopAge || dccColumn.minAge > userBaseAge)`
is easier to understand right?

app/src/state/actions/column-actions.ts
```diff
@@ -705,7 +731,7 @@ export const removeDualColCompColumn = action((column: ColumnInfo) => {
     console.log("WARNING: tried to get", column.parent, "in state.settingsTabs.columnHashMap, but is undefined");
     return;
   }
-  const columnToRemove = prependDualColCompColumnName(column.name);
+  const columnToRemove = dualColCompPrefix + column.name;
```
paologum, line 734: this should still be a function, it's more robust . you can keeep the constant but it should prepend in the function

app/src/settings_tabs/advanced_settings/OverlaySettings.tsx
```diff
@@ -0,0 +1,360 @@
...
+                <Typography className="empty-overlay-name">Choose Column {">>>"}</Typography>
+              )}
+
+              <Box className="overlay-checkbox-container">
+                <TSCCheckbox
+                  checked={state.settingsTabs.columnHashMap.get(dualColCompPrefix + column.name) !== undefined}
+                  className="overlay-checkbox"
+                  onClick={() => {
+                    if (!state.settingsTabs.columnHashMap.get(dualColCompPrefix + column.name)) {
+                      actions.addDualColCompColumn(column);
```
paologum, line 95: prepend here

app/src/settings_tabs/advanced_settings/OverlaySettings.tsx
```diff
@@ -0,0 +1,360 @@
...
+                className={`hide-scrollbar dcc-accordion-wrapper ${state.settingsTabs.columnSearchTerm ? "filtered-border" : ""}`}
+                position="relative">
+                <div className="column-filter-buttons">
+                  <CustomTooltip title="Expand All" placement="top">
+                    <IconButton
+                      disableRipple
+                      className="expand-collapse-column-buttons"
+                      onClick={() => {
+                        if (!state.settingsTabs.columns) return;
+                        actions.setExpansionOfAllChildren(state.settingsTabs.columns, true);
```
paologum, line 121: this affects the main columns

app/src/settings_tabs/advanced_settings/OverlaySettings.tsx
```diff
@@ -0,0 +1,360 @@
...
+          }
+          if (refColumn.columnDisplayType === "Point") {
+            assertPointSettings(refColumn.columnSpecificSettings);
+          } else if (refColumn.columnDisplayType === "Event") {
+            assertEventSettings(refColumn.columnSpecificSettings);
+          } else {
+            return;
+          }
+          refColumn.columnSpecificSettings.drawDualColCompColumn =
+            `class datastore.${column.columnDisplayType}Column:` + column.name;
```
paologum, line 273: does `drawDualColCompColumn` take this full name rather than the original column name?

paologum (changes requested):
one thing, it should probably prune columns not in use and not disable them

paologum:
you should probably add a flex wrap or something for mobile as well as a quick fix

image

app/src/settings_tabs/advanced_settings/OverlaySettings.tsx
jay-jhk-lee: Yeah, that's what the jar file needs to find the column. We could set it as the column name instead and prepend during the parse setting process if that seems better

jay-jhk-lee: I just remembered, I did it this way because I think we didn't want to access state during parse settings. If we only had the name, we would have to access state to find its display column.

paologum: hmm if this is the case then you should make a `util` method to construct it like you did before and use it for constructing the columns as well. (`translateColumnInfoToColumnInfoTSC`'s method)
so we can be more robust from here on

paologum:
Screenshot 2025-01-15 at 9 19 48 PM

it's probably fine making it flat, but i want to think that aaron and ogg will still want the tree hierarchy
we'll know by friday at least but i think it's ok
it just becomes hard for people to find the column they are looking for if they don't know the name and potentially only know the parents

paologum:
Screenshot 2025-01-15 at 9 23 09 PM

since the search bar takes up the full width, the columns should default take the full width as well

paologum:
Screenshot 2025-01-15 at 9 23 56 PM

maybe a distinct background around the selected overlaycolumn to easily distinguish this is important

paologum:
chart (2)

Screenshot 2025-01-15 at 9 28 46 PM

This chart doesn't seem to have the overlay, i turned off the gray off option so it should display all columns regardless of whether or not they have column content in the time interval

jay-jhk-lee:
> This chart doesn't seem to have the overlay, i turned off the gray off option so it should display all columns regardless of whether or not they have column content in the time interval

It works for me, it might be because you clicked "display overlay" when the reference column (Plio...) was turned off. The overlay copies the settings of the reference column. We could default it to on

> additionally, if at least one of the columns has content in the interval doesn't that mean it has data in the time interval

You're right, and from what I've tested the jar program uses the time interval of the reference column for the overlay column. So it doesn't display the column if the time interval is set to outside of the time interval of the reference column, even though an overlay could have data there. To be safe, I made it to check for all the intervals.

paologum:
ok then yeah weshould default to on and probably discuss overlay interval displays in the meeting tomorrow

paologum:
Screenshot 2025-02-12 at 6 11 54 PM

overlay doesn't show up on this column

jay-jhk-lee:
it shows up for me ㅠㅠ

PR #308, opened 2025-01-19
added basic json to excel

jay-jhk-lee:
image


server/package.json
```diff
@@ -21,6 +21,7 @@
     "dotenv": "^16.4.5",
     "fastify": "^4.27.0",
     "fastify-metrics": "^12.1.0",
+    "fs": "^0.0.1-security",
```
paologum, line 24: don't add fs

server/package.json
```diff
@@ -35,6 +36,7 @@
     "nodemon": "^2.0.22",
     "p-map": "^6.0.0",
     "p-queue": "^8.0.1",
+    "path": "^0.12.7",
```
paologum, line 39: don't add path

app/translation/json-to-excel.ts
```diff
@@ -0,0 +1,63 @@
...
+  ...translations
+}));
+
+// Write to Excel
+const worksheet = XLSX.utils.json_to_sheet(rows);
+const workbook = XLSX.utils.book_new();
+XLSX.utils.book_append_sheet(workbook, worksheet, "Translations");
+XLSX.writeFile(workbook, outputExcelFile);
+
+console.log(`Excel file generated at ${outputExcelFile}`);
```
paologum, line 63: minor thing, can you make this `chalk.green`

server/assets/translations/json-to-excel.ts
```diff
@@ -0,0 +1,64 @@
+import * as fs from "fs";
+import * as XLSX from "xlsx";
+import * as path from "path";
+
+// File paths
+const inputDir = "./translation";
+const outputExcelFile = "./translation/translation.xlsx";
```
paologum, line 7: you probably should use join and not hard code paths

app/translation/json-to-excel.ts
jay-jhk-lee: I can change the color, do you want it like how the translation file is rn where the filled translations are green and the empty ones are red

app/translation/json-to-excel.ts
paologum, line 1: remove all app

app/translation/translation.xlsx
paologum, line 1: remove

server/package.json
paologum, line 1: if you're adding everything in this pr shouldn't you add the translation commands here? probably into the dev startup

server/src/translations/json-to-excel.ts
```diff
@@ -0,0 +1,75 @@
+import * as fs from "fs";
+import * as XLSX from "xlsx";
+import path from "path";
+import { assetconfigs, loadAssetConfigs } from "../util.js";
+
+type NestedTranslations = {
+  [key: string]: string | NestedTranslations;
+};
```
paologum, line 8: minor thing you define the same type twice, u can use one so they stay synced

server/src/translations/excel-to-json.ts
```diff
@@ -0,0 +1,74 @@
...
+          current = current[key] as NestedTranslations;
+        }
+      });
+    });
+  });
+
+  // Temporary, Save each language JSON to a file
+  languages.forEach((lang) => {
+    const fileName = `${lang}.json`;
+    fs.writeFileSync("./src/translations/" + fileName, JSON.stringify(languageJsons[lang], null, 2));
```
paologum, line 61: use `join`

paologum: also my main comment of aggregating the `.json` into one file would be needed.
future task to make sure any commit with a changed json has a changed excel in actions would be your next task

paologum (changes requested):
the generated translations should probably not be stored in the server, they should be stored in `shared` and the `i18` translation configuration should point to them

they should also be committed and the previous ones taken out

last thing, lets say you have unique changes to excel and unique changes to the json
how do you preserve both. running one will overwrite the other and vice versa

additionally, we can have a unqiue command for this in `package.json` but it might be better to run these in `index.src` on server startup in addition to that

you do have synchronous filesystem methods, but because it runs on startup it might be fine, i asked aditya to verify

jay-jhk-lee:
The script now merges the two before writing, which resolves the issue you've mentioned. Something we have to think about is what we should do if someone removes a translation

server/src/sync-translations.ts
```diff
@@ -0,0 +1,133 @@
...
+          if (!current[key] || typeof current[key] !== "object") {
+            current[key] = {} as NestedTranslations;
+          }
+          current = current[key] as NestedTranslations;
+        }
+      });
+    });
+  });
+  return languageJsons;
+}
```
paologum, line 61: so this way, changes to the json get overwritten by the excel file? lets say i want to change a current translation in the json. i can't change it in the json, i have to change it in the excel file and rerun the `yarn dev` or `yarn sync-translation` command to see my translation?

just want to make sure i understand this correctly

app/translation/json-to-excel.ts
paologum: if that's easy then yes, that'd help the translators i think

server/src/sync-translations.ts
jay-jhk-lee: Yes, the script prioritizes the excel translation. Would we want it to prioritize the JSON?

server/package.json
```diff
@@ -20,6 +20,7 @@
     "dompurify": "^3.2.4",
     "dotenv": "^16.4.5",
     "fastify": "^4.27.0",
+    "fastify-metrics": "^12.1.0",
```
paologum, line 23: remove this

paologum:
since you already have to push for the change above, can you add the code for making the `available-language.json` with the sync

paologum (approved):

PR #310, opened 2025-01-20
Tsc 569 semver for database migrations

lbostre:
-  semantic versioning for database migrations
- ability to increment major, minor, or patch versions
- upgrading major version requires force flag ("yarn tsx migrate.ts up --force")
- downgrading a major version will throw error and abort


paologum (commented):
overall looks really good, just some comments

server/db/migrate.ts
```diff
@@ -19,32 +20,67 @@ Note: Migrations in the migrations folder are executed in the order of their fil
...
+  if (nextMigration) {
+    const nextMajorVersion = getMajorVersion(nextMigration); // Extract the major version of the next migration
+    if (nextMajorVersion > lastMajorVersion && !force) {
+      logger.error(`Major version change detected: ${lastMajorVersion} -> ${nextMajorVersion}`);
+      logger.error("Please review the migration and apply it manually with flag --force.");
+      process.exit(1); // Exit the process to enforce manual migration
+    }
+  }
+
   const { error, results } = await migrator.migrateToLatest();
```
asivath, line 84: so it looks like what this is doing is getting the first migration that hasn't been applied and then doing the version checking. here's an example of when I think this will go wrong:

- db is currently on 2.8.6
- there are two migrations to be applied 2.8.7 and 3.0.0
- nextMigration finds 2.8.7 and then it passes all the major checks
- migrator.migrateToLatest() will run all migrations, including 3.0.0

the code needs to check all migrations for version compatability

server/db/migrate.ts
```diff
@@ -19,32 +20,67 @@ Note: Migrations in the migrations folder are executed in the order of their fil
...
+async function migrateToLatest(force: boolean = false) {
   const db = new Kysely<Database>({
     dialect: new SqliteDialect({
       database: new BetterSqlite3("TSC.db")
     })
   });
 
+  const migrationFolder = new URL("migrations", import.meta.url).pathname;
+
+  const migrations = await fs.readdir(migrationFolder);
```
asivath, line 49: will readdir get the migrations in the correcct order? your should probably sort it

server/db/migrate.ts
```diff
@@ -19,46 +20,93 @@ Note: Migrations in the migrations folder are executed in the order of their fil
...
+    await db.destroy();
+    return;
+  }
+
+  for (const migration of migrationsToApply) {
+    const { error } = await migrator.migrateUp(); // Migrate one step up
+    if (error) {
+      console.error(`failed to execute migration "${migration}"`);
+      console.error(error);
+      process.exit(1);
```
paologum, line 106: to safely disconnect the database on exits and/or errors for this and the rollback changes, we should probably just (after declaration of the db) wrap the reset of the function in a try and destroy the db in the finally

paologum (changes requested):

paologum (approved):

PR #314, opened 2025-01-24
Tsc 618 timeline changes

paologum:
* ability to create rudimentary cross plots on dev only
* checkbox in preferences is TEMPORARY and only on dev
* timeline capabilities
* split cross plot into timeline support, and backend route and frontend capability

# Video

https://github.com/user-attachments/assets/20b8cd9a-5737-4d19-b7cc-7d9306da3799


asivath (approved):
looks good if you could just wait a bit to merge so I can test the jar change that would be nice

PR #325, opened 2025-01-31
workflow, prometheus compose

Ganthry:
This is more to migrate what I have locally to the server and see if the workflow is correct. If (when) there are problems with this config on the server I'll make the changes there. then I'll add my changes to the repo once it works so they're synced up. Monitoring is gonna be in aaron's folder.


paologum:
to test the workflows you should fork the repository and execute the actions to verify the flows actually work

do you need the email still?

Ganthry:
> do you need the email still?

Yes. Also generating a google app password requires two step verification to be turned on. Do you know if its turned on already? If not I can add it

.github/workflows/deploy-dev.yml
```diff
@@ -275,6 +291,90 @@ jobs:
...
+                echo "Error: Grafana credentials are missing in .env file."
+                exit 1
+              fi
+
+              echo "Updating prometheus.yml with Grafana credentials..."
+              sed -i s/__GRAFANA_USERID__/$GRAFANA_USERID/g $MONITORING_DIR/prometheus.yml
+              sed -i s/__GRAFANA_PSWD__/$GRAFANA_PSWD/g $MONITORING_DIR/prometheus.yml
+
+              docker-compose -f $MONITORING_DIR/docker-compose-prometheus.yml down
+              docker-compose -f $MONITORING_DIR/docker-compose-prometheus.yml up -d
```
paologum, line 328: this assumes the docker is up, probably should have catching if it isn't
also why not just restart?

.github/workflows/deploy-dev.yml
```diff
@@ -275,6 +291,90 @@ jobs:
...
+            --header "Authorization: Bearer $ACCESS_TOKEN" \
+            --header "Content-Type: application/json" \
+            --data "$LIST_PAYLOAD")
+          echo "List Folder Response: $LIST_RESPONSE"
+          FILES_TO_DELETE=$(echo "$LIST_RESPONSE" | jq -r '.entries[] | select(.name | test("\\.jar$"; "i")) | .name' \
+            | grep -v -e "${{ steps.parse_decryption_jar.outputs.name }}" -e "${{ steps.check_active_jar.outputs.new_active_jar }}" -e "${{ steps.check_active_jar.outputs.old_active_jar }}" || true)
+
+          if [[ -z "$FILES_TO_DELETE" ]]; then
+            echo "No files to delete. Exiting."
+            exit 0
```
paologum, line 364: dupe code

server/package.json
```diff
@@ -37,6 +37,7 @@
     "p-queue": "^8.0.1",
     "path-scurry": "^1.10.1",
     "pino-pretty": "^10.0.0",
+    "prom-client": "^15.1.3",
```
paologum, line 40: prom-client is used by `fastify-metrics` under the hood, is there a reason why we need to define it in our package

server/src/index.ts
```diff
@@ -49,6 +49,14 @@ server.register(fastifyMetrics, {
   routeMetrics: { enabled: true }
 });
 
+server.addHook("preHandler", (request, reply, done) => {
+  if (request.raw.url === "/metrics" && request.ip !== "127.0.0.1" && request.ip !== "::1") {
+    reply.code(403).send({ error: "Forbidden" });
+  } else {
+    done();
+  }
+});
```
paologum, line 58: when i look at the `fastify-metrics` library, it has them ordinarily register the `/metrics` with the `metrics plugin` itself. i do understand this controlling it so that internal requests are only allowed, but is the actual plugin ever integrated?

paologum (changes requested):
can you include a link or picture of the metrics board you have made

.github/workflows/deploy-dev.yml
Ganthry: added the check. The restart doesn't reload the prometheus.yml config file so changes would never take.  down/up forces it to

.github/workflows/deploy-dev.yml
Ganthry: woops

server/package.json
Ganthry: fastify-metrics help integrate some metrics easier with fastify but I've been playing around with prom-client while building out the dashboard because it doesn't quite cover everything i want it to cover. The current board just represents server information not actual tsc monitoring which I'm trying to integrate. I might end up replacing fastify-metrics with prom-client fully.

image

(not sure why SWAP isn't connecting working on that). I'm building out the actual dashboard now. Its a lot of fighting with grafana.

paologum: so is `prom-client` needed in this pr, or is it something you're experimenting with.
if it's the latter than you should remove it, if it's before can you show me what it is we're using it for?

paologum: the monitoring for the server looks good, is the integration with tsc monitoring something on the server, grafana, or in actions

.github/workflows/deploy-dev.yml
```diff
@@ -275,6 +291,60 @@ jobs:
...
+              if [ -z "$MONITORING_DIR" ]; then
+                echo "Error: MONITORING_DIR not set in secrets."
+                exit 1
+              fi
+
+              mkdir -p $MONITORING_DIR
+
+              find $MONITORING_DIR/ -type f ! -name ".env" | xargs -r rm -f
+
+              cp monitoring/* $MONITORING_DIR/
```
paologum, line 308: small question, is there a reason we have to have it in a different directory?

Ganthry:
So I ended up getting rid of fastify-metrics. Prom client had everything we needed with more customizability. There are three main things in this PR. The first is prometheus configuration / prometheus workflow. The second is node exporter (which is also an image like prometheus and already on the server b/c of workflow testing) this is what opens all the server metrics shown on the previous dashboard. The second is the "custom metrics" in the index file. The main metrics added are route path count, route duration, and active users (the way i do that last one probably isn't the best but that might be a friday conversation. The board with these metrics looks like this.

image
image

this is all super easy to change based on what would be most helpful to highlight for tsconline and would be all through grafana, I'm not as familiar with what the most "important" routes correspond to what actions (chart generating, authentication, etc...) or which routes we would want to monitor/would be the biggest bottleneck so the current dashboard just lists them all.

I definitely think there are more metrics to add to this / optimizations. Also the restart number is crazy cause I had it running in watch mode and it counted every time I typed anything

.github/workflows/deploy-dev.yml
Ganthry: tbh, someone, either you or aditya, told me to put it there at some point, I just didn't ask questions and assumed there was a reason. If it coulda stayed that'd be funny.

.github/workflows/deploy-dev.yml
paologum: you should add `--remove-orphans` for robustness

server/src/index.ts
```diff
@@ -43,10 +43,63 @@ const server = fastify({
...
+
+// Restrict access to /metrics to localhost only
+server.addHook("preHandler", (request, reply, done) => {
+  if (request.raw.url === "/metrics" && request.ip !== "127.0.0.1" && request.ip !== "::1") {
+    reply.code(403).send({ error: "Forbidden" });
+  } else {
+    done();
+  }
```
paologum, line 102: do we need a prehandler if the actual request handler is right there? we can just add it in the handler

Ganthry: changed it

server/src/index.ts
```diff
@@ -44,10 +44,60 @@ const server = fastify({
...
+    setTimeout(
+      () => {
+        ipSet.delete(ip);
+        activeUsers.set(ipSet.size);
+      },
+      15 * 60 * 1000
+    );
+  }
+  activeUsers.set(ipSet.size);
+});
```
paologum, line 79: can you explain this for me, i'm not sure i understand

Ganthry: This is to check the # of active users. When a request comes in, if its ip isn't already in ipSet, its added. When an ip is added, it gets a timeout that will delete it in 15 minutes. This whole thing is meant to see when a unique ip comes in and if it has been "active" on the site in the past 15 minutes. If not then it does not count as an active user. I treat someone that hasn't interacted on the website for 15 minutes as a non active user. Should i use a different approach for what an active user is?

The last approach i had was to increase the active user count whenever someone accesses the "login" path and decrease active user whenever someone accesses the "logout" path. But if someone doesn't logout it gets kind of janky. Also if someone logged in, the server restarts, and then logs out, I got "-1" users thats why i moved away from that approach.

paologum (approved):

PR #327, opened 2025-02-03
Tsc 605 pdf files made attachable to datapacks

lbostre:
- added ability for user to upload PDFs when creating new datapacks
pdfUploadDesktop
pdfUploadMobile
ae95b8aa35646e75de8b3214a4212622
2bd6b48282936b2d0837aa991d1704e8
07123b717834c309662ad500354a1dcc

DEMO: https://github.com/user-attachments/assets/ccd30090-4c20-4102-8648-76b9197eb640


paologum:
Screenshot 2025-02-06 at 10 04 06 PM

Screenshot 2025-02-06 at 10 04 35 PM

on this width i can't view the pdf files without scrolling
this width can't see the file labels

i don't think the upload button had code for styling it as it decreased so it was always in view 

it would just wrap earlier similar to how you have it 600px

paologum:
i get an error on uploading a datapack on the preview with a pdf attached, when removed, it works

app/src/components/datapack_display/datapack-upload-form-hook.ts
```diff
@@ -113,6 +114,25 @@ const useDatapackUploadForm = (props: DatapackUploadFormProps) => {
...
+    setPDFFiles(files);
+  };
+  const handlePDFFileDelete = (fileName: string) => {
+    if (!pdfFiles) {
+      return;
+    }
+    const filesArray = Array.from(pdfFiles);
+    const updatedFiles = filesArray.filter((file) => file.name !== fileName);
+    const updatedFileList = new DataTransfer();
+    updatedFiles.forEach((file) => updatedFileList.items.add(file));
```
paologum, line 132: there's no way to remove a file from a `DataTransfer` and retain the variable?

server/src/upload-handlers.ts
```diff
@@ -274,6 +278,25 @@ export async function setupNewDatapackDirectoryInUUIDDirectory(
       await rm(datapackImageFilepath, { force: true });
     }
   }
+  if (pdfFields && pdfFields.tempPDFFilePaths && pdfFields.pdfFileNames) {
+    for (let i = 0; i < pdfFields.tempPDFFilePaths.length; i++) {
+      const pdfFilePath = pdfFields.tempPDFFilePaths[i];
+      const pdfFileName = pdfFields.pdfFileNames[i];
```
paologum, line 284: i feel like it's a lot better to have it a `Record< string, string >` where the file name leads to it's path and you just loop through the keys with `Object.keys`

paologum: it also feels weird that you use a `Record<string, string[]>` like a variable storage rather than a dynamic variable key to value

server/src/upload-handlers.ts
```diff
@@ -346,12 +369,16 @@ export async function fetchDatapackProfilePictureFilepath(uuid: string, datapack
 export async function processMultipartPartsForDatapackUpload(
   uuid: string | undefined,
   parts: AsyncIterableIterator<Multipart>
-): Promise<{ fields: { [key: string]: string }; file: MultipartFile } | OperationResult> {
+): Promise<
+  { fields: { [key: string]: string }; file: MultipartFile; pdfFields: { [key: string]: string[] } } | OperationResult
```
paologum, line 373: like the above comment, i want to be able to more or less know how you are planning to use the `pdfFields` just by looking at its type and the name
`pdfFields` seem like an extension of `fields` or related in someway which isn't really the case imo
i think that `pdfFiles: { [fileName: string]: string }}` would better communicate that since it's slightly confusing

paologum (changes requested):
i don't believe i see code for the admin datapack upload form

it simply uses a different function so i think it should be the same

it's in `admin-actions.ts`

# Bug

i can't see multiple pdf files and it fails (like my other comment before said)

https://github.com/user-attachments/assets/afa4ab28-fd2a-4c9b-b2cc-3fe56c8c7a66

app/src/components/datapack_display/datapack-upload-form-hook.ts
lbostre: I think DataTransfer's are read only. I could def be wrong tho

paologum:
image


i'm assuming it can't store the pdfs individually if they have the same name, so it should remove the pdf it already exists when uploading (frontend)

server/src/upload-handlers.ts
```diff
@@ -274,6 +278,19 @@ export async function setupNewDatapackDirectoryInUUIDDirectory(
...
+      const datapackPDFFilepathDest = path.join(datapackFolder, pdfFileName);
+      if (!datapackPDFFilepathDest.startsWith(datapackFolder)) {
+        throw new Error("Invalid datapack PDF filepath destination path");
+      }
+      // remove the original file if it was copied from a temp file
+      if (!manual && pdfFilePath !== datapackPDFFilepathDest) {
+        await rm(pdfFilePath, { force: true });
+      }
+    }
+  }
```
paologum, line 293: don't we want to upload it here

app/src/components/datapack_display/DatapackUploadForm.tsx
```diff
@@ -59,28 +59,64 @@ export const DatapackUploadForm: React.FC<DatapackUploadFormProps> = ({ close, u
...
+                {state.file && (
+                  <IconButton className="icon" onClick={() => setters.setFile(null)}>
+                    <DeleteOutline className="close-icon" />
+                  </IconButton>
+                )}
+              </Box>
+              <Box className="file-upload">
+                <InputFileUpload
+                  startIcon={<CloudUploadIcon />}
+                  text={"PDF Upload"}
```
paologum, line 99: translation

app/src/components/datapack_display/datapack-upload-form-hook.ts
```diff
@@ -113,6 +114,33 @@ const useDatapackUploadForm = (props: DatapackUploadFormProps) => {
...
+    const dataTransfer = new DataTransfer();
+    const existingFileNames = new Set(Array.from(pdfFiles).map((file) => file.name));
+    Array.from(pdfFiles).forEach((file) => dataTransfer.items.add(file));
+    Array.from(newFiles).forEach((file) => {
+      if (!existingFileNames.has(file.name)) {
+        dataTransfer.items.add(file);
+      }
+    });
+    setPDFFiles(dataTransfer.files);
+  };
```
paologum, line 132: don't most file uploads keep the most recent ie the one uploading and overwrite the existing rather than keep the oldest version (which i think is what this is doing)

app/src/components/datapack_display/datapack-upload-form-hook.ts
```diff
@@ -28,6 +28,7 @@ const useDatapackUploadForm = (props: DatapackUploadFormProps) => {
   const [date, setDate] = useState<Dayjs | null>(null);
   const [dateError, setDateError] = useState<string | null>(null);
   const [file, setFile] = useState<File | null>(null);
+  const [pdfFiles, setPDFFiles] = useState<FileList>(new DataTransfer().files);
```
paologum, line 31: is there a reason you chose a `FileList` other than an array? you pretty much just convert the `FileList` to an array every time you use it to my knowledge 
like `File[]`

server/src/upload-handlers.ts
```diff
@@ -276,6 +280,25 @@ export async function setupNewDatapackDirectoryInUUIDDirectory(
       await rm(datapackImageFilepath, { force: true });
     }
   }
+  if (pdfFields && Object.keys(pdfFields).length > 0) {
+    const filesDir = path.join(datapackFolder, "files");
+    if (!filesDir.startsWith(datapackFolder)) {
+      throw new Error("Invalid files directory path");
+    }
+    await mkdir(filesDir, { recursive: true });
```
paologum, line 288: this should be a function so we can refactor it if it ever changes
in `fetch-user-files`

it would handle the verification and `mkdir`
you also should use `verifyNonExistentFilepath` and `verifyFilepath` in their respective cases.
you can see other methods in that file to see what i mean

server/src/upload-handlers.ts
```diff
@@ -276,6 +280,25 @@ export async function setupNewDatapackDirectoryInUUIDDirectory(
...
+    for (const [pdfFileName, pdfFilePath] of Object.entries(pdfFields)) {
+      if (!pdfFilePath || !pdfFileName) continue;
+      const datapackPDFFilepathDest = path.join(filesDir, pdfFileName);
+      if (!datapackPDFFilepathDest.startsWith(filesDir)) {
+        throw new Error("Invalid datapack PDF filepath destination path");
+      }
```
paologum, line 294: `verifyNonExistentFilepath`

paologum (changes requested):
couple things looks good though on mobile and desktop

paologum (approved):

PR #330, opened 2025-02-06
Tsc 623 ranges on cross plots doesn't generate widths correctly

asivath:
Old:
image
New:
image


paologum (approved):

paologum:
imma verify frontend then i'll merge

PR #332, opened 2025-02-06
Tsc 610 make registration button

sbal15:
https://github.com/user-attachments/assets/1ede038a-00bb-4367-b1f3-83c3017f534b

Register Button frontend


app/src/WorkshopDetails.tsx
```diff
@@ -18,6 +20,26 @@ export const WorkshopDetails = observer(() => {
   const navigate = useNavigate();
   const { id } = useParams();
   const { t } = useTranslation();
+
+  const [isRegistered] = useState(true);
+  const [buttonSwitchNotReg, setButtonBool] = useState(false);
+  const [switchColorsBool, setColorBool] = useState(false);
+  const [loading, setLoading] = useState(false);
```
paologum, line 27: `isRegistered` doesn't change so you don't need it to be in a `useState`
(in this pr)
it would make more sense for you to have a variable called `disabled` that disables the button when the user is already registered or the workshop is not public

you don't need to set the color of disabling if you use `TSCLoadingButton` and use disabled

app/src/WorkshopDetails.tsx
```diff
@@ -83,9 +105,24 @@ export const WorkshopDetails = observer(() => {
...
+                      sx={{
+                        marginRight: 2,
+                        backgroundColor: switchColorsBool ? "grey" : "tan"
+                      }}
+                      onClick={handleRegisterClick}
+                      loading={loading}>
+                      {buttonSwitchNotReg
+                        ? t("workshops.details-page.registered-button")
+                        : t("workshops.details-page.register-button")}
+                    </LoadingButton>
```
paologum, line 120: use css for the `sx` and this should use `TSCLoadingButton`
ternary should go is this user already registered? if no then is this workshop able to be publicly registered?, if yes, then show button can be registered and not disabled. the other options i'm sure you can figure out

you should only have two variables
`isRegistered` `isPublicWorkshop`

from these two you can determine the disabled nature of the button and whether they can press it

paologum (changes requested):
Screenshot 2025-02-06 at 9 45 24 PM

if registered already, button should be disabled
additionally, it shouldn't wrap

i also don't see the code for tooltip when the workshop is private

sbal15:
https://github.com/user-attachments/assets/17cf8199-e53d-47c3-9f22-f310e426d608

sbal15:
already registered->registered
if you're already registered it should not be clickable and it should just show registered

sbal15:
https://github.com/user-attachments/assets/55884033-614f-472b-b40b-e4baa81a8c05

app/src/WorkshopDetails.tsx
```diff
@@ -118,6 +145,35 @@ export const WorkshopDetails = observer(() => {
...
+                      title={t("workshops.details-page.messages.not-registered")}
+                      open={showTooltip}
+                      placement="bottom">
+                      <div
+                        onMouseEnter={() => {
+                          if (!isRegistered && !isPublicWorkshop) {
+                            setShowTooltip(true);
+                          }
+                        }}
+                        onMouseLeave={() => setShowTooltip(false)}>
```
paologum, line 159: why do you need a hover div? why can't `open={!isRegistered && !isPublicWorkshop}`

paologum (changes requested):

shared/translations/en.json
```diff
@@ -605,7 +609,9 @@
             "USER_FETCH_HISTORY_FAILED": "Failed to fetch user history. Please try again later.",
             "USER_DELETE_HISTORY_FAILED": "Failed to delete user history. Please try again later",
             "USER_FETCH_HISTORY_METADATA_FAILED": "Failed to fetch user history metadata. Please try again later",
-            "USER_FETCH_HISTORY_FAILED_DATAPACKS_MISSING": "This chart entry couldn&apos;t be loaded due to missing files and has been removed from your history."
+            "USER_FETCH_HISTORY_FAILED_DATAPACKS_MISSING": "This chart entry couldn't be loaded due to missing files and has been removed from your history.",
+            "ADMIN_FETCH_WORKSHOPS_FAILED": "Unable to fetch workshops. Please try again later.",
+            "USER_WORKSHOP_FILE_NOT_FOUND_FOR_DOWNLOAD": "The workshop file requested was not found on the server. Please try again later or contact our support team."
```
paologum, line 614: these were probably removed for a reason, so you probably shouldn't add these back here

shared/translations/zh.json
```diff
@@ -415,7 +415,9 @@
             "ADMIN_WORKSHOP_ALREADY_EXISTS": "该工作坊名称已存在。请另选名称。",
             "FETCH_WORKSHOPS_FAILED": "无法抓取工作坊。请稍后再试。",
             "ADMIN_WORKSHOP_NOT_FOUND": "服务器未找到工作坊。该工作坊可能已结束或被删除。",
-            "WORKSHOP_FILE_NOT_FOUND_FOR_DOWNLOAD": "请求文件不存在，请稍后再试或联系开发团队"
+            "WORKSHOP_FILE_NOT_FOUND_FOR_DOWNLOAD": "请求文件不存在，请稍后再试或联系开发团队",
+            "ADMIN_FETCH_WORKSHOPS_FAILED": "无法抓取工作坊。请稍后再试。",
+            "USER_WORKSHOP_FILE_NOT_FOUND_FOR_DOWNLOAD": "请求文件不存在，请稍后再试或联系开发团队"
```
paologum, line 420: these were probably removed for a reason, so you probably shouldn't add these back here (same)

shared/translations/en.json
```diff
@@ -179,11 +179,15 @@
                 "files": "Files",
                 "download-button": "Download Files",
                 "description": "Description",
+                "register-button": "Register",
+                "registered-button": "Registered",
                 "messages": {
                     "no-datapacks": "No datapacks available",
-                    "no-files": "No files available"
+                    "download-tooltip-not-registered": "Contact jogg@purdue.edu to register!",
```
paologum, line 186: you moved this into `messages` but it used to be outside `messages` object, and i'm not sure you use this translation?

paologum (changes requested):

sbal15:
when I run yarn sync-translations, those changes are added to the json files, I do not know why

sbal15:
i tried deleting the changes and then rerunning yarn-sync translations, but it keeps adding them back after i deleted them

shared/translations/en.json
```diff
@@ -183,11 +183,14 @@
...
+                "register-button": "Register",
+                "registered-button": "Registered",
+                "download-tooltip-not-registered": "Contact jogg@purdue.edu to register!",
                 "messages": {
                     "no-datapacks": "No datapacks available",
-                    "no-files": "No files available"
-                },
-                "download-tooltip-not-registered": "Contact jogg@purdue.edu to register!"
+                    "no-files": "No files available",
+                    "not-registered": "Contact profogg@gmail.com to register!"
```
paologum, line 192: you should just use the `download-tooltip-not-registered` for this, `profogg@gmail.com` doesn't exist, I used it as his placeholder in the mock 😭

app/src/WorkshopDetails.tsx
```diff
@@ -16,11 +18,36 @@ export const WorkshopDetails = observer(() => {
...
+  };
+
   const fetchWorkshop = () => {
     if (!id) return;
     const workshop = state.workshops.find((d) => d.workshopId === Number(id));
     return workshop;
   };
+
+  fetchWorkshop();
+
```
paologum, line 50: is there a reason you call this here, it already gets called right below here

paologum (changes requested):

paologum (approved):

PR #333, opened 2025-02-07
Tsc 601 frontend wireframework

JoshuaYYHo:


paologum:
when you push a pr, make sure to include a picture of it in light and dark mode, and a video if applicable of your frontend

paologum:
additionally, can you resolve the conflicts so that the checks can run so i can review your pr

app/src/Help.tsx
```diff
@@ -148,6 +175,44 @@ export const Help = observer(function Help() {
...
+
+            <Routes>
+              <Route path="presets" element={<HelpPresets />} />
+              <Route path="datapacks" element={<HelpDatapacks />} />
+              <Route path="chart/*" element={<HelpChart />} />
+                <Route path="chart/what_is_a_chart" element={<HelpWhatIsAChart />} />
+                <Route path="chart/column_variants/*" element={<HelpColumnVariants />} />
+                  <Route path="chart/column_variants/block_columns" element={<HelpBlockColumn />} />
+                  <Route path="chart/column_variants/point_columns" element={<HelpPointColumns />} />
+                  <Route path="chart/column_variants/event_columns" element={<HelpEventColumns />} />
```
paologum, line 200: i want to make a quick comment since your checks haven't resolved but there is a lot of unneeded components

if we were to implement it this way, every new page would need a new component. we want this to be dynamic and editable by an admin. if we implement it this way we have to manually create each page ourself. to handle a more dynamic way of implementation, it should be done with a query onto the actual url. given the url's parameters, it should make a request to the server. if that request is fulfilled show the page with breadcrumbs, if not then don't show the page. this way we don't have individual components for each page

JoshuaYYHo:
Link to the demo: https://www.dropbox.com/scl/fi/owaoaynjf80c5sm0adpfy/Screen-Recording-2025-02-23-at-7.52.42-PM.mov?rlkey=wkvlyk09c6gvckse19c7nyent&st=p16pxzq2&dl=0

File size was too big, so I just uploaded to the DropBox

paologum:
* can you fix the animation on the accordion opening arrow so that it doesn't look like it's doing a full 360
* breadcrumbs should be blue probably the same color as `button.main`

app/src/HelpBreadcrumbsUpdated.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+        const contentType = response.headers.get("content-type");
+        if (!contentType || !contentType.includes("application/json")) {
+          throw new Error("Expected JSON response, but got something else.");
+        }
+
+        const data = await response.json();
+
+        if (Array.isArray(data.links)) {
+          setFetchedData(data.links);
```
paologum, line 55: just a comment, but this is where you will need to assert or write a specific `isHelpLinkArray` function to verify it is the object that we receive from the server, but until you create the backend for it you can leave it like this

also, you won't be sending the content in the links, it will just be the hierarchy itself, the content is fetched separately

app/src/HelpBreadcrumbsUpdated.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const pathBreadcrumbs = buildBreadcrumbsForPath(currentPath, allPaths);
+
+  const breadcrumbs: Breadcrumb[] = [{ to: "", label: "All Categories" }, ...pathBreadcrumbs];
+
+  // For debugging
+  useEffect(() => {
+    console.log("All available paths:", Object.keys(allPaths));
+    console.log("Current path:", currentPath);
+    console.log("Generated breadcrumbs:", breadcrumbs);
+  }, [fetchedData, currentPath]);
```
paologum, line 134: remove

app/src/HelpBreadcrumbsUpdated.tsx
```diff
@@ -0,0 +1,174 @@
...
+};
+
+// Tried fixing the ? in What is a chart, but I don't think it's working
+const normalizePath = (path: string): string => {
+  return path.split(/[?#]/)[0];
+};
+
+export default function NewBreadcrumbs() {
+  const [notFound, setNotFound] = useState(false);
+  const [fetchedData, setFetchedData] = useState<LinkPath[]>(defaultLinks);
```
paologum, line 31: wouldn't the type for this be better
all it is a tree/linked list variant where the existence in this hierarchy tells us whether to send another request for the content in this
```
type HelpLinks = {
  [name: string]: HelpLinks | null
}
```

app/src/HelpBreadcrumbsUpdated.tsx
```diff
@@ -0,0 +1,174 @@
...
+            color={index === breadcrumbs.length - 1 ? "text.primary" : "inherit"}>
+            {breadcrumb.label}
+          </Link>
+        ))}
+      </Breadcrumbs>
+
+      {breadcrumbs.length > 0 && (
+        <div>
+          <h2>{breadcrumbs[breadcrumbs.length - 1].label}</h2>
+          <p>{breadcrumbs[breadcrumbs.length - 1].content}</p>
```
paologum, line 169: use `Typography` and the component props for `h2` and `p`

app/src/HelpBreadcrumbsUpdated.tsx
```diff
@@ -0,0 +1,174 @@
...
+  if (notFound) {
+    return <PageNotFound />;
+  }
+
+  // So that the pages does not reload when you click back on a breadcrumb
+  const handleBreadcrumbClick = (breadcrumbPath: string, event: React.MouseEvent) => {
+    event.preventDefault();
+    window.history.pushState({}, "", `/help${breadcrumbPath}`);
+    window.dispatchEvent(new PopStateEvent("popstate"));
+  };
```
paologum, line 149: use the navigate function on `const navigate = useNavigate; navigate('/help')`

using window is something we want to try to avoid

app/src/HelpBreadcrumbsUpdated.tsx
```diff
@@ -0,0 +1,174 @@
...
+  // For debugging
+  useEffect(() => {
+    console.log("All available paths:", Object.keys(allPaths));
+    console.log("Current path:", currentPath);
+    console.log("Generated breadcrumbs:", breadcrumbs);
+  }, [fetchedData, currentPath]);
+
+  useEffect(() => {
+    setNotFound(breadcrumbs.length <= 1 && currentPath !== "" && currentPath !== "/");
+  }, [breadcrumbs, currentPath]);
```
paologum, line 138: if you use the type i talked about above, you can just make a function that sees if the encoded link exists in the `HelpLink` type you get hard coded or from the server and then `setNotFound` appropriately

app/src/HelpDrawerUpdated.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function NavItem({ link, parentPath = "" }: { link: LinkPath; parentPath?: string }) {
+  const [open, setOpen] = useState(false);
+  const navigate = useNavigate();
+
+  // Construct full path for navigation
+  const formattedPath = `${parentPath}/${link.Title.toLowerCase().replace(/\s+/g, "-")}`;
+
+  console.log("Generated Path:", formattedPath); // Debugging
```
paologum, line 20: remove

paologum (changes requested):

paologum:
Screenshot 2025-02-24 at 3 21 57 PM

Screenshot 2025-02-24 at 3 22 01 PM

padding is different than mock

server/assets/translation.xlsx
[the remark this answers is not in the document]
JoshuaYYHo: my bad i highkey just git add ., so i know next time to not do that

app/src/HelpBreadcrumbsUpdated.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const fullPath = path.startsWith("/help") ? path : "/help" + path;
+    const normalizedSearchPath = normalizeForComparison(fullPath);
+
+    const matchingOriginalPath = pathMap[normalizedSearchPath];
+
+    if (!matchingOriginalPath) {
+      console.log("No matching path found");
+      return [];
+    }
+
```
paologum, line 117: why don't you just pop and construct the breadcrumbs with just the path in the url, why do you need the full path and the normalized path (which i'm not sure why we need both but that's a different comment)

paologum (changes requested):
why aren't breadcrumbs

app/src/components/help-menu-items.ts
paologum, line 1: i don't think you use this?

app/src/HelpDrawerUpdated.tsx
```diff
@@ -0,0 +1,183 @@
...
+  }
+];
+
+function NavItem({ link, parentPath = "" }: { link: LinkPath; parentPath?: string }) {
+  const [open, setOpen] = useState(false);
+  const navigate = useNavigate();
+  const theme = useTheme(); // Moved inside the component
+
+  // Construct full path for navigation
+  const formattedPath = `${parentPath}/${link.title.toLowerCase().replace(/\s+/g, "-")}`;
```
paologum, line 105: you use the same formatting as the format in the breadcrumbs component, so you should probably put that format function in `non-action-util.ts` or some util file then use it for both cases, `decodeUriComponent` can stay and be used in both sections i think

app/src/HelpDrawerUpdated.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <ListItem disablePadding>
+        <ListItemButton
+          onClick={handleClick}
+          sx={{
+            flexDirection: "row-reverse",
+            py: 0,
+            my: 0,
+            height: "24px",
+            width: "300px"
+          }}>
```
paologum, line 130: i had a previous comment abt this, can you move this to a css file like `Help.css`

app/src/HelpDrawerUpdated.tsx
```diff
@@ -0,0 +1,92 @@
...
+      {hasChildren(link) && (
+        <Collapse in={open} timeout="auto" unmountOnExit>
+          <List
+            component="div"
+            disablePadding
+            sx={{
+              pl: 1,
+              borderLeft: `1px solid ${theme.palette.divider}`,
+              marginLeft: "27.5px",
+              "& .MuiListItem-root": { margin: 0 }
```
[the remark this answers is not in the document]
paologum: comment not resolved

app/src/HelpDrawerUpdated.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+function NewHelpDrawer() {
+  return (
+    <List
+      disablePadding
+      sx={{
+        "& .MuiListItem-root": {
+          margin: 0
+        }
+      }}>
```
paologum, line 175: lot of space and since you are making a css file, add this there too

paologum (changes requested):

app/src/HelpDrawerUpdated.tsx
```diff
@@ -0,0 +1,170 @@
...
+      {/* Render children if applicable */}
+      {hasChildren(link) && (
+        <Collapse in={open} timeout="auto" unmountOnExit>
+          <List
+            component="div"
+            disablePadding
+            className="menu-item"
+            sx={{
+              // Can't directly feed theme.palette.divider into the css file
+              borderLeft: `1px solid ${theme.palette.divider}`
```
paologum, line 149: `theme.palette.accordionLine.main`

paologum (approved):

PR #334, opened 2025-02-07
Tsc 530 wire recaptcha to match strings on backend

Michael-Knaack:
Adds to the verify.ts to verify the reCAPTCHA strings match

### DIFF
--- a/server/__tests__/admin-routes.test.ts
+++ b/server/__tests__/admin-routes.test.ts
@@ -481,59 +481,62 @@ const routes: { method: HTTPMethods; url: string; body?: object }[] = [
 ];
 const headers = { "mock-uuid": "uuid", "recaptcha-token": "recaptcha-token", "recaptcha-action": "test-action" };
 
-describe("verifyAdmin tests", () => {
-  describe.each(routes)("should return 401 for route $url with method $method", ({ method, url, body }) => {
-    const findUser = vi.spyOn(database, "findUser");
+describe("verifyRecaptcha tests", () => {
+  describe.each(routes)("should return 400 or 422 for route $url with method $method", ({ method, url, body }) => {
+    const checkRecaptchaTokenMock = vi.spyOn(verify, "checkRecaptchaToken");
     beforeEach(() => {
-      findUser.mockClear();
+      checkRecaptchaTokenMock.mockClear();
     });
-    test("should return 401 if not logged in", async () => {
-      const response = await app.inject({
-        method: method as InjectOptions["method"],
-        url: url,
-        payload: body
-      });
-      expect(findUser).not.toHaveBeenCalled();
-      expect(await response.json()).toEqual({ error: "Unauthorized access" });
-      expect(response.statusCode).toBe(401);
-    });
-    test("should return 401 if not found in database", async () => {
-      findUser.mockResolvedValueOnce([]);
+
+    it("should return 400 if missing recaptcha token", async () => {
       const response = await app.inject({
         method: method as InjectOptions["method"],
         url: url,
         payload: body,
-        headers
+        headers: { ...headers, "recaptcha-token": "", "recaptcha-action": "test-action" }
       });
-      expect(findUser).toHaveBeenCalledWith({ uuid: headers["mock-uuid"] });
-      expect(findUser).toHaveBeenCalledTimes(1);
-      expect(await response.json()).toEqual({ error: "Unauthorized access" });
-      expect(response.statusCode).toBe(401);
+      expect(checkRecaptchaTokenMock).not.toHaveBeenCalled();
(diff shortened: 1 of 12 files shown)


server/__tests__/admin-routes.test.ts
```diff
@@ -469,48 +470,62 @@ describe("verifyAdmin tests", () => {
...
+      checkRecaptchaTokenMock.mockClear();
     });
+
     it("should return 400 if missing recaptcha token", async () => {
       const response = await app.inject({
         method: method as InjectOptions["method"],
         url: url,
         payload: body,
-        headers: { ...headers, "recaptcha-token": "" }
+        headers: { ...headers, "recaptcha-token": "", "recaptcha-action": "test-action" }
```
paologum, line 485: this will only need to miss `recaptcha-token`, you don't need to add `recaptcha-action` on top of this

server/__tests__/workshop-routes.test.ts
```diff
@@ -135,45 +135,60 @@ describe("verifyAuthority", () => {
...
+      checkRecaptchaTokenMock.mockClear();
     });
+
     it("should return 400 if missing recaptcha token", async () => {
       const response = await app.inject({
         method: method as InjectOptions["method"],
         url: url,
         payload: body,
-        headers: { ...headers, "recaptcha-token": "" }
+        headers: { ...headers, "recaptcha-token": "", "recaptcha-action": "test-action" }
```
paologum, line 148: same here

server/src/admin/admin-auth.ts
```diff
@@ -46,20 +46,31 @@ async function verifyAdmin(request: FastifyRequest, reply: FastifyReply) {
...
-    const score = await checkRecaptchaToken(recaptcha);
+    const score = await checkRecaptchaToken(recaptchaToken, action);
     if (score < googleRecaptchaBotThreshold) {
-      reply.status(422).send({ error: "Recaptcha failed" });
+      reply.status(422).send({ error: "recaptcha failed" });
       return;
     }
   } catch (e) {
-    reply.status(500).send({ error: "Recaptcha error" });
+    reply.status(500).send({ error: "recaptcha error" });
```
paologum, line 73: i'm not sure why you uncapitalized this, but i would just leave it capitalized

server/src/workshop/workshop-auth.ts
```diff
@@ -36,20 +36,31 @@ async function verifyAuthority(request: FastifyRequest, reply: FastifyReply) {
...
-    const score = await checkRecaptchaToken(recaptcha);
+    const score = await checkRecaptchaToken(recaptchaToken, action);
     if (score < googleRecaptchaBotThreshold) {
-      reply.status(422).send({ error: "Recaptcha failed" });
+      reply.status(422).send({ error: "recaptcha failed" });
       return;
     }
   } catch (e) {
-    reply.status(500).send({ error: "Recaptcha error" });
+    reply.status(500).send({ error: "recaptcha error" });
```
paologum, line 63: same here

paologum (changes requested):

asivath (changes requested):
a couple of things to get this to all work correctly:

1. on the frontend whenever we call a route that uses recaptcha we pass a token to it, for example in the function adminFetchUsers in admin-actions.ts we call: 
`const recaptchaToken = await getRecaptchaToken("adminDeleteUsers");`. That `adminDeleteUsers` token is the token we care about. 

2. what you need to do is modify the prehandlers in each auth file (workshop-auth.ts, user-auth.ts, admin-auth.ts) to verify that the action in the token matches what the route expects. in the auth files there is this line: `fastify.addHook("preHandler", verifyRecaptcha);` which adds the function `verifyRecaptcha` as a global preHandler (code that gets run before actually accessing that route). we no longer want a global preHandler since we need to define an action for each route. this means for every route defined in the auth files you will need to change this:
```
fastify.delete(
    "/user",
    {
      schema: {
        body: adminUUIDbody
      },
      config: {
        rateLimit: moderateRateLimit
      }
    }
```
to this:
```
fastify.delete(
  "/user",
  {
    preHandler: [verifyRecaptcha("adminDeleteUser")],
    schema: { body: adminUUIDbody },
    config: { rateLimit: moderateRateLimit }
  },
  adminDeleteUser
);
```
3. to get the types happy you'll probably need to make the verifyRecaptcha function into a function factory like this:
```
export function verifyRecaptcha(expectedAction: string) {
  return async function (request: FastifyRequest, reply: FastifyReply) {
  ...
  };
}
```
4. you'll need to go to each fetch on the frontend, gets its corresponding token, and then pass it to your new prehandlers on the backend in each auth file.

lmk if you have any questions

PR #336, opened 2025-02-10
Tsc 617 parsing ideas sql to txt

paologum:
* mySql -> csv -> sql lite
* hard coding the kysley table
* i create a dummy `kysley-schema.ts` file that gets created on mySql->csv to help us with that conversion. i don't think we want to necessarily have the auto generated schema from my code pushed to the main so i just copied it so it's more or less manual but exists in two places (only one is imported and in `src`). if it ever changes or gets migrated we would just copy it again. i can see both ways but it feels like the compiled js files if we keep a generated file on the repo and people accidentally change it in their repos if they didn't update their sql dump or something
* not really sure when to do the conversion, if you have ideas let me know ie i'm thinking the same as our database, but would we just not update it if it exists -> haven't thought about yet
* `yarn london` lol for now it's just a command and is not imported in the normal flow of the website since it doesn't affect it
* tbh the fields are not like he described it in the powerpoint so i'm not really sure what to do with this, i've emailed him to clarify
* wish i could use a library, but none specialize in going dump -> sqllite so i coded them myself. csv -> sqllite is straightforward, i can handle errors and there shouldn't be much variation. filp side, dump -> csv is probably better done by a library because there are edge cases i probably haven't caught. probably really minor but i couldn't find a library so i settled on this. lmk anything u find

Screenshot 2025-02-09 at 11 34 53 PM

### DIFF
--- a/server/src/london-db/types.ts
+++ b/server/src/london-db/types.ts
@@ -1,0 +1,115 @@
+//   UNIQUE: ["Taxon"],
+//   PRIMARY KEY: ["id"],
+export interface CPdinos {
+  id: number; // int unsigned NOT NULL
+  Taxon: string; // varchar(255) NOT NULL DEFAULT ''
+  citation: string | null; // varchar(255) DEFAULT NULL
+  basionym: string | null; // varchar(255) DEFAULT ''
+  synonyms: string | null; // varchar(255) DEFAULT NULL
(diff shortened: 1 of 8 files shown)


server/src/london-db/types.ts
paologum, line 1: this is the copied auto generated file
technically manual, but created and copied here from `db/london`

PR #339, opened 2025-02-13
Tsc 450 dcc frontend

jay-jhk-lee:


paologum:
image

i would put the error icon on the other side so the hierarchy stays lined up
additionally, people might not know why the names are grayed out, so if they are disabled they probably should have a tooltip (make sure the tooltips properly load for the icon and the disabled name)

paologum:
also clicking to expand a column doesn't expand unless you click the name or the arrow. it probably should expand when clicking on the whole row

paologum:
https://github.com/user-attachments/assets/8605ab82-a0cb-4d78-8218-db9155e8279e

disabled still clickable

paologum:
https://github.com/user-attachments/assets/66c11cfb-02af-4d51-926a-cdd57eb42543

expanding the whole thing wraps, and unless you zoom out you can shrink it, probably just set a max width on it

paologum:
image

smaller screens don't have the same font change

paologum:
last thing, this is probably something that aaron would comment and something i experienced. going to the overlay the display overlay looks like it can be clicked, but it can't until you choose a column. it probably should be disabled at the very least until they choose a column

jay-jhk-lee:
https://github.com/user-attachments/assets/20f81ebf-9ceb-4d3e-b4b1-f8cf05273fa5

jay-jhk-lee:
https://github.com/user-attachments/assets/922229be-4923-4d77-836d-953bc7905158

app/src/settings_tabs/advanced_settings/OverlaySettings.tsx
```diff
@@ -0,0 +1,252 @@
...
+      <ColumnContainer
+        sx={{
+          cursor: "pointer"
+        }}
+        className="dcc-column-leaf">
+        <Typography className="column-display-name">{column.editName}</Typography>
+      </ColumnContainer>
+    );
+    const NotValidOverlay = () => (
+      <CustomTooltip title="Not an Event or Point Column" placement="left" offset={[0, 10]}>
```
paologum, line 182: this should be a translation

app/src/util/util.ts
```diff
@@ -1,5 +1,24 @@
...
+ * @param name
+ * @returns
+ */
+
+export function discardTscPrefix(name: string | null) {
+  if (!name) {
+    return "";
+  }
+  return name.split(":")[1];
+}
```
paologum, line 17: being super robust, should have checks for if the split is invalid/`:` doesn't exist so there is only one entry in the array

app/src/util/util.ts
```diff
@@ -1,5 +1,24 @@
...
+
+export function prependDualColCompColumnName(text: string): string {
+  return dualColCompPrefix + text;
+}
```
paologum, line 21: do you use this? it seems you just add them in the components above. my preference would probably be to use this but its up to you, otherwise remove this

paologum (changes requested):

app/src/settings_tabs/Column.tsx
```diff
@@ -54,8 +56,16 @@ export const ColumnDisplay = observer(() => {
       scrollRef.current.scrollTo({ top: 0, behavior: "smooth" });
     }
   };
+  const temp = async () => {
+    const value = state.datapackMetadata[8];
+    const datapack = await actions.fetchDatapack(value);
+    actions.addDatapack(datapack!);
+    actions.setUnsavedDatapackConfig([...state.unsavedDatapackConfig, value]);
+    await actions.processDatapackConfig(toJS(state.unsavedDatapackConfig));
+  };
```
paologum, line 65: remove

paologum (approved):

PR #343, opened 2025-02-19
Tsc 619 dot creation

paologum:
# Main Changes
* added markers (different from models)
* have ingrained tooltips into the svg and apply the circle marker on the svg itself
* multiple types, ability to change the age/depth (positioning) of the marker
* change color
* click to remove
* toolbar options are now dynamic for the chart so we can change them as we see fit for cross plot and regular chart
* as normal there is no mobile, that will be new task bc too big if i did that here

# Demo

https://github.com/user-attachments/assets/112aff5e-9ee7-4551-9bc8-ab78aaf59b7f

### DIFF
--- a/app/src/Chart.tsx
+++ b/app/src/Chart.tsx
@@ -76,18 +76,16 @@ export const Chart: React.FC<ChartProps> = observer(({ Component, style, refList
   // we need to setup the chart alignment values when the chart content changes
   // if the user generates again on the chart tab, we have to toggle the change by making sure to set setup false
   useEffect(() => {
+    if (!chartContent || !madeChart) return;
     setSetup(false);
-    const container = transformContainerRef.current;
-    const content = svgContainerRef.current;
-    if (!container || !content) return;
     setChartAlignmentValues();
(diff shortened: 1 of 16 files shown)


app/src/Chart.tsx
```diff
@@ -135,15 +121,30 @@ export const Chart: React.FC<ChartProps> = observer(({ Component, style, refList
...
-    zoomFitScale
-  ]);
+  });
+
+  // we need to setup the chart alignment values when the chart content changes
+  // if the user generates again on the chart tab, we have to toggle the change by making sure to set setup false
+  useEffect(() => {
+    if (!chartContent || !madeChart || !svgContainerRef?.current) return;
+    setChartAlignmentValues();
+  }, [chartContent, madeChart, svgContainerRef?.current]);
```
paologum, line 131: this is bad, but it makes the smallest of flashing before it actually fits,
i've tried everything i know is possible so i can't remove the `svgContainerRef?.current` here

asivath (approved):

PR #353, opened 2025-02-28
Tsc 549 backend route to edit a user as an admin

Ganthry:


paologum:
add demo

Ganthry:
https://github.com/user-attachments/assets/9b8dce8b-e7ae-464f-81ef-0c574cdba2ae

paologum:
* you shouldn't be able to edit the root user to not be an admin
* there should be a confirm changes when you exit before saving

Ganthry:
> * there should be a confirm changes when you exit before saving

Theres already a dialog that pops up saying "You have unsaved changes. Are you sure you want to discard them?" Should I change this to ask to confirm changes or is this for something else?

Ganthry:
https://github.com/user-attachments/assets/49dd461d-b419-4326-a29d-3d56a03d7edf

paologum:
also, since you can't change the root admin after instantiation, can you make it a pro user as well

paologum:
i still see the change after it fails when trying to edit an admin user

app/src/admin/edit-user-stats-hook.ts
```diff
@@ -42,36 +48,33 @@ const useEditUser = ({ data }: UseUserStatsProps) => {
...
     }
+    const modifiedUser = {
+      username: userInfo.username,
+      email: userInfo.email,
+      accountType: userInfo.accountType,
+      isAdmin: userInfo.isAdmin ? 1 : 0
+    };
+
+    const resp = await actions.adminModifyUsers(modifiedUser);
+    console.log(resp);
```
paologum, line 74: log

app/src/types.ts
```diff
@@ -228,6 +228,7 @@ export type EditableUserProperties = {
   username: string;
   email: string;
   isAdmin: boolean;
+  accountType: string;
```
paologum, line 231: this type should only be editable user properties, so pictureUrl, username, email shouldn't be there

paologum (changes requested):

app/src/admin/edit-user-stats-hook.ts
Ganthry: woops

app/src/types.ts
Ganthry: The EditableUserProperties is the same type that is used to display all the information about a user in the modify user dialog (AdminShowAdditionalUserInfo). Would it be better to rework how this works and add another type? Or could I rename EditableUserProperties to something more appropriate, like showAdditionalUserInfoProperties?

paologum: i would just limit the type ti what i said and the display can just straight up use data from the component
if you look at the hook im p sure it just passes it along in the custom hook

app/src/types.ts
```diff
@@ -225,9 +225,8 @@ export type ChartSettings = {
 };
 
 export type EditableUserProperties = {
-  username: string;
-  email: string;
   isAdmin: boolean;
+  accountType: string;
   pictureUrl: string | undefined;
```
paologum, line 230: also remove picture Url, this isn't supposed to be editable by admins

paologum:
it's asking for me to discard changes when i haven't made any changes

https://github.com/user-attachments/assets/35ce70f9-4b1d-4af3-bcf6-3ca798a17717

monitoring/prometheus.yml
```diff
@@ -17,11 +17,26 @@ scrape_configs:
...
+  - job_name: "tscstaging"
+    scrape_interval: 20s
+    static_configs:
+      - targets: ["staging.timescalecreator.org:3000"]
+
+  - job_name: "tscdeep"
+    scrape_interval: 20s
+    static_configs:
+      - targets: ["timescalecreator.deep-time.org:3000"]
+
```
paologum, line 39: is this related to your other pr? do you still want this here in this pr

Ganthry: yea it'd be better for me to do this in the prom pr

paologum (approved):

PR #369, opened 2025-03-12
365 setup local dropbox

paologum:
pull datapacks and the jar if they don't exist locally, this runs on the `yarn dev` of `root` and also `yarn pull-supported-files` in `server`

doesn't affect the rest of the command if the api fetch fails (if dropbox is down or you are deving offline)

.env will probably be on the dropbox


server/src/pull-supporting-files.ts
```diff
@@ -0,0 +1,136 @@
...
+    body: JSON.stringify({
+      path: "/VIP-Fall2023 EarthHistoryVisualization/Supporting TSCOnline Files/datapacks"
+    })
+  });
+  if (!response.ok) {
+    throw new Error("Failed to list files in datapacks directory");
+  }
+  const data = await response.json();
+  const files = data.entries.map((entry: { name: string }) => entry.name);
+  if (!verifyFile(assetconfigs.datapacksDirectory)) {
```
asivath, line 77: await verifyFile

asivath (changes requested):

asivath (approved):

PR #382, opened 2025-03-25
Tsc 635 pdf file frontend download maybe view

lbostre:
- ability to download files that are attached to datapacks
- 
4a73d11c4e24ee0309b32022ba632ccb
4d27b08d53165bbbb354b367b4e0f329
73b5c31d99d6e1d07b60b24f67d89b42
c87eda6d9c6e38eae6c7fe56c4a9ed91


app/src/state/actions/general-actions.ts
```diff
@@ -132,6 +132,54 @@ export const fetchPublicOfficialDatapack = action(
   }
 );
 
+export const fetchDatapackFilesForDownload = action(async (datapackTitle: string, uuid: string, isPublic: boolean) => {
+  const recaptchaToken = await getRecaptchaToken("fetchDatapackFilesForDownload");
+  if (!recaptchaToken) return null;
+  try {
+    const response = await fetcher(
+      `/user/datapack/download/files/${encodeURIComponent(datapackTitle)}/${uuid}/${isPublic}`,
```
paologum, line 140: for the route, since it is a getter the standard is not to include any verbs in the routes so you should remove download in the names

app/src/state/actions/general-actions.ts
```diff
@@ -132,6 +132,54 @@ export const fetchPublicOfficialDatapack = action(
...
+          });
+          if (typeof reader.result !== "string") {
+            throw new Error("Invalid File");
+          }
+          fileURL = reader.result;
+          await downloadFilesByURL(fileURL, `FilesFor${datapackTitle}.zip`);
+        } catch (e) {
+          pushError(ErrorCodes.INVALID_PATH);
+        }
+      }
```
paologum, line 168: there is a function called `downloadFile` that you can use for this

app/src/state/actions/util-actions.ts
```diff
@@ -116,3 +116,21 @@ export function checkIfDccDataIsInRange(dccColumn: ColumnInfo, userTopAge: numbe
...
+    const aTag = document.createElement("a");
+    aTag.href = fileURL;
+    aTag.setAttribute("download", filename);
+    document.body.appendChild(aTag);
+    aTag.click();
+    aTag.remove();
+  } catch (error) {
+    pushError(ErrorCodes.INVALID_PATH);
+  }
+};
```
paologum, line 136: same as `downloadFiles` in `non-action-util` unless i'm understanding incorrectly

app/src/util/util.ts
```diff
@@ -223,3 +223,7 @@ export function findSerialNum(name: string) {
   }
   return serialNumber;
 }
+
+export function getDatapackOfficialOrUUID(datapack: Datapack) {
+  return datapack.type === "user" ? datapack.uuid : "official";
+}
```
paologum, line 229: use the `shared` function for `getUUIDOfDatapackType`

server/assets/translation.xlsx
paologum, line 1: you are removing `translation` items or is this the merge conflict

server/src/routes/user-routes.ts
```diff
@@ -400,3 +411,66 @@ export const downloadWorkshopFilesZip = async function downloadWorkshopFilesZip(
...
+  }
+  let zipfile = path.resolve(directory, `filesFor${datapackTitle}.zip`); //could be non-existent
+  if (!(await verifyNonExistentFilepath(zipfile))) {
+    reply.status(500).send({ error: "Invalid directory path" });
+    return;
+  }
+  if (!zipfile.startsWith(path.join(process.cwd(), directory))) {
+    reply.status(500).send({ error: "Invalid directory path" });
+    return;
+  }
```
paologum, line 447: i assume you did the bottom if statement for the code ql, but `verifyNonExistent` should handle that so you should remove the second if statement here

server/src/routes/user-routes.ts
```diff
@@ -400,3 +411,66 @@ export const downloadWorkshopFilesZip = async function downloadWorkshopFilesZip(
...
+      if (error.code !== "ENOENT") {
+        reply.status(500).send({ error: "An error occurred: " + e });
+        return;
+      }
+    }
+
+    if (!file) {
+      file = await createZipFile(zipfile, filesDir);
+    }
+    reply.send(file);
```
paologum, line 467: this could probably be a lot less lines

like
```
if (await verifyFilepath(zipFile))
await readFile
else
await createZipFile()
```

app/src/DatapackProfile.tsx
```diff
@@ -382,6 +384,13 @@ const About: React.FC<AboutProps> = observer(({ datapack }) => {
       window.removeEventListener("beforeunload", handleBeforeUnload);
     };
   }, [state.datapackProfilePage.unsavedChanges]);
+
+  function downloadDatapackFiles() {
+    if (checkUserAllowedDownloadDatapack(state.user, datapack)) {
+      actions.fetchDatapackFilesForDownload(datapack.title, getDatapackOfficialOrUUID(datapack), datapack.isPublic);
+    }
```
paologum, line 391: can you push a warning that says you are not allowed to download this datapacks pdfs

paologum (changes requested):
ok so i never checked this, but during one of your prs, the admin upload workflow fails for datapacks. can you fix that in this pr because i believe you changed or someone changed the field where the file goes and it may think that the file field is empty or something along those lines.

also does `hasFiles` get sent in the admin actions upload function

server/assets/translation.xlsx
lbostre: merge conflict

paologum (changes requested):

paologum:
datapack metadata isn't properly being sent on the pr preview

paologum:
you also need to change the `dev-config.json` file in the server if you change datapack stuff to include the `hasFiles`

server/src/routes/user-routes.ts
paologum, line 1: i didn't see this before, but i don't want to have to go back myself making these tests so add a test for this, it's not in coverage so it didn't popup unfortunately, also it will need to be added to that routes array

server/__tests__/upload-datapack.test.ts
paologum, line 1: on main, i merged something that requires pdfFields to not have to be a required parameter jsyk

app/src/state/actions/general-actions.ts
```diff
@@ -329,10 +370,14 @@ export const uploadUserDatapack = action(
           pushError(ErrorCodes.USER_FETCH_DATAPACK_FAILED);
           return;
         }
+        if (pdfFiles?.length) {
+          datapack.hasFiles = true;
+        }
```
paologum, line 375: this shouldn't alter the datapack, it's field should be representative of what's on the server and the app, the app shouldn't make a change to it that would make it different than the server

if this is giving an incorrect value it should be changed properly on a datapack upload, not on a fetch

app/src/state/actions/admin-actions.ts
```diff
@@ -358,6 +358,9 @@ export const adminUploadOfficialDatapack: UploadDatapackMethodType = action(
       pdfFiles.forEach((pdfFile) => {
         formData.append("pdfFiles[]", pdfFile);
       });
+      formData.append("hasFiles", "true");
+    } else {
+      formData.append("hasFiles", "false");
```
paologum, line 363: if you already have the `hasFiles` logic written in the datapack metadata that you send in the form, you should use that instead of this. this is more robust obviously but if you decide to do it this way, there is no need to have the hook have the `hasFiles` variable. it should just be one or the other

paologum (changes requested):

app/src/state/actions/general-actions.ts
paologum: same thing this code shouldn't be here

paologum (changes requested):

paologum (approved):

PR #383, opened 2025-03-25
Tsc 498 monitoring with prometheus grafana

Ganthry:


paologum:
can you add a description of what this pr is doing

Ganthry:
This PR adds stage, prod, and deep into prometheus monitoring. Now the prometheus instant running on dev requests the metrics from these other TSC servers in addition to dev. Also the metrics endpoint requires an authentication token which is provided to the prometheus instant when deployed with dev. Every instant has the token in its server .env.

paologum:
it would be better if the grafana had the status at the top regardless of the instant you are on

.github/workflows/deploy-dev.yml
```diff
@@ -297,16 +297,20 @@ jobs:
 
               GRAFANA_USERID=$(grep 'GRAFANA_USERID' $MONITORING_DIR/.env | cut -d '=' -f2)
               GRAFANA_PSWD=$(grep 'GRAFANA_PSWD' $MONITORING_DIR/.env | cut -d '=' -f2)
+              METRICS_AUTH=$(grep 'METRICS_AUTH' $MONITORING_DIR/.env | cut -d '=' -f2)
 
-              if [ -z "$GRAFANA_USERID" ] || [ -z "$GRAFANA_PSWD" ]; then
-                echo "Error: Grafana credentials are missing in .env file."
+              if [ -z "$GRAFANA_USERID" ] || [ -z "$GRAFANA_PSWD" ] || [ -z "$METRICS_AUTH" ]; then
+                echo "Error: Credentials are missing in .env file."
```
paologum, line 303: i forget is this env variable instantiated in the action or do you manually put it in, and is it automatically updated/maintained on each instance

Ganthry: This adds the same thing I did before w Aditya and the Grafana username and Grafana pswd. It checks the .env file aaron/monitoring/.env and replaces all of them with the actual keys. The .env file was made manually. The metric token was given to the server/.env for each server dev,prod,stage,deep. None of the other servers need the monitoring folder setup bc the only prom instance running is on dev and requests the metrics from each tsconline server. (This prom instant should be the only instant able to access each servers /metrics endpoint bc it has the auth token)

paologum (approved):

PR #388, opened 2025-03-26
Tsc 628 settings datapack chart history applies settings

asivath:
https://github.com/user-attachments/assets/a1bc5cb2-2949-4582-a73e-ec6b62fb91b6


app/src/Chart.tsx
```diff
@@ -10,14 +10,17 @@ import {
...
 import { OptionsBar } from "./ChartOptionsBar";
-import { Box, Typography } from "@mui/material";
+import { Box, Typography, IconButton, Drawer, List, ListItemButton, ListItemText, Paper } from "@mui/material";
+import HistoryIcon from "@mui/icons-material/History";
+import DeleteForeverIcon from "@mui/icons-material/DeleteForever";
 import { useTheme } from "@mui/material/styles";
 import { useTranslation } from "react-i18next";
 import { ChartContextType } from "./types";
-import { context } from "./state";
+import { context, state } from "./state";
```
paologum, line 19: no import state, use context

app/src/Chart.tsx
```diff
@@ -182,6 +185,7 @@ export const Chart: React.FC<ChartProps> = observer(({ Component, style, refList
 
   return (
     <Box className="chart-container">
+      {state.isLoggedIn && <HistorySideBar />}
```
paologum, line 188: from above comment, this doesn't use context

app/src/Chart.tsx
```diff
@@ -254,3 +258,82 @@ export const ChartTab: React.FC = observer(() => {
...
+                    actions.pushSnackbar("No history entries to delete", "warning");
+                    return;
+                  }
+                  setLoading(true);
+                  try {
+                    await actions.deleteUserHistory("-1");
+                  } finally {
+                    setLoading(false);
+                  }
+                }}>
```
paologum, line 301: just in case other functions use `deleteUserHistory` shouldn't this code be inside `deleteUserHistory` (not the loading stuff)

server/__tests__/admin-routes.test.ts
```diff
@@ -1135,27 +1136,6 @@ describe("getUsers", () => {
...
-      ...testSharedAdminUser,
-      userId: 123,
-      isAdmin: true,
-      isGoogleUser: false,
-      invalidateSession: false,
-      emailVerified: true
-    });
-    expect(await response.json()).toEqual({ error: "Unknown error" });
-    expect(response.statusCode).toBe(404);
-  });
```
paologum, line 1158: what was wrong with this test case

.github/workflows/deploy-prod.yml
```diff
@@ -108,6 +108,7 @@ jobs:
               ! -name "release-artifacts.zip" \
               ! -path "${{ secrets.RELEASE_PATH }}/.yarn*" \
               ! -path "${{ secrets.RELEASE_PATH }}/server/db/TSC.db" \
+              ! -path "${{ secrets.RELEASE_PATH }}/server/assets/uploads/*" \
```
paologum, line 111: does prod and the dde stuff have the uploads of other things

server/src/admin/admin-routes.ts
```diff
@@ -154,13 +154,11 @@ export const getUsers = async function getUsers(_request: FastifyRequest, reply:
           isAdmin: user.isAdmin === 1,
           emailVerified: user.emailVerified === 1,
           invalidateSession: user.invalidateSession === 1,
-          ...(workshopIds.length > 0 && { workshopIds })
+          ...(workshopIds.length > 0 && { workshopIds }),
+          historyEntries: []
```
paologum, line 158: are these just going to be fetched at some other time? this might be misleading so maybe we do an omit type for displayed users without this ? do you think admins should see history

server/src/routes/routes.ts
```diff
@@ -264,6 +266,11 @@ export const fetchChart = async function fetchChart(request: FastifyRequest, rep
     } else {
       console.log("Request for chart that already exists (hash:", hash, ".  Returning cached version");
       reply.send({ chartpath: chartUrlPath, hash: hash }); // send the browser back the URL equivalent...
+      // after sending, save to history if user is logged in
+      if (uuid)
+        await saveChartHistory(uuid, settingsFilePath, datapacksToSendToCommandLine, chartFilePath, hash).catch((e) => {
+          logger.error(`Failed to save chart history for user ${uuid}: ${e}`);
+        });
```
paologum, line 273: how will this handle cross plots, because this will also technically save cross plot history
this fetch also does cross plots for better or for worse lol but
preview doesn't work yet so i'm not too sure

server/src/user/chart-history.ts
```diff
@@ -0,0 +1,158 @@
+import { assetconfigs } from "../util.js";
+import { join, basename, dirname } from "path";
+import { cp, mkdir, readFile, readdir, realpath, rm, symlink, access, lstat } from "fs/promises";
+import logger from "../error-logger.js";
+import { Datapack, HistoryEntry } from "@tsconline/shared";
+import { getCachedDatapackFromDirectory } from "./fetch-user-files.js";
+
+export function isValidEpoch(timestamp: string) {
+  return /^\d{13}$/.test(timestamp);
+}
```
paologum, line 10: is this not a library function 😭

server/src/user/chart-history.ts
paologum, line 1: tests?

paologum (changes requested):
deploy also fails to fetch user history metadata

paologum:
similar to like google docs, if there are new charts that the user hasn't viewed, i was thinking maybe it could have like a blue notification dot wdyt

paologum:
i was also thinking maybe it might look good as a button next to the generate chart

Screenshot 2025-03-27 at 2 43 38 PM

server/__tests__/admin-routes.test.ts
asivath: they're already a 404 failure case and this was just really testing the assert, which I removed

.github/workflows/deploy-prod.yml
asivath: upload of other servers? no

server/src/admin/admin-routes.ts
asivath: i think we should ask aaron this. i just added to satsify the type for now but I wasn't sure if wanted the admins to see anything

.github/workflows/deploy-prod.yml
paologum: no i mean like you didn't have this before so does it have it or something, i'm asking the implications of not having this from before

paologum: actually i guess you did answer that question but what are the implications if there were any

server/src/routes/routes.ts
asivath: it does show it but the timeline is broken and not on your cross plot page. should I fix these things now or for another pr and for now not save?

server/src/user/chart-history.ts
asivath: maybe but it's so small and simple that's its not worth trying to find a library and I don't think there's a native function for it

server/src/user/chart-history.ts
asivath: i was hoping you wouldn't notice

.github/workflows/deploy-prod.yml
asivath: before it was deleting from the uploads directory though the permissions make it weird since this folder is generated and not owned by deployuser so its unable to really delete from here. it need sudo and I've gave it permissions to delete certain things but rather than having to do all that it's better to manage it here

paologum: wait it deletes the upload directories?

server/src/routes/routes.ts
paologum: it should break because it is a different compilation of setting and a different component that displays it because it should be on `/crossplot`
you can do it in a different pr but this pr should like not allow storing of crossplot data for now

server/src/user/chart-history.ts
paologum: hehehe

.github/workflows/deploy-prod.yml
asivath: no it never really did I'm just changing slightly how it deletes thing, before it was through permissions, now it's through this change

asivath:
yeah i think both the blue dot and putting it in the navbar make sense

paologum:
it is also failing on preview still

asivath:
yeah I'd rather do all frontend changes in one pr

paologum (approved):

PR #390, opened 2025-03-27
github secrets

Ganthry:
Third times the charm. I'm hoping this all I need to add github secrets to the .env files. Also is there a seperate script for deploying to deep?


paologum:
me or aditya have to make a new release tag thingie to deploy to staging and then run an action manually with a tag to deploy to prod which also deploys to dde

although idr if aditya said that perms are messed up so anyone on the team CAN do it but you probably shouldn't and u should let us know

paologum:
also metrics auth exist on all of these right

Ganthry:
I know aditya said he added it as a repo secret but idk if thats the same thing

paologum (approved):

PR #420, opened 2025-04-23
revert changes properly

paologum:
this properly makes the revert button work

known issue: expanding the column and pressing revert will revert the expansion in some cases. no real way around this tbh

i track the dirty state manually in each action. I tried using reactions but it seems against the philosophy of reactions to make changes to the observable AND it doesn't work after a while debugging

https://github.com/user-attachments/assets/8f58d981-783c-4804-a15b-a08c57c12649

Closes 410

### DIFF
--- a/app/src/state/actions/crossplot-actions.ts
+++ b/app/src/state/actions/crossplot-actions.ts
@@ -835,9 +835,6 @@ export const setCrossPlotDatapackConfig = action(
     if (!state.crossPlot.chartX) state.crossPlot.chartX = state.crossPlot.columns!.children[0];
     state.crossPlot.chartY = state.crossPlot.columns!.children[0];
     state.crossPlot.state.matchesSettings = true;
-    runInAction(() => {
-      state.crossPlot.state.matchesSettings = true;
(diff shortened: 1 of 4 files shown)


app/src/state/actions/crossplot-actions.ts
```diff
@@ -801,22 +828,26 @@ export const setCrossPlotDatapackConfig = action(
...
       }
     }
-    if (!state.crossPlot.chartX) setCrossPlotChartX(state.crossPlot.columns!.children[0]);
-    setCrossPlotChartY(state.crossPlot.columns!.children[0]);
+    if (!state.crossPlot.chartX) state.crossPlot.chartX = state.crossPlot.columns!.children[0];
+    state.crossPlot.chartY = state.crossPlot.columns!.children[0];
+    state.crossPlot.state.matchesSettings = true;
+    runInAction(() => {
+      state.crossPlot.state.matchesSettings = true;
+    });
```
asivath, line 840: setting twice

asivath (approved):

PR #422, opened 2025-04-26
418 bug report page

JacquiFreeElectron:
https://github.com/user-attachments/assets/3cf0e11d-154a-4c85-ad39-ead61baaf082


paologum:
add translations to en.json

app/src/ReportBug.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+        {!isSubmitted && (
+          <Typography variant="h6" maxWidth={900} mx="auto">
+            {t("report-bug.subtitle")}
+          </Typography>
+        )}
+      </Box>
+
+      {!isSubmitted ? (
+        <Box bgcolor="background.main" border={1} p={3} width="100%" maxWidth="1000px" mt={2}>
```
paologum, line 117: border should be a color that matches so `divider`

paologum (changes requested):
needs more margin to the top of the page

Screenshot 2025-04-26 at 10 49 09 PM

when a user presses send, and it is a correct input, then there should be a loading variable that is used by `TSCDialogLoader` that once the send request is completed is set back to false

on smaller screens there should be some margin still

Screenshot 2025-04-26 at 10 52 38 PM

paologum:
the text field component used in the mock has a resizable thing in the bottom right, but it shouldn't be resizable imo so you should probably keep it static 

https://github.com/user-attachments/assets/289abe1b-1d17-447e-9c59-e63206f3b258

paologum:
there's a weird margin here, could you make it more flush

Screenshot 2025-04-29 at 2 42 17 PM

paologum (approved):

PR #426, opened 2025-04-30
381 frontend datapack comment

lbostre:
- frontend for datapack comments

DEMO
https://github.com/user-attachments/assets/b0cc489e-8f80-4ef9-a107-72e0f93d9215

551a3b75427a627f0278cef9f913d751
68263349decef0052e25aa43e466ae4b
7fb8a1204616f8ec4690454340a0130f
e4d6c42afa59610d2ba91233831bba4a


paologum:
some frontend comments

* can you make the report menu and its buttons inside the menu spaced like that; also the positioning relative to the three dots should be the same as the last comment placement
* line up account icons
* 

image

image

image

.idea/workspace.xml
paologum, line 1: what are these changes

app/src/components/TSCComment.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  const handleReport = () => {
+    console.log("reported");
+    handleClose();
+  };
+
+  const handleDelete = () => {
+    console.log("delete");
+    handleClose();
+  };
```
paologum, line 56: logs

paologum (changes requested):
shouldn't most recent be at the top? or how is this sorted/should it be sorted this way

also how do you decide the 10 days ago, because they should probably be consistent, for instance my new comment i just posted is at the bottom with the date and not like minutes or days ago


Screenshot 2025-05-07 at 4 21 17 PM

.idea/workspace.xml
lbostre: this got added on accident it's removed now

lbostre:
The 10 days ago part was hardcoded for that screenshot. I could have it displayed like that, or I could display just the date the comment was posted. Right now it shows the date it was posted.

paologum:
i think that admins should be able to delete any comment


Screenshot 2025-05-08 at 4 21 34 PM

.idea/workspace.xml
paologum, line 1: i think you should just remove the change

paologum (changes requested):
this should probably keep the new lines between

for flagging and deleting, just use state variables for signed in user and admin


https://github.com/user-attachments/assets/1ceadc6c-23a8-4655-a925-e97909ab4210

app/src/components/TSCComment.tsx
```diff
@@ -0,0 +1,143 @@
...
+          <Box>
+            <MenuItem onClick={() => handleReport()}>
+              <OutlinedFlagIcon sx={{ color: "text.primary" }} />
+              <Typography color="text.primary">Report</Typography>
+            </MenuItem>
+            {((isSelf && isSignedIn) || isAdmin) && (
+              <MenuItem onClick={() => handleDeleteComment()}>
+                <DeleteIcon sx={{ color: "text.primary" }} />
+                <Typography color="text.primary">Delete</Typography>
+              </MenuItem>
```
paologum, line 121: translation

app/src/components/TSCComment.tsx
```diff
@@ -0,0 +1,143 @@
...
+      </Avatar>
+      <div className={styles.textContainer}>
+        <div className={styles.topTextContainer}>
+          <Typography className={styles.usernameText}>{username}</Typography>
+          <Typography className={styles.dateText}>{formattedDate}</Typography>
+        </div>
+        <Typography className={styles.commentText}>{text}</Typography>
+        {flagged && (
+          <Typography className={styles.flaggedText} style={{ color: theme.palette.error.main }}>
+            This comment has been flagged.
```
paologum, line 137: translation

paologum (changes requested):
ok last thing, just add translation otherwise it is all good

paologum:
probably add these to general actions part, because they are pretty general words (for report and delete)

paologum (approved):

app/src/components/TSCComment.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { t } = useTranslation();
+  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
+  const open = Boolean(anchorEl);
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down("sm"));
+  const [mobileOpen, setMobileOpen] = useState(false);
+  const [flagged, setFlagged] = useState(isFlagged);
+  const { state } = useContext(context);
+  const [isSignedIn] = useState(state.user !== null);
+  const [isAdmin] = useState(state.user.isAdmin);
```
paologum, line 45: oops actually, because you are abstracting this as a component outside of a global state, you shouldn't use the `useContext` variables from the global state here. you should probably have those as passed parameters above, which will render the proper updates when they change.

paologum:
Screenshot 2025-05-11 at 6 06 03 PM

i think my previous comment might be the reason this is happening so it may be fixed, but in the picture you can see i'm not signed in, but i'm still able to delete a reported comment. is this intentional? i don't believe this should be the case seeing as I am not the original user (not signed in) and i'm also not the admin

app/src/components/TSCDiscussion.tsx
```diff
@@ -18,17 +69,42 @@ export const Discussion = () => {
...
+              <Comment
+                key={index}
+                id={comment.id}
+                username={comment.username}
+                date={comment.date}
+                text={comment.text}
+                isSelf={comment?.isSelf}
+                isFlagged={comment?.isFlagged}
+                handleDelete={() => deleteComment(comment.id)}
+                userSignedIn={state.user !== null}
```
paologum, line 102: user always exists, you should use `state.isLoggedIn`

PR #440, opened 2025-05-15
changed treatise datapack to be official and function to be external …

Changk820:
…not treatise upload


Changk820:
on that admin page we only fetch metadata, so things like the ageunits and format isn't fetched since its part of basedatapackprops, then the asserts prevented me from deleting things. This is a bug also on dev site.

Changk820:
https://development.treatise.geolex.org/

server/src/upload-datapack.ts
```diff
@@ -78,7 +78,11 @@ export const processAndUploadDatapack = async (uuid: string, parts: AsyncIterabl
   }
   const { filepath, tempProfilePictureFilepath, datapackMetadata, pdfFields } = result;
   try {
-    if ((isOfficialDatapack(datapackMetadata) || isWorkshopDatapack(datapackMetadata)) && !isAdmin) {
+    if (
+      (isOfficialDatapack(datapackMetadata) || isWorkshopDatapack(datapackMetadata)) &&
+      !isAdmin &&
+      uuid !== "official"
+    ) {
```
paologum, line 85: shouldn't `isOfficialDatapack` catch this?

paologum (changes requested):

server/src/upload-datapack.ts
paologum: just remove the uuid !== official, unless you have a case against that

Changk820: I run into the issue of not having permission: Only admins can upload official or workshop datapacks

The "official" uuid is not really even an actual user right? It's just a folder we have. Should we then make it an actual admin user? or just have this check?

paologum: ok this would potentially lead to a misuse of the function in the future, so instead change this to verify IF there is an official datapack or workshop datapack being uploaded, either the user is an admin, or the bearer token is the same. to add the bearer token to this function make a parameter `options?: { bearerToken?: string }`
you can keep the previous bearerToken checks in `user-routes` or remove them, up to you since with this check they may be redundant
this way, usage of this function is clear and doesn't let anyone who passes official through upload a datapack (since users can edit the request)

PR #444, opened 2025-05-21
380 backend datapack comment

lbostre:
https://github.com/user-attachments/assets/ae8b59d7-6f2e-49a4-8cc1-18f5ad7cf14f

- backend for datapack comments


app/src/components/TSCComment.tsx
```diff
@@ -58,12 +54,44 @@ export const Comment = ({
...
+      } else {
+        if (response.status === 500) {
+          actions.pushError(ErrorCodes.DATAPACK_COMMENT_REPORT_FAILED);
+        } else if (response.status === 404) {
+          actions.pushError(ErrorCodes.DATAPACK_COMMENT_NOT_FOUND);
+        }
+      }
+      handleClose();
+    } catch (e) {
+      console.log("error", e);
```
paologum, line 90: probably `console` error this instead

app/src/components/TSCDiscussion.tsx
```diff
@@ -4,55 +4,171 @@ import { CustomDivider } from "./TSCComponents";
...
+          loadedComments.sort((a, b) => b.dateCreated.getTime() - a.dateCreated.getTime());
+          setComments(loadedComments);
+          actions.removeAllErrors();
+        } else {
+          if (response.status === 500) {
+            actions.pushError(ErrorCodes.DATAPACK_COMMENT_FETCH_FAILED);
+          }
+        }
+      } catch (e) {
+        console.log("error", e);
```
paologum, line 71: push server error probably

paologum (changes requested):
Screenshot 2025-05-22 at 11 19 59 AM

if someone is not signed in, then you should probably still let them type in and intereact with the box and button. if they press send or submit comment, then you can push an error or warning saying that they must sign in first. the way it is right now, it is difficult to tell what you have to do

paologum:
when you try to post a comment before logging in, it should keep the comment present in the box. this way they can copy paste it or keep it before logging in, because how it is right now is it disappears and they lose their work they have

app/src/components/TSCDiscussion.tsx
```diff
@@ -4,55 +4,178 @@ import { CustomDivider } from "./TSCComponents";
...
+      }
+      try {
+        const response = await fetcher(`/user/datapack/comments/${encodeURIComponent(id)}`, {
+          method: "GET",
+          credentials: "include"
+        });
+        if (response.ok) {
+          const commentsArray = await response.json();
+          const loadedComments: CommentType[] = [];
+          for (const com of commentsArray) {
```
paologum, line 46: asserts here right?

paologum (changes requested):

paologum (changes requested):
ok i did some debugging. there is a lot of flashing as you said so the easiest solution is to create a global variable for comments on `datapackProfilePage`. you can load it in the same way, but it must be cleared when the user exists the page so in the return method of the `useEffect` in the `DatapackProfile.tsx` NOT in the discussion since that will cause weird flashing

app/src/components/TSCDiscussion.tsx
```diff
@@ -4,55 +4,158 @@ import { CustomDivider } from "./TSCComponents";
...
+  const deleteComment = async (id: number) => {
+    if (state.user.isAdmin) {
+      const response = await actions.adminDeleteDatapackComment(id);
+      if (!response) {
+        actions.pushError(ErrorCodes.DATAPACK_COMMENT_DELETE_FAILED);
+        return;
+      }
+      actions.deleteDatapackProfileComment(id);
+      actions.pushSnackbar("Comment deleted.", "success");
+      actions.removeAllErrors();
```
paologum, line 131: since it is a state level variable, i would put these into the `adminDeleteDatapackComment` since these are directly related to that function and would be needed if someone called this somewhere else

paologum (changes requested):
ok one last thing

paologum (approved):

Issue #450, opened 2025-05-22
Bug Report: Issue with chart


tsconline:
## Description

non image/svg file attached

## Attachments

![Screenshot 2025-05-22 144106.jpg](https://github.com/earthhistoryviz/tsconline-bug-reports/raw/main/bug-reports/2025-05-22T18-52-32-163Z-ct/Screenshot%202025-05-22%20144106.jpg)
- [issue.txt](https://github.com/earthhistoryviz/tsconline-bug-reports/raw/main/bug-reports/2025-05-22T18-52-32-163Z-ct/issue.txt)

PR #455, opened 2025-05-22
431 route bug report to GitHub issues email

asivath:
example of issue this will create: https://github.com/earthhistoryviz/tsconline/issues/450

will also send emails to users specified in secret

this won't work in preview since dev doesn't have the secrets yet but lmk if you want me to add them


app/src/ReportBug.tsx
```diff
@@ -80,14 +80,33 @@ export const ReportBug: React.FC = () => {
...
+    const allowedExtensions = /\.(png|jpe?g|gif|svg|txt|log|json|csv)$/i;
+    const allowedMimeTypes = [
+      "image/png",
+      "image/jpeg",
+      "image/gif",
+      "image/svg+xml",
+      "text/plain",
+      "application/json",
+      "text/csv"
+    ];
```
paologum, line 92: i think you can limit what file they can upload in the actual component, so you don't only catch it after they select it, the component will limit their selectability (i think) but that creates a better ui

server/src/routes/routes.ts
```diff
@@ -32,6 +32,130 @@ import { loadPublicUserDatapacks } from "../public-datapack-handler.js";
...
+    "image/svg+xml",
+    "text/plain",
+    "application/json",
+    "text/csv"
+  ];
+  for await (const part of parts) {
+    if (part.type === "file") {
+      const { filename, mimetype } = part;
+      if (!allowedExtensions.test(filename)) {
+        reply.status(400).send({ error: `Unsupported file extension: ${filename}` });
```
paologum, line 63: lowkey this function is like insane/huge, do you not think you should like break this into smaller functions? for easier readability like part processing, file uploading, issue uploading, and email sending?

bc also this needs tests :D and breaking it up would be easier for that

paologum (changes requested):

paologum:
can you upload a video of you uploading it and also how the repo looks after the issue and where the issue goes? or you can't do local

server/src/routes/routes.ts
```diff
@@ -32,6 +34,92 @@ import { loadPublicUserDatapacks } from "../public-datapack-handler.js";
...
+      console.error("Error submitting bug report:", errorText);
+      reply.status(500).send({ error: "Failed to submit bug report" });
+      return;
+    }
+    reply.send({ message: "Bug report submitted successfully" });
+  } catch (error) {
+    console.error("Error processing bug report:", error);
+    reply.status(500).send({ error: "Internal Server Error" });
+  }
+};
```
paologum, line 122: prolly logger error these

PR #456, opened 2025-05-22
Fix UI lag

asivath:
- `ColumnInfo` is no longer observed directly. instead, we maintain a flattened map of `RenderColumnInfo` objects. each `RenderColumnInfo` includes a children array containing the names of child columns, which are used as keys to look up their entries in `columnHashMap`
- reactions are used to sync changes from `RenderColumnInfo` to `ColumnInfo`
- adding reactions like this does have a performance impact, especially for `fontsInfo`. If we wanted more performance we could change it to manual updates for all fields, or choose some fields that should be completely manual rather than relying on actions.
- lmk if some of the fields i've marked as non-observable in `RenderColumnInfo` need to be observable and if the reactions cover what needs to be synced


app/src/state/actions/column-actions.ts
```diff
@@ -467,36 +469,142 @@ export const applyRowOrder = action(
...
+      column.popup = updated.popup;
+      column.parent = updated.parent;
+      column.enableTitle = updated.enableTitle;
+      column.width = updated.width;
+      column.rgb = updated.rgb;
+      column.columnSpecificSettings = updated.columnSpecificSettings;
+      column.showAgeLabels = updated.showAgeLabels;
+      column.showUncertaintyLabels = updated.showUncertaintyLabels;
+    }
+  );
```
paologum, line 587: what is the reason for separating the two

app/src/state/actions/column-actions.ts
```diff
@@ -608,40 +728,48 @@ export const setColumnMenuTabValue = action((tabValue: number) => {
   state.columnMenu.tabValue = tabValue;
 });
 
-let searchColumnsAbortController: AbortController | null = null;
+let currentSearchToken = 0;
 export const searchColumns = action(async (searchTerm: string, counter = { count: 0 }) => {
```
paologum, line 732: you said you can't just add a flag for searching before and after this for adding loading?

app/src/state/actions/column-actions.ts
asivath: the second reaction has heavier toJS calls, so if everything were in one reaction expanding/showing the column would unnecessarily trigger those. the first reaction is for lightweight fields that change more frequently

app/src/state/actions/column-actions.ts
asivath: are you saying to add a loader so when it's searching it shows that instead? if that's the case then I can add that

paologum: although its like a small delay but still it felt weird

asivath:
https://github.com/user-attachments/assets/a0751ae1-122c-4fa3-abaa-a4895cdcabd0

paologum (approved):

PR #457, opened 2025-05-25
437 help backend

paologum:
* add help page full functionality
* use command `yarn pull-help` to pull help files from source (also runs on `yarn dev`)
* any edits to the `tsconline-help` repo will automatically deploy to all the functional servers
* an `index.md` file will replace the parent component of markdown as shown in the picture below, otherwise it will just show and link all the children
* you can put html in the markdown and it will still render (file format example)
* i made `help-routes.ts` because of potentially the want to edit it on the main site rather than github.

Screenshot 2025-05-25 at 12 53 57 AM

https://github.com/user-attachments/assets/1f84ae56-7a8a-4da1-8703-b0f5afc91e29


asivath (commented):
when clicking on a breadcrumb it leads me to a 404 page since the url is this:
image

also are you going to make this work in preview? or in another pr

PR #472, opened 2025-06-06
417 wire workshops to the display paolo

paologum:
* added functionality for the editing of `regLink` `regRestrict` and `description` and `creatorUUID` on workshop editing and cleaned up routes
* added workshop util tests

Screenshot 2025-06-05 at 11 42 13 PM

Screenshot 2025-06-05 at 11 40 30 PM

Screenshot 2025-06-05 at 11 42 00 PM


asivath (commented):
the check for whether any fields has changed when editing isn't working, it tries to edit it when I change nothing. also I can't add users/add files if I don't edit other things like title/description as well. reset form also removes the reg link when a workshop has one.

server/src/admin/admin-routes.ts
```diff
@@ -591,6 +620,19 @@ export const adminEditWorkshop = async function adminEditWorkshop(
...
+    if (regLink) {
+      if (!validator.isURL(regLink)) {
+        reply.status(400).send({ error: "Invalid registration link" });
+        return;
+      }
+      fieldsToUpdate.regLink = regLink;
+    }
+    if (creatorUUID) {
+      fieldsToUpdate.creatorUUID = creatorUUID;
+    }
```
asivath, line 635: i do think we should start moving these kinds of checks into fastify's schema validation instead of having every route manually check fields. i think it's cleaner and handles things like email/link validation without extra dependencies. but since that involves the auth files, it might make more sense to do it in the recaptcha pr instead.

paologum: yeah sounds good i can do that there bc im changing a lot of the config there alr

paologum:
i think i fixed your bugs? send a video so i know for sure if they still aren't working

asivath:
https://github.com/user-attachments/assets/db8688c5-875b-40de-90f1-6a8462031dd1

it tries to edit when I change nothing

paologum:
ok oops i fixed that

asivath (approved):
pushed to wrong branch lol but lg

PR #480, opened 2025-06-07
441 notifications admin email

lbostre:
comment-summary-email
- daily comments if any are sent via email to provided recipients


app/src/components/TSCDiscussion.tsx
```diff
@@ -43,7 +43,8 @@ export const Discussion = observer(() => {
               dateCreated: new Date(com.dateCreated),
               flagged: Boolean(com.flagged),
               isSelf: com.uuid === uuid,
-              commentText: com.commentText
+              commentText: com.commentText,
+              datapackTitle: com.datapackTitle
```
paologum, line 47: i can't highlight the whole thing, but i realize this is kinda weird. if you assert the comment type, what is the need to reconstruct it into a variable of the same type?

shared/src/index.ts
```diff
@@ -815,6 +827,23 @@ export function assertAutoPlotMarker(o: any): asserts o is AutoPlotMarker {
...
+  if (typeof o.commentText !== "string") throwError("CommentType", "commentText", "string", o.commentText);
+  if (typeof o.datapackTitle !== "string") throwError("CommentType", "datapackTitle", "string", o.datapackTitle);
+  if (!(o.dateCreated instanceof Date) && typeof o.dateCreated !== "string") {
+    throwError("CommentType", "dateCreated", "Date or ISO string", o.dateCreated);
+  }
+  if (o.flagged !== undefined && typeof o.flagged !== "number")
+    throwError("CommentType", "flagged", "number", o.flagged);
+  if (typeof o.username !== "string") throwError("CommentType", "username", "string", o.username);
+  if (o.pictureUrl && typeof o.pictureUrl !== "string") throwError("CommentType", "pictureUrl", "string", o.pictureUrl);
+  if (o.isSelf && typeof o.isSelf !== "boolean") throwError("CommentType", "isSelf", "boolean", o.isSelf);
```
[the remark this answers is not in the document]
paologum: hmm it doesn't really make sense for the comment itself to have a field in a shared state that depends on a frontend state. since like on the backend, `isSelf` doesn't make sense really at least in my opinion.

paologum (changes requested):

app/src/components/TSCDiscussion.tsx
paologum: if that's the case ,it probably should not be shared right? these changes are specifically front end specifications right so you would leave the type on the frontend as an extension of the shared comment type that the route sends

paologum: what was the reasoning to put it in shared

lbostre: I put it in shared because the send-email functionality in the server now fetches comments from the database in almost the same format as the TSCDiscussion component (everything except pictureURL). And the isSelf attribute is slightly weird because it almost doesn't need to be part of the type at all. It is only necessary when posting a new comment and updating the global state via actions.addDatapackProfileComment to reflect that. All the existing comments that are fetched simply check if the uuid in global state matches the uuid stored in the comment, which can be determined upon loading.

lbostre: ok i just realized that the isSelf stuff is not needed at all so i removed it. The reasoning for the shared stuff is still the same tho

paologum (approved):

PR #482, opened 2025-06-07
481 external chart front end changes

Changk820:


app/src/GenerateExternalChart.tsx
```diff
@@ -58,43 +53,69 @@ export const GenerateExternalChart: React.FC = () => {
...
+          }
+        } catch (err) {
+          console.warn("Failed to set stage ages or unit steps: ", err);
+        }
 
-            const values = parts.slice(2).map(Number); // [minTotal, maxTotal, minNew, maxNew, minExtinct, maxExtinct]
+        if (chartType === "Treatise" && minMaxPlot) {
+          const parts = minMaxPlot.split("-");
+          if (parts.length == 6) {
+            const values = parts.map(Number); // [minTotal, maxTotal, minNew, maxNew, minExtinct, maxExtinct]
```
paologum, line 118: you don't want to do parameters for this too?

server/src/upload-handlers.ts
```diff
@@ -438,8 +439,10 @@ export async function processMultipartPartsForDatapackUpload(
           return { code: 415, message: "Invalid file type for datapack file" };
         }
         if (file.file.bytesRead > 3000 && !isProOrAdmin) {
-          await cleanupTempFiles();
-          return { code: 413, message: "File is too large" };
+          if (!(process.env.BEARER_TOKEN && options?.bearerToken === process.env.BEARER_TOKEN)) {
+            await cleanupTempFiles();
+            return { code: 413, message: "File is too large" };
+          }
```
paologum, line 445: did you update a test case for this?

app/src/GenerateExternalChart.tsx
```diff
@@ -58,43 +53,69 @@ export const GenerateExternalChart: React.FC = () => {
...
-                console.warn(
-                  `Warning: Oldest time should be greater than newst time. Oldest time: ${oldestTime}, Newest time: ${newestTime}`
-                );
-              } else {
-                actions.setBaseStageAge(oldestTime, "Ma");
-                actions.setTopStageAge(newestTime, "Ma");
-                actions.setUnitsPerMY(0.1, "Ma");
-              }
+        try {
+          if (!isNaN(topVal) && !isNaN(baseVal) && !isNaN(unitStep)) {
```
paologum, line 95: it's impossible for this to ever be not a number no?

app/src/GenerateExternalChart.tsx
```diff
@@ -58,43 +53,69 @@ export const GenerateExternalChart: React.FC = () => {
...
+            if (topVal > baseVal) {
+              console.warn("Top Unit Value must be less than Base Unit Value. No changes made.");
+            }
+            if (unitStep <= 0) {
+              console.warn("Unit Step must be greater than 0. No changes made.");
             }
+            actions.setBaseStageAge(baseVal, unitType);
+            actions.setTopStageAge(topVal, unitType);
+            actions.setUnitsPerMY(unitStep, unitType);
```
paologum, line 104: you have these if statements but they don't stop the setting of them? what is the meaning of printing no changes are made then

app/src/GenerateExternalChart.tsx
```diff
@@ -58,43 +53,69 @@ export const GenerateExternalChart: React.FC = () => {
...
+          } else {
+            console.warn(
+              "Base Unit Value, Top Unit Value, and Unit Step must be provided with number values. No changes made."
+            );
+            console.warn("Provided values: baseVal: ", baseVal, ", topVal: ", topVal, ", unitStep: ", unitStep);
+          }
+        } catch (err) {
```
paologum, line 111: can you explain why this and the above code wouldn't be redundant?, additionally this block would never get run due to my previous comment

paologum (changes requested):

app/src/GenerateExternalChart.tsx
Changk820: Not really, since this is specific to treatise, which should always have three charts.

server/src/upload-handlers.ts
Changk820: I had to add one to meet the bare minimum branching reqs

app/src/GenerateExternalChart.tsx
Changk820: This:
        if (isNaN(parsedTop) || isNaN(parsedBase) || isNaN(parsedStep)) {
          console.warn(
            "Base Unit Value, Top Unit Value, and Unit Step must ALL be provided with number values. No changes made."
          );
          console.warn("Provided values: baseVal: ", baseVal, ", topVal: ", topVal, ", unitStep: ", unitStep);
        }

Was added because the value of baseVal, topVal, and Unitstep can never not be a number due to
  const parsedBase = parseFloat(queryParams.get("baseVal") ?? "");
  const baseVal = isNaN(parsedBase) ? 10 : parsedBase;
  the default values.

I thought it would be nice to know your numbers provided didn't work, and got set default ones.

I changed to Using default values instead of no changes made to be more clear

With regards to being impossible to not a number, I guess it is true now since I have default values. before it didn't use default values if the parameters was NaN in the get request.

Before defaults values were there, this was a parameter that could be optional, and just doing 
          if (topVal && baseVal && unitStep) {
Instead of
          if (!isNaN(topVal) && !isNaN(baseVal) && !isNaN(unitStep)) {
         would prevent me from using 0 as values, thus i did if isNaN.

Should i remove this redundacy or leave it for safety even if impossible?

Changk820: on second thought ill rmeove it because itll never get there. but i think the first nan should be left

app/src/GenerateExternalChart.tsx
```diff
@@ -84,35 +84,29 @@ export const GenerateExternalChart: React.FC = () => {
...
-              console.warn("Top Unit Value must be less than Base Unit Value. No changes made.");
-            }
-            if (unitStep <= 0) {
-              console.warn("Unit Step must be greater than 0. No changes made.");
-            }
+          if (topVal > baseVal) {
+            console.warn("Top Unit Value must be less than Base Unit Value. No changes made.");
+          } else if (unitStep <= 0) {
+            console.warn("Unit Step must be greater than 0. No changes made.");
+          } else {
```
paologum, line 100: topVal and baseVal being wrong shouldn't stop unitStep from being set though

paologum (changes requested):

PR #484, opened 2025-06-09
468 hyperlinks powerpoints and pdfs

asivath:
https://github.com/user-attachments/assets/50f30237-0533-4364-b48a-dcdb91462a09

image

- most of the diff comes from refactoring the workshop component and tests
- changed how workshop auth works by making it work per route instead of global, since i didn't add recaptcha checks to hyperlinks. also made it possible for the prehandlers to add the user object so the routes don't need to call findUser again
- i also started the process of using more of the features of schema validation in the fastify routes instead of manual checks in the routes, but only did it for a couple workshop routes


paologum:
if i fill in a reglink field, then remove it, the workshop form has to be refilled out or it keeeps throwing errors that reglink isn't in uri form. i also have reglink editing work on my pr btw

paologum:
image


wouldn't we want more of a popup similar to our other popups ?
you could also do the below but i would prefer our popups that we have normally that would be like register first 

although im not 100% sure if aaron would support the use of a popup but that's my preference at least so lmk

image

asivath:
i fixed both those bugs you mentioned.

> image
> 
> wouldn't we want more of a popup similar to our other popups ? you could also do the below but i would prefer our popups that we have normally that would be like register first

i'm assuming you clicked the link as an admin and got this page. on the frontend I allowed admins to click links even when not registered but forgot to allow admins through on the backend, so you got that page, a normal user shouldn't be able to see it. 

the problem with trying to use a popup is that once they click the hyperlink the browser moves away from our app and we no longer have control for things like popups. rn I have it so that the link is disabled and a tooltip pops up to register, which is shown in the video. if we really wanted a popup we could stop using hyperlinks entirely and use buttons instead.

paologum:
ok i think the disabled is fine

server/src/index.ts
```diff
@@ -97,8 +102,9 @@ server.addHook("onRequest", async (request: FastifyRequest & { startTime?: [numb
 server.addHook("onResponse", async (request: FastifyRequest & { startTime?: [number, number] }, reply) => {
   const duration = process.hrtime(request.startTime);
   const durationInMs = duration[0] * 1000 + duration[1] / 1e6;
-  totalHttpRequestCount.labels(request.method, request.routerPath, reply.statusCode.toString()).inc();
-  totalHttpRequestDuration.labels(request.method, request.routerPath, reply.statusCode.toString()).set(durationInMs);
+  const routePath = request.routeOptions?.url || "unknown";
+  totalHttpRequestCount.labels(request.method, routePath, reply.statusCode.toString()).inc();
+  totalHttpRequestDuration.labels(request.method, routePath, reply.statusCode.toString()).set(durationInMs);
```
paologum, line 107: what is this for

server/src/index.ts
```diff
@@ -59,6 +63,7 @@ const server = fastify({
     }
   }*/
 });
+server.decorateRequest("user", undefined);
```
paologum, line 66: also this

paologum (changes requested):
since you also abstracted the add datapacks, there is a bug that is similar to the ones you fixed, where adding a datapack doesn't immediately update the workshop on the app unless it refreshes

server/src/index.ts
asivath: it gets rid of the deprecation warning, functionally it's the same

server/src/index.ts
asivath: this actually adds a user field to every request object at run time, which is important for optimizations. from the fastify docs:

>Decorating core objects with this API allows the underlying JavaScript engine to optimize the handling of server, request, and reply objects. This is accomplished by defining the shape of all such object instances before they are instantiated and used.

app/src/admin/AdminWorkshopDatapacksForm.tsx
```diff
@@ -0,0 +1,110 @@
...
+            {loading && <TSCDialogLoader open={loading} headerText="Adding Datapack" />}
+          </Box>
+        </Box>
+      </Dialog>
+    ) : (
+      <Dialog open={true} onClose={() => setUploadDatapacks(false)} maxWidth={false}>
+        <DatapackUploadForm
+          close={() => setUploadDatapacks(false)}
+          upload={actions.adminUploadDatapackToWorkshop}
+          type={{ type: "workshop", uuid: `workshop-${currentWorkshop.workshopId}` }}
```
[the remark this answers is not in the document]
asivath: i added it to shared

server/src/index.ts
paologum: do you think you can comment that here for future people to understand

paologum (approved):

PR #486, opened 2025-06-12
script for removing extra translations in other languages

jay-jhk-lee:


paologum:
i'm confused, shouldn't this include something during modify translations that removes and cleans up as you remove keys during that command

paologum:
but yes that would be nice if you had a good idea for that ?

.github/workflows/translation-pipeline.yml
```diff
@@ -21,6 +21,10 @@ jobs:
       - name: Install Dependencies
         run: yarn install && yarn build
 
+      - name: Remove Extra Translations
+        working-directory: server
+        run: yarn remove-extra-translations
+
       - name: Verify Translations
         working-directory: server
         run: yarn verify-translations
```
paologum, line 30: i forget, does verify translation fail if there are extra keys that aren't present in english

jay-jhk-lee: yeah

paologum (approved):

PR #501, opened 2025-06-19
477 london organize columns based on path

lbostre:
- create data column groups based on path

### DIFF
--- a/server/src/db/london/json-to-txt.ts
+++ b/server/src/db/london/json-to-txt.ts
@@ -35,9 +35,16 @@ async function loadJSONS() {
   return { datasets, events, intervals, columns, subdatasets };
 }
 
-type NestedDict = { [key: string]: NestedDict | string[] };
-async function processEventColumns(datasets: arkL_datasets[], columns: arkL_columns[], events: arkL_events[]) {
-  const pathDict: NestedDict = {};
+type StringDict = { [key: string]: string[] };
+type StringDictSet = { [key: string]: Set<string> };
+
+async function processEventColumns(
+  datasets: arkL_datasets[],
+  columns: arkL_columns[],
+  events: arkL_events[],
+  pathDict: StringDictSet,
+  linesDict: StringDict
+) {
   for (const column of columns) {
     let columnLines: string[] = [];
     if (column.column_type === "events") {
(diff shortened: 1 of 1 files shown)


server/src/db/london/json-to-txt.ts
```diff
@@ -174,20 +186,110 @@ async function processBlockColumns(
...
+    lines.push("");
+    if (linesDict[path]) {
+      lines = [...lines, ...(linesDict[path] || [])];
     }
   }
   return lines;
 }
 
+// custom split function that handles paths with "/"'s that are not meant to split the path, as well as "/"'s that are part of the group name
+const customSplit = (path: string): string[] => {
```
paologum, line 253: if path is empty after a trim or is just empty, you probably wanna return `[]` right

server/src/db/london/json-to-txt.ts
```diff
@@ -174,20 +186,110 @@ async function processBlockColumns(
...
+      parentDepth++;
+      currentPart += char;
+    } else if (char === ")") {
+      parentDepth = Math.max(parentDepth - 1, 0);
+      currentPart += char;
+    } else if (char === "/" && parentDepth === 0) {
+      if (!(path[i - 1] === " " && path[i + 1] === " ")) {
+        // split here
+        parts.push(currentPart.trim());
+        currentPart = "";
```
paologum, line 269: this is not a split if there are spaces on both sides? do you have an example of this? also what about one space on each side

server/src/db/london/json-to-txt.ts
```diff
@@ -174,20 +186,110 @@ async function processBlockColumns(
...
+        // if path is the last element, add the column lines
+        if (i === pathArray.length - 1) {
+          pathDict[path]?.add(columnName!);
+          linesDict[path] = [...(linesDict[path] || []), ...columnLines];
+        } else {
+          pathDict[path]?.add(pathArray[i + 1]!);
+        }
       }
-      lines.push("");
+    }
```
paologum, line 223: i guess i'm slightly confused here, can you explain to me how you keep the order of the paths in consistent ordering going down the txt file if you just add the single path from the path array instead of keeping track of the order? am i missing something

server/src/db/london/json-to-txt.ts
```diff
@@ -174,20 +186,110 @@ async function processBlockColumns(
           line = `\t${inter[column.col_if_not_intvx as keyof arkL_intervals]}\t${inter.base_age}\t\t${inter.interval_notes !== null ? inter.interval_notes.replace(/[\r\n]+/g, " ") : ""}\t${inter.colour || ""}`;
         } else
           line = `\t${inter.intervalx}\t${inter.base_age}\t\t${inter.interval_notes !== null ? inter.interval_notes.replace(/[\r\n]+/g, " ") : ""}\t${inter.colour || ""}`;
-        lines.push(line);
+        columnLines.push(line);
+      }
+      columnLines.push("");
+      organizeColumn(column.path!, column.columnx!, columnLines, pathDict, linesDict);
```
paologum, line 192: like my previous comment, is path guaranteed to be empty

server/src/db/london/json-to-txt.ts
lbostre: I think i was mistaken the column names for the path for that part. ill remove it

server/src/db/london/json-to-txt.ts
lbostre: based on the json file, every column does have a path. I'm not sure what the sql table looks like but it seems like a column in the json file is guaranteed to have a path and column name

PR #502, opened 2025-06-23
485 download datapacks individually

Neelp895:
Added APIs to download files and data packs individually from the Workshop details page. Modified workshop details frontend to have download links for each file and data pack.


app/src/state/actions/workshop-actions.ts
```diff
@@ -65,3 +66,101 @@ export const fetchWorkshopFilesForDownload = action(async (workshop: SharedWorks
...
+    return;
+  }
+
+  const data = await response.json();
+  const { fileName, fileData, fileType } = data;
+
+  // Convert base64 to Blob
+  const byteCharacters = atob(fileData);
+  const byteArrays = [];
+
```
paologum, line 151: with `shared` responses that come from the server to the app, we probably wanna have an asserts statement that is defined in the `index.ts` file in the `shared` module that we call here to make sure the types are correct, if the response is incorrect the asserts should throw the error and propogate to the try catch i asked for above in the comments

app/src/state/actions/workshop-actions.ts
```diff
@@ -65,3 +66,101 @@ export const fetchWorkshopFilesForDownload = action(async (workshop: SharedWorks
...
+  for (let i = 0; i < byteCharacters.length; i += 512) {
+    const slice = byteCharacters.slice(i, i + 512);
+    const byteNumbers = Array.from(slice).map((char) => char.charCodeAt(0));
+    byteArrays.push(new Uint8Array(byteNumbers));
+  }
+
+  const file = new Blob(byteArrays, { type: fileType });
```
paologum, line 158: can you define this function in `non-action-util.ts` or in a `util` file in general, since this may be useful in other ways for other routes in the future? accepting a base64 and outputting a blob

server/__tests__/workshop-routes.test.ts
```diff
@@ -10,9 +10,12 @@ import * as generalFileHandlerRequests from "../src/file-handlers/general-file-h
 import { User, Workshop } from "../src/types";
 import * as util from "../src/util";
 import * as fsp from "fs/promises";
+import * as fs from "fs";
+// import { createReadStream } from "fs";
 import * as uploadHandlers from "../src/upload-handlers";
 import { SharedWorkshop } from "@tsconline/shared";
 import { fetchAllWorkshops, fetchWorkshopCoverImage } from "../src/workshop/workshop-routes";
+// import { Readable } from "stream";
```
paologum, line 18: remove comments

server/src/workshop/workshop-auth.ts
```diff
@@ -76,6 +82,22 @@ export const workshopRoutes = async (fastify: FastifyInstance, _options: Registe
...
+    required: ["workshopId", "fileName"]
+  };
+  const workshopDataPackParams = {
+    type: "object",
+    properties: {
+      workshopId: { type: "number" },
+      datapackTitle: { type: "string" }
+    },
+    required: ["workshopId", "datapackTitle"]
+  };
```
paologum, line 100: add ajv parameter checks, so you can probably do `workshopId: { type: "integer", minimum: "1" }`
and `datapackTitle: { type: "string", minLength: 1 }`
this isn't in the other one schemas on main but i have added this in a new pr that hasn't merged yet so i'd like to make sure this stays in that same realm

also, then you'll need to make sure the tests account for these schema restrictions, so verify the response when the workshopId is not a number or less than 0 fails with the proper json response and also for fileName and datapackTitle

server/src/workshop/workshop-routes.ts
```diff
@@ -157,6 +159,95 @@ export const downloadWorkshopFilesZip = async (
...
+    const workshopUUID = getWorkshopUUIDFromWorkshopId(workshopId);
+    const directory = await getUserUUIDDirectory(workshopUUID, true);
+
+    let filePath = path.join(directory, "files");
+    filePath = path.join(filePath, fileName);
+
+    if (!(await verifyFilepath(filePath))) {
+      reply.status(500).send({ error: "Invalid file path" });
+      return;
+    }
```
paologum, line 191: this should probably be a function in `fetch-user-files.ts` called `getFileFromWorkshop`

server/src/workshop/workshop-routes.ts
```diff
@@ -157,6 +159,95 @@ export const downloadWorkshopFilesZip = async (
...
+    if (!isAuthorized) {
+      reply.status(403).send({ error: "Unauthorized access" });
+      return;
+    }
+
+    //check if fileName is valid
+    if (!fileName || fileName.includes("..") || fileName.includes("/")) {
+      reply.status(500).send({ error: "Invalid file name" });
+      return;
+    }
```
paologum, line 180: `verifyFilepath` should handle this so you don't have to add this here

server/src/workshop/workshop-routes.ts
```diff
@@ -157,6 +159,95 @@ export const downloadWorkshopFilesZip = async (
...
+    if (!isAuthorized) {
+      reply.status(403).send({ error: "Unauthorized access" });
+      return;
+    }
+
+    //check if datapackTitle is valid
+    if (!datapackTitle || datapackTitle.includes("..") || datapackTitle.includes("/")) {
+      reply.status(500).send({ error: "Invalid datapack name" });
+      return;
+    }
```
paologum, line 224: same as above

server/src/workshop/workshop-routes.ts
```diff
@@ -157,6 +159,95 @@ export const downloadWorkshopFilesZip = async (
...
+      return reply.send(stream);
+    } catch (e) {
+      reply.status(500).send({ error: "Error creating file stream" });
+    }
+  } catch (e) {
+    reply.status(500).send({ error: "An error has occured" });
+  }
+};
+
+export const downloadWorkshopDataPack = async function downloadWorkshopDataPack(
```
paologum, line 204: minor thing, but our naming convention for datapacks is as if it is one word, so do you mind naming it to be `downloadWorkshopDatapack` ?

server/src/workshop/workshop-routes.ts
```diff
@@ -157,6 +159,95 @@ export const downloadWorkshopFilesZip = async (
...
+      return;
+    }
+
+    const workshopUUID = getWorkshopUUIDFromWorkshopId(workshopId);
+    const dataPackPath = await getUploadedDatapackFilepath(workshopUUID, datapackTitle);
+
+    if (!(await checkFileExists(dataPackPath))) {
+      reply.status(500).send({ error: "File does not exist" });
+      return;
+    }
```
paologum, line 232: the `getUploadedDatapackFilepath` runs `realpath` so it already checks for file existence within

server/src/workshop/workshop-routes.ts
```diff
@@ -157,6 +159,95 @@ export const downloadWorkshopFilesZip = async (
...
+
+    try {
+      const fileBuffer = readFileSync(dataPackPath);
```
paologum, line 235: we want all file checks to be asynchronous, so can you use the `await` version of `readFile`?

paologum (changes requested):
probably shouldn't show up as a hyperlink when the user is not registered since they shouldn't be able to download the files if they aren't registered to the workshop

there should be a reliable error that gets thrown if that is the case as well. (if an unregistered user to the workshop somehow tries to access the link or is able to send that request).

 Also the screenshot is returning a recaptcha error and hanging but that shouldn't matter if the links disappear if they aren't registered

Screenshot 2025-06-29 at 7 45 31 PM

server/src/workshop/workshop-auth.ts
```diff
@@ -92,4 +114,22 @@ export const workshopRoutes = async (fastify: FastifyInstance, _options: Registe
...
+    downloadWorkshopFile
+  );
+  fastify.get(
+    "/workshop-datapack/:workshopId/:datapackTitle",
+    {
+      config: { rateLimit: moderateRateLimit },
+      schema: { params: workshopDataPackParams },
+      preHandler: [verifyAuthority]
+    },
+    downloadWorkshopDatapack
```
paologum, line 133: these should also have recaptcha

vite.config.ts.timestamp-1750814718223-0f796c95ca61c.mjs
paologum, line 1: remove this file

paologum (changes requested):
shoot i approved, but the pr preview isn't working so you'll need to fix that

image

paologum (approved):

PR #504, opened 2025-06-25
Tsc 530 wire recaptcha to match strings on backend

paologum:
* added the recaptcha prehandlers to be more specific
* added route checks to make sure the app defined routes and the ones directly defined are the same and define the prehandlers correctly so that proper testing of those prehandlers can happen for those routes. (this lets us define what the recaptcha actions that should be cross referenced in the actual prehandler during testing of the recaptcha prehandler)
* use enum to make recaptcha action keys for the required items to be easier to keep consistent across app and server
* also made two different routes for downloading a private and public datapack's pdfs that have dfiferent pdfs. then abstracted the zip file handler out of the routes and into `fetch-user-files.ts` (`/datapack/download/public/files/:datapackTitle/:uuid` and `/datapack/download/private/files/:datapackTitle/:uuid`)
* will add the workshop auth prehandler that checks if the user is in the workshop before accessing the route


server/__tests__/admin-routes.test.ts
```diff
@@ -291,8 +294,15 @@ beforeAll(async () => {
       }
     };
   });
-  await app.register(adminAuth.adminRoutes, { prefix: "/admin" });
   app.get("/admin/official/private/metadata", adminFetchPrivateOfficialDatapacksMetadata);
+  app.addHook("onRoute", (routeOptions: RouteOptions) => {
+    appRoutes.push(
+      ...initializeAppRoutes(routeOptions, {
+        recatpchaHandlerName: "verifyRecaptchaPrehandler"
```
asivath, line 301: typo in recatpchaHandlerName

asivath (commented):
login-routes should also be changed to align with the other routes. probably not this pr though

asivath (approved):

PR #513, opened 2025-07-04
478 london chron

Ganthry:


server/src/db/london/json-to-txt.ts
```diff
@@ -35,6 +35,58 @@ async function loadJSONS() {
...
+      for (const event of sequenceEvents) {
+        const sequenceType =
+          event.event_type === "seq bdy" ? "SB" : event.event_type === "mfs" ? "MFS" : event.event_type;
+        line = `\t${event.eventx != null ? event.eventx : ""}\t${sequenceType}\t${event.age}\t${event.seq_scale}${event.notes_2020 !== null ? `\t${event.notes_2020?.replace(/[\r\n]+/g, " ")}` : ""}`;
+        lines.push(line);
+      }
+      // for blank space between columns for tscreator to parse
+      lines.push("");
+      if (column.path && column.columnx) {
+        organizeColumn(column.path, column.columnx, lines, pathDict, linesDict);
```
paologum, line 83: Can you refactor the organize columns so it's more modular

Basically calling the OrganizeColumn function after every column is processed
So processSequenceColumns, eventColumns, etc. would return a collection of paths and column lines and organizecolumn would take all of them
not much would change, just the order of things we do the organizing

paologum (changes requested):
Looks Good!

paologum:
the labels on these are different when i'm comparing the internal, shouldn't it not include mfs?

Screenshot 2025-07-16 at 8 57 40 PM

image

Ganthry:
> the labels on these are different when i'm comparing the internal, shouldn't it not include mfs?
> 


Should be good now

image

image

PR #518, opened 2025-07-22
516 london chron columns

Ganthry:
So all time ranges work, and the output seems to be right. The only issue now, is that between the three chron columns, they all have the same data, this is how they are in the viewer. The only difference is their names, I don't know if this is intentional but it's a little bit weird.  And the path for Late Triassic is different from the rest as well, which I also don't know if its intentional.
image

... fast forward to end of data

image

### DIFF
--- a/server/src/db/london/json-to-txt.ts
+++ b/server/src/db/london/json-to-txt.ts
@@ -199,6 +199,92 @@ async function processBlockColumns(
   return;
 }
 
+
+async function processChronColumns(
+  datasets: arkL_datasets[],
+  columns: arkL_columns[],
+  intervals: arkL_intervals[],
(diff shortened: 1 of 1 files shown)


Ganthry:
Nevermind, I compared it to the internal and its a bit different, ill change it to match up

Ganthry:
So now label uses the Chron Labels, when there isn't a Chron Label available it leaves itself blank. I tried to get it to use Chron Label for the rightmost block column, but that meant a lot of data was left out. Everything past Oxfordian until Torcian
image
image

This shows no labels past Oxfordian. I think this should be fine as long as its not the norm that most data has labels that go with them (a totally full first block column). The series columns are not changed.
image

Issue #524, opened 2025-07-31
cleanup mcp transports


paologum:
make sure to clean them up if they get stale

PR #525, opened 2025-07-31
522 mcp part of repo incremental work on remote work

paologum:
This sets up a really basic mcp server using fastify.

I added fill in tools that aren't tested with vitest as placeholders to see how they work on the remote server.

The fastify part should work with sse and geo gpt and will just use a POST, GET, and DELETE request only. The MCP part should handle usage of any tools and specifics so the fastify part should be done. We will just need to work on tools now.

Therefore `fastify.ts` shouldn't really be touched, other than for adding cleanup of transports which i added a task for #524


mcp/src/fastify.ts
```diff
@@ -0,0 +1,89 @@
...
+      // Create new session
+      transport = new StreamableHTTPServerTransport({
+        sessionIdGenerator: () => randomUUID(),
+        onsessioninitialized: (id) => {
+          transports[id] = transport;
+        }
+      });
+
+      transport.onclose = () => {
+        if (transport.sessionId) delete transports[transport.sessionId];
```
Ganthry, line 35: Is there something that deletes session data even if onclose isn't called for some reason like a timeout or something? Or does it close automatically after a bit?

PR #536, opened 2025-09-11
527 scheduled london db sync

Ganthry:


app/src/components/datapack_display/TSCCompactDatapackRow.tsx
```diff
@@ -62,8 +63,13 @@ export const TSCCompactDatapackRow: React.FC<TSCCompactDatapackRowProps> = obser
           e.stopPropagation();
           if (loading) return;
           setLoading(true);
-          !skeleton && (await onChange(datapack));
-          setLoading(false);
+          try {
+            if (!skeleton) {
+              await onChange(datapack);
+            }
+          } finally {
```
Neelp895, line 70: is a catch statement needed here?

server/src/index.ts
```diff
@@ -467,6 +469,16 @@ if (process.env.EMAIL_USER && process.env.EMAIL_PASS && process.env.NODE_ENV ===
...
 
+cron.schedule("0 0 * * *", async () => {
+  try {
+    await processLondonDatapack();
+  } catch (e) {
+    logger.error("Error migrating cached datapacks", e);
+  }
+});
+
+server.get("/migrate-london", moderateRateLimit, fetchLondonDatapack);
```
Neelp895, line 480: Not sure if we have some logging system, but would it be a good idea to have some logging function here where if the database refresh fails on the cron job it would show up in some logs?

Neelp895 (commented):
/deploy

PR #540, opened 2025-09-22
534 transfer jar and datapack

### LINKED ISSUE #534: Transfer jar / datapack pull to 2025 drop box

nd1739655076:
resolve #534


Ganthry (commented):
This looks great! Could you also add a package.json command that will run the decode from root directory like "yarn decrypt" or something like that? Also is there a way to add checks, at least during boot up with "yarn dev" that could run this script or make sure that it was run?

Ganthry:
Also would it be possible, just from the perspective of someone else working on something different, once they decrypt the .env, they're able to commit the decrypted .env, is there a way to protect from this and make sure yarn encrypt is run before commiting? 

This might require setup with something like githooks. We had this previously running quality and test before you were able to push but we realized this was super annoying to do EVERY time we pushed but yarn ecrypt is a lot shorter so it should be fine. If theres already a built in way to do that, that's fine too.

PR #547, opened 2025-10-05
Issue #538 Feature/auto load default datapack

VIll999:
Issue #538


app/src/state/initialize.ts
Ganthry: For this, you shouldn't be changing the way initialization works, you should be able to change the tests charts.test.ts. Right now the tests load AfricaBight and use that for most things. You can change these tests to work with your new initialization and also add a test to test for this initialization.

Ganthry (approved):

PR #566, opened 2025-10-31
559 workshop alert bubble

nd1739655076:


Ganthry:
Ok I think the last change before merging would be make it a bit like this. Default (if there's a workshop) it should be open. And a bit like the video I linked

https://github.com/user-attachments/assets/a039d7bf-138c-4fa8-90c1-4d359275be0b

PR #580, opened 2025-11-10
550 chart window popoff

Neelp895:


app/src/ChartPreview.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { context } from "./state";
+import { TSCSvgComponent } from "./components";
+import { Box } from "@mui/material";
+import TimeLine from "./assets/icons/axes=one.svg";
+
+export function ChartPreview() {
+  const { state, actions } = useContext(context);
+
+  useEffect(() => {
+    const channel = new BroadcastChannel("tsconline-chart");
```
Ganthry: I think we're recreating BroadcastChannel listeners and reactions after some state changes we'd end up having multiple listeners all firing. Would creating just one channel cause problems to come up with the implementation? It might be a little memory safer

app/__tests__/app/charts.test.ts
```diff
@@ -250,3 +250,93 @@ test("check if Map Points Functional", async ({ page }) => {
 
   await expect(page.locator("text=Nigeria Coast")).toBeVisible();
 });
+
+test("check if new window button works", async ({ page, context }) => {
```
Ganthry: nice tests!

PR #589, opened 2025-11-18
584 mcp generate chart

eselznick2:


mcp/src/mcp.ts
```diff
@@ -1,5 +1,74 @@
 import { McpServer, ResourceTemplate } from "@modelcontextprotocol/sdk/server/mcp.js";
 import z from "zod";
+import { readFile } from "fs/promises";
+import * as path from "path";
+import dotenv from "dotenv";
+
```
Neelp895: So I read something about session level context in MCPs, where we can feed preloaded information into the MCP session - context across tools. No idea if you already did this, but if not, would def be worth adding. This way we can feed the session some important context about our chart generation process and maybe reduce confusion errors.

server/src/routes/mcp-routes.ts
```diff
@@ -0,0 +1,210 @@
...
+  ColumnToggles
+} from "../settings-generation/unified-chart-generation.js";
+import { SettingsSchema } from "../settings-generation/build-settings-schema.js";
+
+/**
+ * MCP route: list datapacks (public + official)
+ */
+export async function mcpListDatapacks(_request: FastifyRequest, reply: FastifyReply) {
+  const publicDatapacks = await loadPublicUserDatapacks();
+  const officialDatapacks = await fetchAllPrivateOfficialDatapacks();
```
Neelp895: No error handling here. May need it

server/src/settings-generation/build-settings-from-datapacks.ts
```diff
@@ -0,0 +1,161 @@
+import {
+  ColumnInfo,
```
Neelp895: Could you just briefly explain what this file does

mcp/src/mcp.ts
```diff
@@ -13,32 +82,392 @@ export const createMCPServer = () => {
...
+{
+  "datapackTitles": ["GTS2020"],
+  "overrides": { "topAge": 5, "baseAge": 150 },
+  "columnToggles": { "on": ["column-id-1"], "off": ["column-id-2"] },
+  "useCache": true
+}
+
+These are just examples for changing datapacks, overrides, and column toggles. You can mix and match as needed.
+
+Remember: datapack settings will persist across calls until resetChartState is used. If a column is toggled off, it stays off until explicitly toggled on again. Same with toggling it on.
```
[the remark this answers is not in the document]
Ganthry: This was only tested on ChatGPT 5.1 and 5.2. I tried testing it with copilot but it was really, really, really bad at everything

mcp/src/mcp.ts
Ganthry: Yea this should be added but I think it might have to be another PR. There's already so much in here. I want to add it when I end up adding session ID inside the states (cause right now we have one shared state, definitely not final)

Neelp895 (approved):

PR #632, opened 2026-01-29
finished Hosung Ryu's about page

dennis42702:
finished Hosung Ryu's about page


Ganthry:
To get past a quality check you just have to go back into the repo and run "yarn quality" and this runs our linter. We thought about adding githooks to make this automatic but it ended up being really annoying being forced to run quality with every push so now we're always forgetting to run quality.

PR #633, opened 2026-01-29
607 MCP link to Tsconline chart

Neelp895:
Notes:
- is not tested with private datapacks yet.  (in a new PR)
-  Need to fix window popoff for this specific feature (in a new PR)

### DIFF
--- a/app/src/Settings.tsx
+++ b/app/src/Settings.tsx
@@ -16,8 +16,6 @@ import { TabWrapper, TabsWrapper } from "./components";
 import LoadSettings from "./settings_tabs/LoadSettings";
 import SaveSettings from "./settings_tabs/SaveSettings";
 import { useTranslation } from "react-i18next";
-import { toJS } from "mobx";
-import { useEffect } from "react";
 
 export const Settings = observer(function Settings() {
   const { state, actions } = useContext(context);
--- a/mcp/src/mcp.ts
+++ b/mcp/src/mcp.ts
@@ -309,7 +309,6 @@ The assistant SHOULD still provide the direct URL as plain text under the embed.
         const chartPath = typeof json.chartpath === "string" ? json.chartpath : "";
         const chartHash = typeof json.hash === "string" ? json.hash : "";
 
-
         const filePath = path.join("..", "server", chartPath);
         const svg = await readFile(filePath, "utf8");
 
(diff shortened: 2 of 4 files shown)


app/src/state/initialize.ts
```diff
@@ -6,13 +6,18 @@ export async function initialize() {
   // as the server hosts the /charts endpoint.  So, we'll hard-code that for ourselves here.
 
   try {
+    if (state.skipInitialization) {
+      console.log("Skipping initialization as per state flag.");
```
Ganthry: you can get rid of the console

mcp/src/mcp.ts
```diff
@@ -313,11 +318,30 @@ The assistant SHOULD still provide the direct URL as plain text under the embed.
...
+        const mcpLinkJson = JSON.stringify(mcpLinkObj);
+        const mcpLinkBase64 = Buffer.from(mcpLinkJson).toString("base64");
+        const mcpToolUrl = `${frontendUrl}/mcp-chart?mcpChartState=${mcpLinkBase64}`;
+
         return {
           content: [
             {
               type: "text",
-              text: `Chart generated!\n\nURL: ${serverUrl}${chartPath}\n\nCurrent state:\n${JSON.stringify(currentChartState, null, 2)}`
+              text: `Chart generated!\n\nURL: ${mcpToolUrl}
```
Ganthry: For the actual AI it displays charts inline using markdown using a link direct to the chart svg (like this https://pr-preview.geolex.org/public/charts/abc1e9bf53357ee54887a8e311431a7a/chart.svg) so pass in the simple svg of the chart too so it can display it inline otherwise it looks like this 

Image Also you can kinda see in the screenshot ^ but it totally prints out the wrong link. Like the tool returns the right link but the AI adds some nonesense and then /chart.svg so it might be getting confused.

All this means is be really clear in the description for this tool now like display` Chart` and then your mcp chart link can just be given straight to the user to take them to TSCOnline to play around with it. Bully the AI if you have to

app/src/Settings.tsx
```diff
@@ -29,6 +31,12 @@ export const Settings = observer(function Settings() {
   const tabKeys = Object.keys(SettingsMenuOptionLabels);
   const tabIndex = tabKeys.indexOf(state.settingsTabs.selected);
 
+  useEffect(() => {
+    console.log("Settings applied, state : ", toJS(state.settings));
```
Ganthry: console.log core

Ganthry:
This is really really well done and works great, just being clearer for the mcp tool and the other stuff and should be good

Here's the link to the geogpt published agent for PR Preview https://geogpt-sg.zero2x.org/tools/agent-builder?isEmoji=false&siteCode=ZZAlfxUezu5SwXec&sessionId=daabfa96-9ac2-4d25-821b-e1e240d0597e it *should* work if it doesn't I can show how to make a GeoGPT agent

PR #644, opened 2026-02-02
implemented mcp server auth

JiBing17:


Ganthry (changes requested):
This is a googe implementation. I have just noticed a problem with our merge pipeline and how we prepare the PR preview. It seems that for pr preview for the .env file we actually copy the .env file from server which doesn't have MCP_AUTH_TOKEN yet (correctly). So the merge pipeline needs a quick change to line 596 and needs to copy in secrets the way that deploy dev and the rest receive github secrets. Idk why its setup this way tbh.

PR #647, opened 2026-02-04
616 mcp session ids

dennis42702:
finished mapping user's chart state to the corresponding user's session_id


mcp/src/mcp.ts
```diff
@@ -446,6 +527,123 @@ The assistant SHOULD still provide the direct URL as plain text under the embed.
       } catch (e) {
         return { content: [{ type: "text", text: `Error generating chart: ${String(e)}` }] };
       }
+
```
Ganthry: Not sure what this is for, if we don't need it we can get rid of it

Ganthry (changes requested):

PR #661, opened 2026-02-09
639 mcp login page redirect

JiBing17:


Ganthry (changes requested):
I like it, could you make it so on login the navbar at top is also hidden not just after successful login? 

This screen: 

Image

Ganthry:
Nice I want to run how it looks past Aaron tmrw but it should be good

Ganthry:
I think icon removal and then its good

PR #662, opened 2026-02-10
660 clean up mcp chart generation

eselznick2:
Now that we have chart linking we don't really need to be able to work with schemas or see the actual settings cause linking will just display them in TSCOnline for us. So I carved out AI->schema->columnInfo->settings xml so now its just AI->columnInfo->settings xml. It was short enough to group into one file too. When this is near the end and GeoGPT settings match up with app settings 1:1 I want to totally get rid of settings-to-xml and just make a route so the 3000 does all settings generation, AI will just supply the columnInfo


Neelp895 (approved):
Looks Good!

PR #671, opened 2026-02-11
Made generate chart use internal datapack if no datapacks were selected

sivathanuas:


app/src/state/actions/generate-chart-actions.ts
```diff
@@ -156,6 +157,43 @@ export const compileChartRequest = action(
       from?: string;
     }
   ) => {
+    // Auto-load internal datapack if none selected
```
Ganthry: Could you add a guard here to prevent multiple executions from this

app/src/state/actions/generate-chart-actions.ts
```diff
@@ -156,6 +157,43 @@ export const compileChartRequest = action(
...
+    if (state.config.datapacks.length === 0) {
+      generalActions.setLoadingInternalDatapackWhenNoDatapacksSelected(true);
+      if (!state.datapacks.some((dp) => dp.title === "TimeScale Creator Internal Datapack")) {
```
Ganthry: For this it might be better if it always has to load the datapack. This is kind of a work around because it seems like theres a small delay on concurrent generate charts once its already loaded in. This doesn't seem to be happenning with normal datapack generation once its loaded in.

If you can figure out where the delay is coming from no need. But if you can't it is possible to just load the datapack on each run

Ganthry (approved):

PR #672, opened 2026-02-12
655 generate general sessionid on tools

dennis42702:
added verifying sessionid and creating sessionid to the tools. Made unverified session last for 5min from 2min.

### DIFF
--- a/mcp/src/mcp.ts
+++ b/mcp/src/mcp.ts
@@ -147,7 +147,7 @@ function requireSession(es: { sessionId: string; entry: SessionEntry; internalNo
 }
 
 // Helper to wrap tool responses with sessionId
-function wrapResponse(content: any, sessionId: string) {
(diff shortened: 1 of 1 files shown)


Ganthry (changes requested):
This is a good implementation. I think the intention was to have every tool check sessionID/generate one even listDatapacks. We can check with aaron on Friday to be safe though

Ganthry:
Ok so for this task, its gonna expand into also letting some functionality through for an unsigned in user. For each tool:
because EVERY tool call gives the AI a sessionID, we can assume every tool will be called with a sessionID. We want to allow both logged in and non logged in users (even if userInfo is undefined) to be able to update charts, work with charts, full functionality.

The purpose of the login comes more into play in the specific upload datapack tool (in development), the updateChart tool (limit whether a user can edit their private datapacks), or if a user can list their private datapacks.

PR #738, opened 2026-04-20
added geogpt settings page

dennis42702:


app/src/account_settings/GeoGPTSettingsPopup.tsx
eselznick2, line 1: for this you don't need the read more on the switches because they've already got the i icons (that means you can probably make it thinner). 

Image

eselznick2:
I also think, as part of this, you could add a little green "active or not active" icon to quickly show users wether they are or are not linked to a GeoGPT session, this should be somewhere on the navbar. There should also be another button at the bottom, that only shows if you have a session active at the moment that is basically "sever" connection to your current session (this would basically delete the geogpt session from memory)

eselznick2:
At this point the UI is about done and we would need to hookup the settings to the actual functions for "cowork"

eselznick2 (approved):

PR #750, opened 2026-04-30
730 geogpt background color

sivathanuas:


eselznick2:
Two things for this PR now. 
1) Some small merge conflicts have come up with recent changes (I believe harmless, like lowercase checking for column names), those can be resolved.
2) In the case where a user has a GeoGPT session open and it is tied to their TSCOnline browser. The TSCOnline server has to monitor the UI and see if a user is changing the background color of any columns. If a user does, right now TSCOnline has no way of telling the mcp server that "a user has changed the background color." This should be another quick addition of assembling all this information and adding it to the chartState object we are already sending to the MCPServer (we already have the infrastructure for this with, width, columns, settings, etc... so you would just be adding background color as another thing getting collected).

eselznick2:
Calcareous Nannofossils, modify tool descriptions to have GeoGPT call list column and apply settings to leafs and not branches of our charts

eselznick2:
@copilot resolve the merge conflicts in this pull request

eselznick2 (approved):

PR #764, opened 2026-05-29
Aditya Patel Tutorial

apatel773:
Completed tutorial and added profile image to the website.


eselznick2:
For this, to get these tests to pass, run "yarn quality" from your terminal to run the linter. You've also been added as a co-owner so you should be able to run the pipeline. You can also see the preview on the server by typing "/deploy" as a comment in this conversation

